PR #22, opened 2019-11-15
Update README.md

kikou:
- Add Luma, RodeoFX and other people to acknowledgements
- Formatting fixes and corrections


sebastienblor:
Maybe we should add a line warning about the linux crashes and the LD_PRELOAD thing

kikou:
> Maybe we should add a warning about the linux crashes and the LD_PRELOAD thing

What would be the warning?

PR #23, opened 2019-11-15
Build fixes

frenchdog:
- Add missing "main" plugInfo.json file
- Fix compilation error on OSX and missing library in usd_translator

### DIFF
--- a/render_delegate/material.cpp
+++ b/render_delegate/material.cpp
@@ -358,8 +358,8 @@ void HdArnoldMaterial::Sync(HdSceneDelegate* sceneDelegate, HdRenderParam* rende
             const auto* volumeNetwork = TfMapLookupPtr(map.map, HdMaterialTerminalTokens->volume);
 #else
             const auto* surfaceNetwork = TfMapLookupPtr(map.map, UsdImagingTokens->bxdf);
-            const auto* displacementNetwork = nullptr;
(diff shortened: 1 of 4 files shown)


render_delegate/material.cpp
```diff
@@ -358,8 +358,8 @@ void HdArnoldMaterial::Sync(HdSceneDelegate* sceneDelegate, HdRenderParam* rende
             const auto* volumeNetwork = TfMapLookupPtr(map.map, HdMaterialTerminalTokens->volume);
 #else
             const auto* surfaceNetwork = TfMapLookupPtr(map.map, UsdImagingTokens->bxdf);
-            const auto* displacementNetwork = nullptr;
-            const auto* volumeNetwork = nullptr;
+            const auto displacementNetwork = nullptr;
+            const auto volumeNetwork = nullptr;
```
sirpalee: This compiles fine on 10.15 and the toolset there, so type inference must have been improved. Which MacOS/Compiler version are you using?

Instead of auto, we could just use `decltype(surfaceNetwork)` for the `displacementNetwork` and `volumeNetwork` type. This better at expressing that `displacementNetwork` and `volumeNetwork` are the same type as `surfaceNetwork` but they are initialized with a nullptr value.

frenchdog: Excellent suggestion! I will give it a try.

PR #37, opened 2019-11-19
Document, and enforce naming convention rules

### LINKED ISSUE #6: Document naming conventions
Document naming conventions for the various Arnold entities, via a markdown file and add it to doxygen.

- Properties
- Attributes
- Primvars
- Schemas
- Namespaces

sebastienblor:
- Added a file NAMING_CONVENTIONS.md that gives the current naming conventions for arnold-usd nodes and attributes
- Enforce these rules in the usd reader / writer / schemas

Fixes #6

**Additional context**


sebastienblor:
Do we really want to call the file NAMING_CONVENTIONS.md ?

NAMING_CONVENTIONS.md
```diff
@@ -0,0 +1,37 @@
...
+
+Additional arnold attributes on USD-native nodes (geometries, lights, etc...) need to be in the "arnold" namespace
+
+Example:
+def DomeLight "my_light"
+{
+ float intensity=2
+ int arnold:max_bounces=99
+}
+
```
sirpalee: We should document that primvars named `primvars:arnold:xxx` (ie `primvars:arnold:max_bounces`) are going to be translated to built-in parameters, rather than as user attributes.

NAMING_CONVENTIONS.md
```diff
@@ -0,0 +1,37 @@
...
+## Arnold Shaders
+
+Arnold shaders are supported with UsdShade Shader primitives. The info:id attribute should be the shader name, with an "arnold" namespace. The attributes will be in the "inputs" scope.
```
sirpalee: I don't think we need to specify that shader parameters have to be prefixed with `inputs:` that's part of the UsdShadeShader schema, and considered a standard.

NAMING_CONVENTIONS.md
sebastienblor: ok, I need to add support for this in the usd procedural then

NAMING_CONVENTIONS.md
BigRoy: Maybe the "best of both world" is to keep it documented but mention it is the UsdShadeShader standard practice, potentially with a link to it that elaborates it. Or would that end up being too verbose? At this stage with it being so new for many I think even the smallest hints to reasoning on implementation/workflow could be helpful.

sirpalee: Sounds good to me, a link to the UsdShadeShader could be useful.

NAMING_CONVENTIONS.md
sirpalee: Let's make a separate ticket for that.

sirpalee:
Sorry, I missed your question about the name of the file. There is no need to use CAMEL_CASE.md for the file name, that's only for special files, so `naming_conventions.md` would be just as good. Also, I think placing it in the docs subfolder would be a better place, as it's nothing more special than our build documentation.

We should also add a link to the file from the main page, so it's easily accessible.

NAMING_CONVENTIONS.md
sebastienblor: created #43

NAMING_CONVENTIONS.md
sebastienblor: done. Let me know if more details could be useful.

sirpalee:
Looks good!

NAMING_CONVENTIONS.md
```diff
@@ -0,0 +1,37 @@
+Arnold USD naming conventions
```
kikou: Should be 'Arnold USD Naming Conventions"

NAMING_CONVENTIONS.md
```diff
@@ -0,0 +1,37 @@
...
+=======================
+
+The Arnold USD implementation supports arnold-specific attributes through custom schemas.
```
kikou: How about:
"In order to expose the entirety of the Arnold features in USD, we supply custom schemas for Arnold primitives in USD, and we support Arnold-specific attributes on regular USD prims. This documents the naming conventions in use to map Arnold node and parameters names in USD."

NAMING_CONVENTIONS.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Arnold attributes on USD nodes
+
+Additional arnold attributes on USD-native nodes (geometries, lights, etc...) need to be in the "arnold" namespace
```
kikou: Additional Arnold attributes on native USD nodes (geometry, lights, etc.) need to be in the `arnold` namespace:

NAMING_CONVENTIONS.md
```diff
@@ -0,0 +1,37 @@
...
+
+Example:
+def DomeLight "my_light"
```
kikou: Can you put USD code this in a code block?

NAMING_CONVENTIONS.md
```diff
@@ -0,0 +1,37 @@
...
+Arnold shaders are supported with UsdShade Shader primitives. The info:id attribute should be the shader name, with an "arnold" namespace. The attributes will be in the "inputs" scope.
+
+Example:
+def Shader "my_shader"
+{
+    uniform token info:id = "arnold:standard_surface"
+    color3f inputs:base_color = (0,0,1)
+}
+
+## Arnold nodes
```
kikou: Arnold Nodes

NAMING_CONVENTIONS.md
kikou: Arnold shaders are supported as UsdShade Shader primitives. The `info:id` attribute should be the Arnold shader name, prefixed by the `arnold` namespace. The shader parameter are placed in the `inputs` scope.

NAMING_CONVENTIONS.md
```diff
@@ -0,0 +1,37 @@
+Arnold USD naming conventions
+=======================
+
+The Arnold USD implementation supports arnold-specific attributes through custom schemas.
+
+## Arnold attributes on USD nodes
+
+Additional arnold attributes on USD-native nodes (geometries, lights, etc...) need to be in the "arnold" namespace
+
+Example:
```
kikou: Please delete this line.

NAMING_CONVENTIONS.md
```diff
@@ -0,0 +1,37 @@
...
+{
+ float intensity=2
+ int arnold:max_bounces=99
+}
+
+## Arnold Shaders
+
+Arnold shaders are supported with UsdShade Shader primitives. The info:id attribute should be the shader name, with an "arnold" namespace. The attributes will be in the "inputs" scope.
+
+Example:
```
kikou: Please delete this line.

NAMING_CONVENTIONS.md
```diff
@@ -0,0 +1,37 @@
...
+def Shader "my_shader"
```
kikou: Please put in a code block.

NAMING_CONVENTIONS.md
```diff
@@ -0,0 +1,37 @@
...
+{
+    uniform token info:id = "arnold:standard_surface"
+    color3f inputs:base_color = (0,0,1)
+}
+
+## Arnold nodes
+
+Any Arnold node can be represented in USD, with a camel-cased node type, prefixed with "Arnold".
```
kikou: Can you explain why we need to transliterate Arnold node names to CamelCase (the Arnold node names are normally snake_case)? This makes it more complicated to write translation tools. What are the rules when converting from snake_case to CamelCase, especially when dealing with capitals and acronyms?

NAMING_CONVENTIONS.md
```diff
@@ -0,0 +1,37 @@
...
+def Shader "my_shader"
+{
+    uniform token info:id = "arnold:standard_surface"
+    color3f inputs:base_color = (0,0,1)
+}
+
+## Arnold nodes
+
+Any Arnold node can be represented in USD, with a camel-cased node type, prefixed with "Arnold". 
+Example:
```
kikou: Please remove, and replace with a colon at the end of the previous sentence.

NAMING_CONVENTIONS.md
```diff
@@ -0,0 +1,37 @@
...
+def ArnoldStandardSurface "my_shader" {}
```
kikou: Please put in a code block.

NAMING_CONVENTIONS.md
```diff
@@ -0,0 +1,37 @@
...
+def ArnoldSetOperator "my_operator" {}
+def ArnoldPolymesh "my_mesh" {}
+def ArnoldSkydomeLight "my_light" {}
+
+For these nodes, the attributes don't need to be namespaced.
```
kikou: For these nodes, the attribute names are the Arnold parameter names and are not namespaced.

NAMING_CONVENTIONS.md
frenchdog: Interesting point. FYI, every USD Schemas I saw were using camel case style.
If Arnold ones keep the snake case style, it could just look a little bit strange when looking at a usda file. Currently a Skydome prim looks like that:
`def ArnoldSkydomeLight "domeLight1" {}`
but would like a little but "use alien" like that:
`def arnold_skydome_light "domeLight1" {}`

Not a big deal for sure, but I would vote for keeping consistency with USD current Hydra plugin echo system and so use camel case.

sebastienblor: I just did a test, and it's not possible to build a schema class called `arnold_skydome_light`. It fails at linking time, because it needs to start with a capital. It does work with `Arnold_skydome_light`.

frenchdog: I would keep it without underscores. It does not makes sense to me to try to look "almost" like native name but not really. Converting from snake_case to UpperCamelCase should not be a big issue on the tool translation side.

docs/naming_conventions.md
```diff
@@ -0,0 +1,47 @@
+Arnold USD Naming Conventions
+=======================
+
+In order to expose the entirety of the Arnold features in USD, we supply custom schemas for Arnold primitives in USD, and we support Arnold-specific attributes on regular USD prims. This documents the naming conventions in use to map Arnold node and parameters names in USD.
+
+## Arnold attributes on USD nodes
```
frenchdog: Should be `USD prims` instead of `USD nodes`

docs/naming_conventions.md
```diff
@@ -0,0 +1,47 @@
...
+
+Additional Arnold attributes on native USD nodes (geometry, lights, etc.) need to be in the `arnold` namespace:
```
frenchdog: Should be `native USD prims`

docs/naming_conventions.md
```diff
@@ -0,0 +1,47 @@
...
+}
+```
+
+Primvars will be converted as user data on the Arnold nodes, at the exception of those in the arnold: scope that will be translated as built-in attributes. 
+In the above example "int primvars:arnold:max_bounces = 99" will be equivalent to "int arnold:max_bounces=99"
+
+
+## Arnold Shaders
+
+Arnold shaders are supported as UsdShade Shader primitives. The `info:id` attribute should be the Arnold shader name, prefixed by the `arnold` namespace. The shader parameter are placed in the `inputs` scope.
```
frenchdog: `as UsdShade Shader prims`

PR #38, opened 2019-11-19
Invert smooth normals

### LINKED ISSUE #24: Invert smoothing default
The default smoothing follows the Arnold default (off). We should invert by default like in all plugins.

sebastienblor:
- Set the default value of parameter smoothing to ON, as in all plugins.

Fixes #24

**Additional context**


sebastienblor:
I will close this pull request, since it's conflicting with other changes, and not actually doing what's requested in #24

PR #42, opened 2019-11-20
Implement the viewport API in the USD procedural #41

### LINKED ISSUE #41: [Procedural] Implement the procedural viewport API
In Arnold 6, procedurals can define a  new function procedural_viewport that determines what is displayed in the viewport. This ticket is about implementing this function in an efficient way, by just loading what is needed

sebastienblor:
This PR implements the new procedural viewport API being introduced in Arnold 6.
It will only parse the meaningful primitives and attributes and convert them to a simplified data for efficient viewport display.

Fixes #41


sebastienblor:
Note that this would only compile when building against Arnold 6, so we might want to include this code only for Arnold 6 and up (is there a define for arnold versions ?)

sirpalee:
Yep!  The arnold_usd.h gets generated before any build happens and it'll contain the USD and Arnold versions used to build, and also provides some nice macros.

If you check `render_delegate/hdarnold.h` you can see how I'm using it.

You could do something like:

```
#pragma once

#include "../usd_arnold.h"

#if USED_ARNOLD_VERSION_GREATER_EQ(6, 0)
#define ARNOLD_USE_VIEWPORT_API
#endif
```

I personally prefer using nicely named macros for guarding features, instead of using `USED_ARNOLD_VERSION_GREATER_EQ` directly. This way it's clearer which feature turns certain parts of the code on/off.

sebastienblor:
Thanks ! Added these version guards so that we don't get compilation errors with Arnold 5, I made it so that the arnold_usd.h file is also generated for the procedural.

sebastienblor:
Btw, I didn't mention it, but I do get proper results in Maya standins in the viewport for the different modes (boxes, points, polygons), and no "free render" is done.

Also, when the AtParamValue *params has a boolean "list" value enabled, it will parse the file as usual to create the nodes, but won't convert any attribute so it should be very fast. Plugins could use this function to get the nodes list, with their names and types. Maya is about to use this to populate the standin UI, and we won't need any "free render" to get this info.

sirpalee:
> Btw, I didn't mention it, but I do get proper results in Maya standins in the viewport for the different modes (boxes, points, polygons), and no "free render" is done.
> 
> Also, when the AtParamValue *params has a boolean "list" value enabled, it will parse the file as usual to create the nodes, but won't convert any attribute so it should be very fast. Plugins could use this function to get the nodes list, with their names and types. Maya is about to use this to populate the standin UI, and we won't need any "free render" to get this info.

That's cool! Should we add some screenshots and explanations to the documentation? It could be a good idea to start application-specific documentation, mention the features and some of the use cases there. @kikou what do you think?

BigRoy:
I would love to start playing with this in maya together with `maya-usd` but am kind of lurking around until I saw some documentation popping up on how to get started. If anyone can chime in and set up some very rudimentary getting started documentation I would be very happy! :)

So :+1:.

translator/reader/registry.cpp
```diff
@@ -114,6 +114,7 @@ void UsdArnoldReaderRegistry::registerReader(const std::string &primName, UsdArn
     _readersMap[primName] = primReader;
 }
 
+#ifdef ARNOLD_USE_VIEWPORT_API
```
kikou: Why introduce a new `ARNOLD_USE_VIEWPORT_API` macro when we could just check the Arnold version is above 6?

`#if AI_VERSION_ARCH_NUM >= 6`

sirpalee: Introducing a named macro for each feature we want to enable or disable based on a dependency's version makes the code more readable, and it's easier to see which segment belongs to which feature.

sebastienblor: I replaced the macro `ARNOLD_USE_VIEWPORT_API` by a test on `AI_VERSION_ARCH_NUM` as proposed. Added the comments on why we do so

Issue #44, opened 2019-11-21
[Render Delegate] Move houdini shader translation module and LOP .ds files to HtoA


jhodgson:
Let's move the _husdshadertranslators_ python module, and the _HdArnoldRendererPlugin*.ds_ files to HtoA, and so remove them from the _arnold_usd_ repository.

BigRoy:
Do you have a specific link on where one could test/find this in `HtoA`? Do you currently need to be signed up for beta? Or is it available to give it a test run?

PR #45, opened 2019-11-21
Move Houdini-specific Python and dialog code to HtoA

jhodgson:
See #44 and htoa#1329
* Remove `husdshadertranslators` python module
* Remove dialog scripts for lights, geometry and viewport
* Fix the build now that third_party scripts don't need installing


jhodgson:
Done, created a separate pull request for that.

kikou:
Resolves #44

BigRoy:
Shouldn't this also have removed the `BUILD_HOUDINI_TOOLS` option in `SConstruct` and the related `#ifdef BUILD_HOUDINI_TOOLS` that's defined in some files like the `render_delegate/volume` and `render_delegate/render_delegate`? Or is that somehow unrelated?

PR #46, opened 2019-11-21
Use correct python library for windows houdini build

jhodgson:
Fix windows build when linking against houdini's python (See #47)


jhodgson:
We could make a new variable like _PYTHON_LIB_NAME_SUFFIX_ instead, so we don't break the USD build, setting this to _-mt_ for the houdini windows case.

If we change to system.IS_WINDOWS we should do it for previous occurrences in the file as well.

sirpalee:
> We could make a new variable like _PYTHON_LIB_NAME_SUFFIX_ instead, so we don't break the USD build, setting this to _-mt_ for the houdini windows case.
> 
> If we change to system.IS_WINDOWS we should do it for previous occurrences in the file as well.

Sounds good to me, let's go with this. We could add an example `custom.py` to `building.md` while documenting the new switch there.

jhodgson:
We can in fact set `BOOST_LIB_NAME='hboost_%s-mt'` instead of `BOOST_LIB_NAME='hboost_%s'` in the windows custom.py, so we don't need an additional option.

PR #50, opened 2019-11-21
Set subdiv_iterations to zero by default #49

### LINKED ISSUE #49: [Procedural] Subdivision shouldn't be enabled by default
The procedural checks if a mesh has subdivision by invoking the function `UsdGeomMesh::GetSubdivisionSchemeAttr()` , which returns catmull-clark by default. 

The problem is that Arnold has `polymesh.subdiv_iteration=1` by default. The combination of these 2 defaults causes all meshes to always appear with a 1-level catmull-clark subdivision when rendered with the procedural, which should be changed

sebastienblor:
This pull request sets `polymesh.subdiv_iterations` to zero by default, as it is done in the render delegate

Fixes #49


sebastienblor:
Now I'm not sure of how to get the proper subdivision value from a USD mesh.
I see that the render delegate calls `GetDisplayStyle` that ends up returning a default `HdDisplayStyle` with `refineLevel=0`.

So maybe that just setting subdivision iterations to 0 (thus disabling subdivs) is the most consistent approach in a first step ?

sirpalee:
The usdGeom/mesh schema has all the subdiv related parameters (you can see how these are converted to Arnold parameters by looking at `render_delegate/mesh.cpp` and the handling of the subdiv tags. The subdivTags is a Hydra struct with parameters equivalent to the usdGeom/mesh schema attributes.

sebastienblor:
Yes, I had checked `render_delegate/mesh.cpp` and saw some subdiv tags (for creases, etc...) that I'm not translating yet in the procedural. That would surely deserve another ticket. But for the subdiv iterations, when I follow the code, I end up with a hardcoded level of 0 coming from a default `HdDisplayStyle`. I'm wondering if this is expected to be some global option in the display, to subdivide the whole scene ?

This means that all the subdiv parameters that we're translating are actually useless, since we end up with 0-level subdiv anyway.

I guess we should start with this behaviour in the procedural, as currently in the delegate. And we'll have to find a consistent solution for both in the future.

sirpalee:
> I guess we should start with this behaviour in the procedural, as currently in the delegate. And we'll have to find a consistent solution for both in the future.

Sure, sounds good to me. Could you please add a new ticket for investigating the handling of `subdiv_iterations`?

PR #58, opened 2019-11-24
Removing USD_HAS_PYTHON_SUPPORT flag.

### LINKED ISSUE #5: [Build] Remove USD_HAS_PYTHON_SUPPORT and check for PXR_PYTHON_SUPPORT_ENABLED in pxr/pxr.h instead
The build system requires setting up the USD_HAS_PYTHON_SUPPORT to know if USD was built with python support or not. This is tracked in the installed headers, so we should be able to check for this during parsing the USD version.

sirpalee:
- Removing the USD_HAS_PYTHON_SUPPORT flag from scons and using the `pxr.h` header to decide if USD was built with python support enabled.

Fixes #5


sebastienblor:
LGTM

PR #64, opened 2019-11-25
Fix #63: Define USD_HAS_UPDATED_COMPOSITOR for USD 19.11

### LINKED ISSUE #63: USD_HAS_UPDATED_COMPOSITOR is wrong with USD 19.11
**Describe the bug**

Compiling fails to build against USD 19.11 with this error:

```
render_delegate\render_pass.cpp(177): error C2660: 'pxrInternal_v0_19__pxrReserved__::HdxCompositor::UpdateColor': function does not take 3 arguments
D:\dev\usd\build\include\pxr/imaging/hdx/compositor.h(66): note: see declaration of 'pxrInternal_v0_19__pxrReserved__::HdxCompositor::UpdateColor'
```

The built usd version is: `0.19.11`. Which according to the [merge release of 19.11](https://github.com/PixarAnimationStudios/USD/tree/4b1162957b2ad219e1635c81de8343be490e41fc) contains the [updated `UpdateColor`](https://github.com/PixarAnimationStudios/USD/blob/4b1162957b2ad219e1635c81de8343be490e41fc/pxr/imaging/lib/hdx/compositor.h#L66) arguments including `HdFormat format`.

As such [this](https://github.com/Autodesk/arnold-usd/blob/master/render_delegate/render_pass.cpp#L174) falsely enters the `else` statement due to [`USD_HAS_UPDATED_COMPOSITOR` only being defined for USD 19.12 and up.](https://github.com/Autodesk/arnold-usd/blob/dd40ebbed8524ed1052862114fc942d94decffdc/render_delegate/hdarnold.h#L35)

**To Reproduce**
1. Compile USD 19.11
2. Compile `arnold-usd` against it.
3. Error.

**Expected behavior**
It should compile correctly.

It continues compiling when [this line](https://github.com/Autodesk/arnold-usd/blob/dd40ebbed8524ed1052862114fc942d94decffdc/render_delegate/hdarnold.h#L35) is changed to:
```cpp
#if USED_USD_VERSION_GREATER_EQ(19, 11)
```

**Used Software Versions**
 - Arnold: `5.4.0.2`
 - USD: `0.19.11
 - Compiler: Visual Studio 2017
 - OS: Windows 10

BigRoy:

- Define `USD_HAS_UPDATED_COMPOSITOR` for USD 19.11+ instead of USD 19.12+

As far as I can tell the updated compositor API is already part of the [official 19.11 release](https://github.com/PixarAnimationStudios/USD/commit/4b1162957b2ad219e1635c81de8343be490e41fc) as described in #63. Similarly it's also in the [USD 19.11 archived release](https://github.com/PixarAnimationStudios/USD/releases/tag/v19.11)

Fixes #63

### DIFF
--- a/tools/utils/build_tools.py
+++ b/tools/utils/build_tools.py
@@ -131,7 +131,7 @@ def get_usd_header_info(usd_include_dir):
     r = re.search('#if ([01]).#define PXR_PYTHON_SUPPORT_ENABLED', pxr_h, re.DOTALL)
     if r:
         HAS_PYTHON_SUPPORT = r.group(1) == '1'
-        
+
     # Detect updated Compositor
     compositor_h_path = os.path.join(usd_include_dir, "pxr", "imaging", "hdx", "compositor.h")
     if os.path.exists(compositor_h_path):
@@ -144,7 +144,7 @@ def get_usd_header_info(usd_include_dir):
                 # Assume new updated compositor UpdateColor() function
                 # introduced since 19.11 but not present in all, see #64
                 HAS_UPDATED_COMPOSITOR = True
-    
+
     return {
         'USD_VERSION': '.'.join(VERSION),
         'USD_HAS_PYTHON_SUPPORT': HAS_PYTHON_SUPPORT,
--- a/tools/utils/build_tools.py
+++ b/tools/utils/build_tools.py
@@ -117,6 +117,7 @@ def get_arnold_version(arnold_include_dir, components = 4):
 def get_usd_header_info(usd_include_dir):
     VERSION = [''] * 3
     HAS_PYTHON_SUPPORT = False
+    HAS_UPDATED_COMPOSITOR = False
 
     pxr_h = open(os.path.join(usd_include_dir, 'pxr', 'pxr.h'), 'r').read()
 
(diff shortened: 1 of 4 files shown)


BigRoy:
I see, that does complicate things. Why ship with a non-final version of `0.19.11`? That's bound to create confusion. Time for [vfx platform](https://vfxplatform.com/) to make some statements regarding `USD` versions.

So we would match it similar to #58 with something like this:
```python
import os
import re

# usd_include_dir = root to built USD/include folder
compositor_h_path = os.path.join(usd_include_dir, "pxr", "imaging", "hdx", "compositor.h")
compositor_h = open(compositor_h_path, "r").read()

match = re.search("UpdateColor\(([^)]*)", compositor_h, re.DOTALL)
if not match:
    raise RuntimeError("Unable to find UpdateColor() in %s" % (compositor_h_path))

args = match.group(1).split(",")
if len(args) == 4 and args[3].strip() == "Hdformat format":
    HAS_UPDATED_COMPOSITOR = True
else:
    # Assume it's the old compositor
    HAS_UPDATED_COMPOSITOR = False
```

sirpalee:
Sorry, I was inaccurate in my previous comment, I shouldn't have used the word "shipped". No app was shipped with a non-finished 0.19.11, but at the time of testing the final 0.19.11 was not available yet. That created a weird case, where we had to work with 0.19.11 with the change and some 0.19.11 without the change.

The vfx platform is outside the scope of this project, so let's focus on parsing the hdx compositor header.

As for the change, it should be added to `tools\utils\build_tools.py` / `get_usd_header_info`, and in case of a missing header (for whatever reason, like no imaging was built), it should just default to `False` instead of raising an exception.

BigRoy:
This now has all the commits, I missed a file on my commit before. This works on my end to correctly detect the updated compositor. 

Could someone give this a go against an older version of USD, maybe a 19.11 that didn't have the updated compositor as described above?

tools/utils/build_tools.py
```diff
@@ -130,10 +131,23 @@ def get_usd_header_info(usd_include_dir):
     r = re.search('#if ([01]).#define PXR_PYTHON_SUPPORT_ENABLED', pxr_h, re.DOTALL)
     if r:
         HAS_PYTHON_SUPPORT = r.group(1) == '1'
-
+        
+    # Detect updated Compositor
+    compositor_h_path = os.path.join(usd_include_dir, "pxr", "imaging", "hdx", "compositor.h")
```
sirpalee: This needs to work if compositor.h does not exist and default to False. `get_usd_header_info` runs for all builds, even when the render delegate is not built, or when USD does not have its imaging component built. We don't need to print any warnings, so a simple try/except with pass for the exception should be enough.

BigRoy: Should be fixed with https://github.com/Autodesk/arnold-usd/pull/64/commits/5c41639839eae9f7fb5aa8260408e8d92a3f8449.

sirpalee: Yep, this looks good, thanks!

sirpalee:
> Could someone give this a go against an older version of USD, maybe a 19.11 that didn't have the updated compositor as described above?

I tested your changes against 0.19.07, 0.19.11 and a USD build without the imaging headers on OSX.  Everything compiles fine, thanks for the pull request!

PR #81, opened 2019-11-28
Create a new Xform cache for animated xforms

### LINKED ISSUE #80: Regression in animated xforms
**Describe the bug**
Since we adressed #51 , we now use a `UsdGeomXformCache` in order to get the world matrix of the geometries. But this now prevents animated xforms from being rendered properly

**To Reproduce**
Run `abuild test_0018`

sebastienblor:
This PR is meant to address the regression about animated xforms in #80.
I see different solutions to address this issue, and I'm not totally sure which one is best :

- 1 )  every time one primitive has keys on a different frame than the current one, create one `UsdGeomXformCache` just to compute the world matrix and delete it. 

- 2 ) Create on-demand one `UsdGeomXformCache` per requested frame. The `UsdReader` would store a map of xform caches, and would create them dynamically when needed.

- 3 ) Change the time of the `UsdGeomXformCache` every time we need to get a new matrix.

///////////////////////////// 

In this commit I went for the simplest solution 1), but I'm fine with implementing a different solution if better.

3 ) would be very simple to implement but changing the cache time will clear its data under the hood, so if there are a lot of animated xforms in the scene we might loose the benefits from the cache.

If there are a lot of animated xforms in the hierarchy that have values at the same keys, it would be more efficient to store each xform cache in a map as in 2), so that these intermediate matrices aren't computed over and over. But if the xforms have keys at different time values, we could end up with a lot of xform caches being computed, and maybe it could get overkill ?

Fixes #80

### DIFF
--- a/translator/reader/utils.cpp
+++ b/translator/reader/utils.cpp
@@ -44,13 +44,7 @@ static inline void getMatrix(const UsdPrim &prim, AtMatrix &matrix, float frame,
 {
     GfMatrix4d xform;
     bool dummyBool = false;
-    UsdGeomXformCache *xformCache = reader.getXformCache();
-
-    // The frame is different than the one used for the xform cache, we're 
-    // probably dealing with animated xforms. Let's reset our pointer in 
-    // this case, so that we generate a new one below. 
-    if (xformCache && std::abs(frame - (float)xformCache->GetTime().GetValue()) > AI_EPSILON)
-        xformCache = NULL;
+    UsdGeomXformCache *xformCache = reader.getXformCache(frame);
     
     bool createXformCache = (xformCache == NULL);
     if (createXformCache)
--- a/translator/reader/utils.cpp
+++ b/translator/reader/utils.cpp
@@ -39,14 +39,19 @@
 
 PXR_NAMESPACE_USING_DIRECTIVE
 
+
 static inline void getMatrix(const UsdPrim &prim, AtMatrix &matrix, float frame, UsdArnoldReader &reader)
 {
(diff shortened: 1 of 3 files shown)


translator/reader/utils.cpp
```diff
@@ -39,14 +39,19 @@
...
     bool dummyBool = false;
     UsdGeomXformCache *xformCache = reader.getXformCache();
 
-    // The reader should have a xform cache, if not let's create one 
-    // just for this purpose (not optimized)
+    // The frame is different than the one used for the xform cache, we're 
+    // probably dealing with animated xforms. Let's reset our pointer in 
+    // this case, so that we generate a new one below. 
+    if (xformCache && std::abs(frame - (float)xformCache->GetTime().GetValue()) > AI_EPSILON)
+        xformCache = NULL;
```
sirpalee: Please use nullptr instead of NULL.

sebastienblor: well, if one cache per frame is where we want to go, I can just do it right away. I just wasn't sure that accumulating the xform caches per primitive key was the best thing to do

sebastienblor:
The last commit stores instead a map of xform caches, for each of the animated keys, as proposed in solution 2).

translator/reader/utils.cpp
sebastienblor: this piece of code with `NULL` disappeared in the last commit

sebastienblor:
In the last commit I'm making the xform cache evaluation thread-safe, by using an arnold mutex (`AtCritSec`).

PR #93, opened 2019-12-04
Supporting building against prefixed USD libraries.

### LINKED ISSUE #48: Building against Houdini Libraries
Hi !

Amazing to see this repo go live !
I've been successfully able to compile against Houdini's shipped libraries with a few modifications and was wondering if these could become part of the main build system.

First, USD libs in Houdini have the libpxr_ prefix, so it would be nice to have a build option like BOOST_LIB_NAME for USD.
Second, it seems like the USD_HAS_UPDATED_COMPOSITOR doesn't set correctly when building against Houdini's USD libs. Setting it up manually has fixed it in my experience.

Thanks !

sirpalee:
- Adding a new USD_LIB_PREFIX flag.
- Adding example custom.py files to the documentation.
- Extending building.md about using abuild and configuring the build.
- Fixing a typo in the render delegate that turns on -fPIC on windows.
- Adding a warning about the python app shortcut on Windows 10.
- Adding dl to the dependencies when building the render delegate on Linux.
- Adding a new utility function for SCons scripts, link_usd_libraries that handles adding dependencies to builds.

Fixes #48


sebastienblor:
The changes are looking good to me. I've tested them on windows, while building a static procedural as well as usd writer. Both worked fine, and the testsuite is ok

Issue #65, opened 2019-11-26
Some arnold attribute names are a problem for USD


sebastienblor:
**Describe the bug**
Currently, our schemas generation does some exceptions for a few attributes, whose name (or values) causes problems for USD. This happens because `usdGenSchema` creates a file `tokens.h` with all the attribute names and enum tokens, and compiles it. But some of the names can't compile.

Here's the current list :
//////////////////////////////////////////////

The attribute `namespace` (from `procedural`), will introduce this in `tokens.h` : 
```
/// UsdArnoldAlembic
const TfToken nameprefix;
/// \brief "namespace"
```
and then fails to compile : 
```
tokens.h(2992): error C2059: syntax error: 'namespace'
```
----------------------------------------

The attribute `default` (from `user_data_*` ) seems to be a problem for `usdGenSchema` as it adds an underscore : 
```
/// 
/// UsdArnoldUserDataString, UsdArnoldUserDataRgb, UsdArnoldUserDataRgba, UsdArnoldUserDataInt, UsdArnoldUserDataFloat
const TfToken default_;
```
and then fails to compile :   
```
build\windows_x86_64\msvc_opt\usd-0.19.5_arnold-6.0.1.0\schemas\source\arnoldUse
rDataRgba.cpp(125): error C2039: 'default': is not a member of 'arnoldPxrInterna
l__pxrReserved__::UsdArnoldTokensType'
```
-----------------------------------------

We have a similar problem with enum values, since they're also added as tokens in this same file.

Value `auto` (from `polymesh.subdiv_adaptive_metric`) creates this:
```
/// \brief "auto"
/// 
/// Possible value for UsdArnoldPolymeshAPI::GetArnoldSubdiv_adaptive_metricAttr(), Default value for UsdArnoldPolymeshAPI::GetArnoldSubdiv_adaptive_metricAttr(), Possible value for UsdArnoldPolymesh::GetSubdiv_adaptive_metricAttr(), Default value for UsdArnoldPolymesh::GetSubdiv_adaptive_metricAttr()
const TfToken auto;
```
which obviously fails to compile 
```
0\schemas\source\.//tokens.h(684): error C2628: 'arnoldPxrInternal__pxrReserved_
_::TfToken' followed by 'auto' is illegal (did you forget a ';'?)
```
---------------------------------------------

Our shader attribute `compare.test` has enum values like `==`, `<`, `>`, etc...
and this is a problem for `usdGenSchema`:
```
[...]

PR #98, opened 2019-12-04
Don't build typed schemas for shaders

### LINKED ISSUE #96: Don't create typed schemas for shaders
We're currently creating typed schemas for *any* arnold node type. We can get rid of this for shaders, since we're fully relying on `UsdShade` schemas.

sebastienblor:
No longer build typed schemas for shaders since we rely on `UsdShade` schemas, with the `info:id` token returning the arnold shader name.
This allows to get rid of several special cases during the schemas generation, some of which are explained in #65 

Fixes #96

**Additional c


sebastienblor:
I'm now making it so that the usd writes converts shaders to `UsdShade` schemas, in order to be consistent with the change proposed in this PR

PR #100, opened 2019-12-05
Re-rendering when visibility of an object changes.

### LINKED ISSUE #95: Hiding an item in usdView does not rerender
**Describe the bug**
When hiding something in usdView the Arnold render does not update the change. If I move the camera then the hidden object disappears, so the visibility states are probably set. Showing a hidden object updates properly though.

**To Reproduce**
1. Open anything in usdView
2. Set Arnold Hydra renderer
3. Hide an object
4. Move the camera

**Used Software Versions**
 - Arnold: 5.4.0.2
 - USD: 0.19.11
 - Compiler: MSVC 2019
 - OS: Win 10

sirpalee:
- Restarting renders when visibility is dirtied.

Fixes #95

### DIFF
--- a/render_delegate/mesh.cpp
+++ b/render_delegate/mesh.cpp
@@ -91,6 +91,7 @@ void HdArnoldMesh::Sync(
     }
 
     if (HdChangeTracker::IsVisibilityDirty(*dirtyBits, id)) {
+        param->End();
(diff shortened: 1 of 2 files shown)


sebastienblor:
The change is straight-forward, looks good

PR #104, opened 2019-12-06
Fix display settings of type int and bool on OSX

### LINKED ISSUE #103: Display settings of type int and bool not reaching arnold in OSX
Fix display settings of type int and bool on OSX.

Due to type being long long, ignore shaders, smoothing, threads, diffuse samples and more are not coming through.

jhodgson:
Fix display settings of type int and bool on OSX.

Due to type being long long, ignore shaders, smoothing, threads, diffuse samples and more are not coming through.

Fixes #103


sebastienblor:
Looks good to me.

PR #106, opened 2019-12-18
Raise a runtime error if usdGenSchema can't be found and the schema build is enabled.

### LINKED ISSUE #105: Stop build if the schemas are enabled and usdGenSchema is not available
The build should stop if the schemas are enabled, but `usdGenSchema` is not available in the PATH, to avoid confusion regarding missing schema files.

Instead of return an empty target, we should just raise an exception.

sirpalee:
- Raise a runtime error if `usdGenSchema` can't be found and the schema build is enabled

Fixes #105


sebastienblor:
Looks good

PR #110, opened 2019-12-19
Flipping polygon order if the topology orientation is left handed in the render delegate

### LINKED ISSUE #83: SSS not translating correctly
Hi !

Testing with Houdini 18.0.287, the Standard Surface SSS does not seem to be handled correctly, set the SSS weight to 1 and the object turns black.

sirpalee:
- When the topology received from Hydra is leftHanded, flip the polygon order to match Arnold's standard.

Fixes #83


sebastienblor:
This is similar to what is done in the translator, in `readGeometry.cpp`. Looks good

Issue #26, opened 2019-11-15
Add tests for Arnold to USD


sirpalee:
Add tests for the Writer command.

A good approach is to execute python scripts as part of the testsuite and compare the resulting USD file to a reference file.

sirpalee:
![](https://github.trello.services/images/mini-trello-icon.png) [Adding a test for the Arnold Writer command](https://trello.com/c/hLbdjpRL/62-adding-a-test-for-the-arnold-writer-command)

PR #128, opened 2020-01-03
Write Arnold nodes as USD builtins when possible

### LINKED ISSUE #39: Allow to convert Arnold to USD-native schemas
Currently the USD writer only converts Arnold to arnold-specific schemas (ArnoldStandardSurface, ArnoldPolymesh, ArnoldSkydomeLight, etc....).

We should allow, at least optionally if not by default, to export to USD-native schemas when it makes sense.
For example an skydome_light could exported as a Dome primitive, and we'd add the additional arnold attributes in the arnold: scope.

sebastienblor:
Modify the writer module so that it can convert some arnold nodes (geometries, lights), as USD builtin nodes, instead of always creating Arnold-specific schemas.

Fixes #39 

**Additional context**
This is a first draft, that is still missing some rarely used arnold shapes (cube, sphere, cylinder), and also material bindings. We can address this in future tickets.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,10 @@ custom*.py
 dist
 build
 install
+bin
+lib
+include
+procedural
 
 # Eclipse
 .settings
--- a/translator/writer/prim_writer.cpp
+++ b/translator/writer/prim_writer.cpp
@@ -19,6 +19,7 @@
 #include <pxr/base/gf/matrix4f.h>
 #include <pxr/base/tf/token.h>
 #include <pxr/usd/usd/prim.h>
+#include <pxr/usd/usdGeom/primvarsAPI.h>
 #include <pxr/usd/usdShade/material.h>
 #include <pxr/usd/usdShade/shader.h>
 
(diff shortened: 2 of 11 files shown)


sebastienblor:
the last commit isn't directly related to this ticket, but the install folders (bin, lib, include, procedural) needed to be added to .gitignore

sebastienblor:
The commits introduced lots of changes, that might be hard to review... The biggest refactoring was to allow to author parameters whether they're usd builting parameters, arnold-specific attributes, or arnold user data, without duplicating the conversion code. Here I'm relying on templates that can be `UsdArnoldBuiltinParamWriter` ,  `UsdArnoldCustomParamWriter` and `UsdArnoldPrimvarWriter`. 

All the attributes that weren't exported as usd builtin params are exported as arnold-specific attributes with the `arnold:` namespace. The geometries user data are exported as primvars without any namespacing.

Exporting arnold nodes as usd builtins is made controlled by an option here, but it's the default behaviour as it probably makes more sense. This PR handles polymeshes, curves, and points geometries, as well as distant, skydome, disk, quad, point and mesh lights. I'm still missing other arnold shapes like cube, sphere, cylinder, that I can eventually add to this PR. Most of all, material bindings aren't handled yet, so the shader assignments are currently exported as `arnold:shader = lambert`, but we can do another ticket for that.

I've done a few tests on input .ass files and things seem to be working as expected, but a proper testsuite as requested in #26 will become important now.

PR #132, opened 2020-01-07
Changing the type of progressive_min_AA_samples from bool to int.

### LINKED ISSUE #131: progressive_min_AA_samples is type of bool instead of int
In the Render Delegate settings, progressive_min_AA_samples has the type of bool, not int. This will lead to the minimum progressive samples not set.

**To Reproduce**
1. Open usdview and switch to the Arnold render delegate
2. See that` Progressive Min AA Samples` is a bool and progressive rendering is more sluggish.

**Expected behavior**
Progressive Min AA samples should be an integer and the default -4 should result in better user experience when rendering a simple scene.

**Used Software Versions**
 - Arnold: 6.0.1.0
 - USD: 0.19.11
 - Compiler: GCC 4.8.5
 - OS: CentOS 7

sirpalee:
Changing the type of progressive_min_AA_samples from bool to int.

- Changing the type of progressive_min_AA_samples from bool to int.

Fixes #131

### DIFF
--- a/render_delegate/config.h
+++ b/render_delegate/config.h
@@ -113,7 +113,7 @@ public:
 
     /// Use HDARNOLD_progressive_min_AA_samples to set the value.
     ///
-    bool progressive_min_AA_samples;
+    int progressive_min_AA_samples;
 
     /// Use HDARNOLD_enable_adaptive_sampling to set the value.
     ///


sebastienblor:
LGTM

PR #136, opened 2020-01-08
Checking for root layer validity when loading the file.

### LINKED ISSUE #74: The procedural crashes when rendering an ass file with a missing relative USD file path
**Describe the bug**
I was rendering one of the testsuit ass files that uses a relative usd file dir, but the render was launched from a different directory. And this is what I get:
--
Coding Error (secondary thread): in Open at line 992 of build\USD\pxr\usd\lib\usd\stage.cpp -- Invalid root layer
00:00:07   169MB ERROR   |   signal caught: error 80000003
--

**To Reproduce**
Just start a render from another dir.

**Expected behavior**
Perhaps a warning message.

**Used Software Versions**
 - Arnold: 5.4.0.2
 - USD: 0.19.11
 - Compiler: MSCVC 2019
 - OS: Win 10

sirpalee:
- When the reader loads the file, we are returning early if the rootLayer is nullptr.

Fixes #74

### DIFF
--- a/translator/reader/reader.cpp
+++ b/translator/reader/reader.cpp
@@ -62,6 +62,10 @@ void UsdArnoldReader::read(const std::string &filename, AtArray *overrides, cons
     }
 
     SdfLayerRefPtr rootLayer = SdfLayer::FindOrOpen(filename);
+    if (rootLayer == nullptr) {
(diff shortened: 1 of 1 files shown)


sirpalee:
@sebastienblor Good point! Added an error message and updated the pull request.

PR #138, opened 2020-01-08
Create a default shader that reads displayColor

### LINKED ISSUE #137: Procedural should assign a default shader showing displayColor
Currently, when no material binding is found on a geometry, it won't have any shader assigned and the objects will be rendered with arnold default shader (fully white). Hydra viewers show instead a shader which color is eventually determined by the primvar `displayColor`. 

Every procedural could eventually create a default shading tree with a `standard_surface`  with `base_color` linked to a `user_data_rgba` shader. This last shader would look for the user data `displayColor`. Since primvars are automatically converted as user data, this would give the expected behaviour.

Note however, that as soon as a material is assigned, we will loose the display Color since we need to follow what the shading tree dictates. If we think that this would be confusing for users, we might prefer to keep the current behaviour.

sebastienblor:
USD geometries that don't have any material binding will be assigned a default shader which color corresponds to the primvar `displayColor`. We do this by linking a `user_data_rgb` to `standard_surface.base_color`, and by setting the input user data name to `displayColor`.

Every USD procedural would create and own its default shader and eventually assign it to its children. In this PR I'm only creating it if it's needed. 

Of course, since I'm creating a shader (2 shaders actually), I need to give them a name, which means that we're up for a round of usual naming debates. 
In all my wisdom I went for `_default_arnold_shader` for the standard surface, and `_default_arnold_shader_color` for the user data node, but I don't have any strong opinion at all.

Fixes #137


sebastienblor:
In the second commit I'm just ensuring that the default shader is properly reset when translator's nodes are cleared

sebastienblor:
I updated the test scenes and added `test_0021` for this feature.

Note that the default shader is now a `standard_surface` that is affected by the scene lighting, whereas before we were using the default arnold `utility` shader that did a flat shading that ignores the lights.

This means that by default, if you open a scene with no lighting you won't see anything, whereas before you did see the objects flat-shaded. So if we prefer, I could replace this default shader by a `flat` shader instead.

sirpalee:
Thanks for the updates, all the changes look good and everything runs perfectly locally.

PR #142, opened 2020-01-10
Parameter duplication fix

kreppene:
The writeArnoldParameters only exports parameters on the first occurrence of an parameter name, resulting in missing paramaters on next occurrence  of a parameter name,

the changes inn this pull request checks if (nodeName+paramName) exists before exporting parameters

fixed missing paramters on next occurrence of a paramater name

**Additional context**
Screenshot_1


sebastienblor:
Good catch, that's indeed a bug.
I think I would have preferred to clear the list of exported attributes every time a new node is written, but then we'd either have to do this in every implementation of `write` , or we'd have to ensure we clear the list every time we call `primWriter->write`.

So maybe storing the node name in the map as proposed here is safer. Could we just add a dot between the name and the attribute ? Something like
`std::string fullParamName = std::string(AiNodeGetName(node)) + std::string(".") + std::string(paramName);`

Otherwise there could still be some naming conflicts in (very unlikely) corner cases.
Thanks

kreppene:
Thanks, @sebastienblor ,

yes this list can get very long i guess, so its not the perfect solution.
In what cases is this check actually needed?

sebastienblor:
The list is needed to keep track of the attributes that were exported to USD builtin attributes, so that we don't re-export them as arnold-namespaced attributes.

We don't strictly need what's in prim_writer.cpp, line 596 
`_exportedAttrs.insert(attrs.begin(), attrs.end()); `
which is also adding to this list the arnold-specific attributes, but it was just for safety. 

But still it's a shame that we concatenate attributes from different nodes, so maybe just clearing this list at every new node would be better.

Since there have already been several commits in this PR that could be reverted with this other solution, do you mind if I close this PR and create another one with just a new call that clears this list at every node ? 

Thanks

sebastienblor:
I created an issue #145 for this problem and I'll shortly post a new PR, then we'll discuss which one we go for.

sebastienblor:
@kreppene  I made a pull request #146 , please let me know if it works for you. If it does, I'll close this one.
Thanks

kreppene:
@sebastienblor  #146 works fine for me, go ahead and close this one 

Thanks

PR #148, opened 2020-01-10
Link linux builds against pthread and dl

### LINKED ISSUE #147: Fix build issue on linux
**Describe the bug**
We're currently having issues related to pthreads and tbb when building the procedural on linux.

sebastienblor:
Link the procedural against `pthread `and `dl` . This change was proposed by Oscar Anson in order to fix the current linux builds.

Fixes #147

### DIFF
--- a/cmd/SConscript
+++ b/cmd/SConscript
@@ -61,6 +61,8 @@ elif local_env['USD_BUILD_MODE'] == 'static':
             'libusd_m', 
             'tbb',
         ]
+        if system.IS_LINUX:
+            local_env.Append(LIBS = ['dl', 'pthread'])
 else:  # shared libs
     usd_libs = [
         'sdf',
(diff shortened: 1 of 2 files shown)


sebastienblor:
We need to determine if this change needs to be done only for the procedural, or more globally in `Sconstruct`. Also if it's only for static builds.

sirpalee:
I'm doing the same thing in the cmake build for the static/Linux builds. `target_link_libraries(translator PUBLIC dl pthread)`

When USD is static, we only should build the procedural and the writer, I think it's easier to add it to both, no need to make it global.

sebastienblor:
In the latest commit I'm also doing this in `cmd/Sconscript` for the `arnold_to_usd` command. 
We seem to need this only for static linux builds. Even though it wouldn't harm to do it for all build modes I'm just doing it in static mode for now

PR #159, opened 2020-01-22
Supporting SdfAssetPath on shaders

### LINKED ISSUE #158: Support SdfAssetPaths as Arnold shader parameters (procedural)
**Describe the bug**
Hello everyone!
The procedural does not seem to currently handle `SdfAssetPaths` as valid parameters for Arnold shaders, possibly preventing them from being resolved by a custom asset resolver. 
When used in lieu of strings to define map filepaths, they are not sent to Arnold which then attempts to create textures with empty paths.
As always, thank you for your help.

**To Reproduce**
1. Unzip the [archive](https://github.com/Autodesk/arnold-usd/files/4093484/filepath_as_asset.zip)
2. Run the following from the archive's directory:
`
kick filepath_as_asset.ass
`

**Expected behavior**
A rendered image showing a textured quad, similar to what `usdview` would show. The lack of lighting should not impact the `image` shader being used in the scene.

**Used Software Versions**
 - Arnold: 6.0.1.0
 - USD: 0.19.11
 - Compiler: GCC 4.8.5
 - OS: CentOS Linux release 7.6.1810

sirpalee:
- Checking for SdfAssetPath when the reader encounters an AI_TYPE_STRING parameter
- Updating test_0015 to include a texture and the use of an asset path

Fixes #158


sebastienblor:
The change looks good. 
Is there a situation where we also need to author this in the writer ?

PR #179, opened 2020-01-29
Refactor the parallel reader

### LINKED ISSUE #170: Make parallel nodes creation thread-safe in the translator
**Describe the bug**
The translator has an option to read the scenes in multi-thread. However, this is giving random results, surely meaning that something isn't thread-safe. 

The nodes creation doesn't seem to be locked which is wrong. Calling `AiNode()` in parallel should be fine in the arnold side, but there could be multiple threads adding simultaneously an element to the vector `_nodes`.

### LINKED ISSUE #178: Refactor the parallel reader
Loading / Converting a scene file always end up needing a special treatment to handle connections between nodes. In order to set the link between 2 nodes both must exist, which is not always the case while loading it, depending on the order in which they're defined in the file. This gets even more complicated if the loading is done in parallel.
This is what happens when the translator converts the USD stage into an arnold scene.

There are different ways to handle this, here are 3 typical solutions :

1) when we create node A we see that it has a connection with node B (not created yet). At this point we force the creation of node B before we continue, so that we can create the link. We must then be sure that we won't create B twice when we get to it later on. This might seem simple but it means that the threads must be constantly synchronizing to check if a node already exists before creating it, and be sure that it's not being exported simultaneously by concurrent threads

2) We first create all the nodes to populate the scene, and then we do a second pass to set all the attributes and their connections. If we want to do this in parallel, we need to join the threads after the first pass, so that the second only happens when the first has completely finished. An advantage of this solution is that each thread can be totally independent, and we just need to synchronize them once between the passes. But it also means that we do 2 passes over the whole scene which can be slower. The code also needs to do a distinction between the node creation and conversion.

3) Instead of setting the connections between nodes while loading each of them, we stack/serialize these connections so that they're processed after the scene was fully loaded. In parallel, each thread can be independent with this solution, they just need to synchronize it after the whole scene was created, and before the connections are applied. The advantage of this solution over 2) is that it doesn't require to loop over the whole scene twice. The second stage is just a loop over serialized connections. De-serializing the connections in the last step also has a cost, but possibly lower than looping twice over the scene.

So far, I've been following method 2) in the translator, but since the loading times are important for us, I'm wondering if we shouldn't consider solution 3).

sebastienblor:
This PR refactors the whole parallel reader in the translator. As explained in #178 , this replaces solution 2) by 3) , where the connections between nodes are stacked to be applied when the scene finished loading.
Also, in the past for parallel loading we were creating a new `UsdArnoldReader` per thread and we merged them at the end, which was a bit confusing (we didn't even realize this when implementing the xform cache). Here I introduce a reader "context"  (per-thread), that we pass to the different functions instead of the reader itself, which is now unique. Each context has its own list of nodes that are merged when the process finishes. The xform cache is now handled in the context since it's not meant to be used by concurrent threads (see #170 ).

This PR solves the hangs that we were having when using more than 1 thread for the usd procedural. If I force it to be set to 10 by default, the whole testsuite succeeds. However, I still have a small randomness in the kitchen scene (test_0016) with a lot of threads on a couple of objects, which I still need to debug.

Note that for array connections, instead of having a vector of strings for the target (one for each array element) I'm using a serialized string with a separator being ' '. The reason for this is just for memory allocation, since we push to a vector of `UsdArnoldRenderContext::Connection`. If each connection element has a vector of strings, the memory allocation / reallocation could be expensive. But we could simplify this if it's not worth doing it.

Fixes #178 
Fixes #170


sirpalee:
@sebastienblor could you please update the pull request? After merging in the latest Pull Requests there are some conflicts.

sebastienblor:
I merged the master branch and fixed the conflicts. Note that #153 will need some changes if this PR gets merged.

sebastienblor:
I would create a separate pr for the kitchen in multithread, as it shouldn't affect the general refactoring

PR #181, opened 2020-01-30
Updating test-0014 reference image.

### LINKED ISSUE #155: Issue with the render context priority (procedural)
**Describe the bug**
Hello everyone.

Maybe it is not an actual bug, but the priority with which the procedural sorts a UsdShadeMaterial's resolved surface outputs for different render contexts seems counter-intuitive. A material `outputs:surface` property takes precedence over a more specific `outputs:arnold:surface` property. 

Karma behaves the other way for instance: using the material's `outputs:surface` as a fallback when `outputs:karma:surface` is undefined instead.

The involved code seems to be [here](https://github.com/Autodesk/arnold-usd/blob/47711e8839a230f1210722653a698911ff09357b/translator/reader/utils.cpp#L317-L319).

**To Reproduce**
1. Unzip the [archive](https://github.com/Autodesk/arnold-usd/files/4087357/context_issue.zip).
2. Run the following from the archive's directory
```
kick context_issue.ass
```

**Expected behavior**
The cube appears blue, whereas an Arnold-dedicated surface output is defined on its bound material and should make it appear red instead. Removing the property `/my_material/outputs:surface` enables the procedural to use the surface source resolved for the Arnold context.

**Used Software Versions**
 - Arnold: 6.0.1.0
 - USD: 0.19.11
 - Compiler: GCC 4.8.5
 - OS: CentOS Linux release 7.6.1810

sirpalee:
- Updating test-0014 reference image.

Fixes #155


sirpalee:
Note, the image does not need to be updates. I got confused for a second as my local testsuite was rendering green. I'm going to keep looking into the issue and keep the original ticket open as it might be something related to the new USD version I'm using.

PR #183, opened 2020-01-30
Export material bindings in the writer

### LINKED ISSUE #163: USD writer should create USD materials
Currently, the USD writes saves material bindings as arnold-specific attributes, e.g.
`arnold:shader = ["my_shader"]`

and it saves directly a link between a mesh and a shader, whereas USD expects a material between the mesh and the shader. This isn't supported in Hydra render delegate and should be changed.

sebastienblor:
The usd writer now exports arnold shader assignments by connecting a `UsdShadeMaterial` primitive, on which both the surface shader and displacement are linked.

Since the material node doesn't exist in the arnold scene, we need to give it a unique name. Here I'm appending both the name of the surface shader and the eventual displacement, so that every combination surface + displacement is unique, and so that we share the same material between geometries.

While doing it, I saw that we were sometimes doing the export of a node multiple times uselessly (in case of shader links or node connections), so I've enforced it here. The list of exported attributes was also cleared by mistake when setting connections, this is now fixed.

I am not handling the geometry subset assignments yet. We can add the logic to this current PR if it's not too complicated to translate.

Fixes #163

### DIFF
--- a/translator/writer/prim_writer.cpp
+++ b/translator/writer/prim_writer.cpp
@@ -21,14 +21,13 @@
 #include <pxr/usd/usd/prim.h>
 #include <pxr/usd/usdGeom/primvarsAPI.h>
 #include <pxr/usd/usdShade/material.h>
+#include <pxr/usd/usdShade/materialBindingAPI.h>
 #include <pxr/usd/usdShade/shader.h>
 
 #include <cstdio>
 #include <cstring>
 #include <string>
 #include <vector>
-
-#include <pxr/usd/usdShade/materialBindingAPI.h>
 
 //-*************************************************************************
 
@@ -363,14 +362,14 @@ void UsdArnoldPrimWriter::writeNode(const AtNode *node, UsdArnoldWriter &writer)
 {
     // we're exporting a new node, so we store the previous list of exported 
     // attributes and we clear it for the new node being written
-    std::unordered_set<std::string> prevExportedAttrs = _exportedAttrs;
-    _exportedAttrs.clear();
+    decltype(_exportedAttrs) prevExportedAttrs;
+    prevExportedAttrs.swap(_exportedAttrs);
 
     // Now call the virtual function write() defined for each primitive type
     write(node, writer); 
 
     // restore the previous list (likely empty, unless there have been recursive creation of nodes)
-    _exportedAttrs = prevExportedAttrs;
+    _exportedAttrs.swap(prevExportedAttrs);
 }
 /**
  *    Get the USD node name for this Arnold node. We need to replace the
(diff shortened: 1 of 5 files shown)


translator/writer/prim_writer.cpp
```diff
@@ -652,4 +658,55 @@ void UsdArnoldPrimWriter::writeMatrix(UsdGeomXformable &xformable, const AtNode
...
+    std::string shaderName = (shader) ? UsdArnoldPrimWriter::getArnoldNodeName(shader) : "";
+    std::string dispName = (displacement) ? UsdArnoldPrimWriter::getArnoldNodeName(displacement) : "";
+    
+    // The material node doesn't exist in Arnold, but is required in USD, 
+    // let's create one based on the name of the shader plus the name of 
+    // the eventual displacement. This way we'll have a unique material in USD
+    // per combination of surface shader + displacement instead of duplicating it
+    // for every geometry.
+    std::string materialName = "/materials";
+    materialName += shaderName;
```
sirpalee: The returned `shaderName` and `dispName` will both start with a `/` creating a child in the `/materials` scope. I think that's good because it allows us to organize the materials nicely.

However, it'll create a weird hierarchy inside the materials, especially when displacement shaders are used. For example, our scene has two shaders, `ShaderA` (surface) and `ShaderB` (displacement). If we have two meshes, `A`, to which only `ShaderA` is assigned, and `B` to which both `ShaderA` and `ShaderB` is assigned, we'll end up with a layout like this:

```
def "materials"
{
    def Material "ShaderA"
    {
        # ...
    }

   def "ShaderA_disp_"
   {
        def Material "ShaderB"
        {
             # ...
        }
   }
}
```

Is this intentional?

translator/writer/writer.h
```diff
@@ -49,11 +49,14 @@ class UsdArnoldWriter {
...
+    bool isNodeExported(const std::string &name) { return _exportedNodes.find(name) != _exportedNodes.end();}
+
 
 private:
     const AtUniverse *_universe;        // Arnold universe to be converted
     UsdArnoldWriterRegistry *_registry; // custom registry used for this writer. If null, a global
                                         // registry will be used.
     UsdStageRefPtr _stage;              // USD stage where the primitives are added
     bool _writeBuiltin;                 // do we want to create usd-builtin primitives, or arnold schemas
+    std::unordered_set<std::string> _exportedNodes; // list of arnold attributes that were exported
```
sirpalee: Instead of using `std::string` in the `unordered_set` we could use AtString as it comes with a cheap hash function. And the logic in `UsdArnoldWriter::writePrimitive` doesn't explicitly require the conversion to `std::string`, so we can keep using `AtString` there too.

translator/writer/writer.h
```diff
@@ -49,11 +49,14 @@ class UsdArnoldWriter {
     void setWriteBuiltin(bool b) { _writeBuiltin = b;}
     bool getWriteBuiltin() const { return _writeBuiltin;}
 
+    bool isNodeExported(const std::string &name) { return _exportedNodes.find(name) != _exportedNodes.end();}
```
sirpalee: We could use `std::unordered_map::count` here: `_exportedNodes.count(name) == 1`

[See reference](https://en.cppreference.com/w/cpp/container/unordered_map/count)

> Returns the number of elements with key that compares equal to the specified argument key, which is either 1 or 0 since this container does not allow duplicates.

translator/writer/prim_writer.cpp
```diff
@@ -28,6 +28,8 @@
 #include <string>
 #include <vector>
 
+#include <pxr/usd/usdShade/materialBindingAPI.h>
```
sirpalee: Please group this `#include` with the other pxr includes.

sebastienblor:
Ok, I don't have the error message, but I'm building against USD 19.05. I think the result is the same anyway, I get this in the final usd file :

```
def Mesh "sphere"
{
   .....
   string[] primvars:arnold:shader = ["ai_default_reflection_shader"]
}
def Shader "grey_shader"
{
    uniform token info:id = "arnold:standard_surface"
    float inputs:base = 0.7
    float inputs:specular = 0
}
def Shader "disp_shader"
{
    uniform token info:id = "arnold:checkerboard"
}
```

So there is no Material created, and it's interesting to see that when there is no shader assigned to a shape, calling AiNodeGetPtr(shape, "shader")` actually returns the default shader `ai_default_reflection_shader`. I would have expected that it returns `nullptr` and that the default shader is just handled internally by Arnold. We'd need to treat this use case, and also probably add some tests about the primitives being valid before trying to connect them

translator/writer/prim_writer.cpp
```diff
@@ -359,12 +361,16 @@ const UsdArnoldPrimWriter::ParamConversion* UsdArnoldPrimWriter::getParamConvers
...
+    // we're exporting a new node, so we store the previous list of exported 
+    // attributes and we clear it for the new node being written
+    std::unordered_set<std::string> prevExportedAttrs = _exportedAttrs;
     _exportedAttrs.clear();
 
     // Now call the virtual function write() defined for each primitive type
     write(node, writer); 
+
+    // restore the previous list (likely empty, unless there have been recursive creation of nodes)
+    _exportedAttrs = prevExportedAttrs;
```
sirpalee: The assignment operator and the constructor of the `std::unordered_map` copies elements, so we would end up duplicating elements without actually needing them. A more optimal solution is to use swap, which only swaps internal pointers between two `std::unordered_map`, instead of moving elements.

[Reference](https://en.cppreference.com/w/cpp/container/unordered_map/swap)

> Exchanges the contents of the container with those of other. Does not invoke any move, copy, or swap operations on individual elements.

So something like this:

```cpp
decltype(_exportedAttrs) prevExportedAttrs;
prevExportedAttrs.swap(_exportedAttrs);
write(node, writer);

_exportedAttrs.swap(prevExportedAttrs);
```

Bonus, using `decltype` lets you to change the stored type in the `std::unordered_set`, without breaking the code. :)

sirpalee:
@sebastienblor Yep, I'm using 0.20.2 not 0.19.5.

PR #200, opened 2020-02-06
Applying the OSX fix of delaying the USD library load to Linux

### LINKED ISSUE #199: The procedural causing a crash in the core during shutdown on Linux
**Describe the bug**
When rendering a USD procedural, the core crashes while shutting down and clearing its resources. There is no usable traceback and it seems to happen only on select Linux machines.

**To Reproduce**
1. Render a usd procedural via kick
2. Observe crash at the end of the render

**Expected behavior**
Arnold shuts down without a crash.

**Used Software Versions**
 - Arnold: 6.0.1.0
 - USD: 0.19.5 (static)
 - OS: Centos 6

sirpalee:
- Applying the OSX fix of delaying the USD library load to Linux

Fixes #199

### DIFF
--- a/procedural/SConscript
+++ b/procedural/SConscript
@@ -46,6 +46,9 @@ myenv.Append(LIBS = ['ai'])
 
 usd_deps = []
 
+if system.IS_LINUX:
+    myenv.Append(LIBS = ['dl'])
+
 if myenv['USD_BUILD_MODE'] == 'monolithic':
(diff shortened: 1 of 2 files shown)


sebastienblor:
Is this related to the linux crashes we face with usdview ?
In any case, we can do this trick for now in order to avoid the crashes, and hopefully in the future we can find a better solution

PR #213, opened 2020-02-13
Support leftHanded topologies for face varying primvars

### LINKED ISSUE #207: UV Index order reversed in render delegate
After 776ea72323fd674c79863918708d9d5bfdf31b80 the Arnold render delegate in Solaris is showing the wrong uvs.  

We need to apply isLeftHanded logic to the uvidxs as well, reversing the ordering, and potentially other face varying properties too.

sirpalee:
- Support leftHanded topologies for faceVarying primvars, including specialized path for uvs and normals.

Fixes #207

### DIFF
--- a/render_delegate/mesh.cpp
+++ b/render_delegate/mesh.cpp
@@ -65,7 +65,8 @@ void _ConvertVertexPrimvarToBuiltin(
 
 template <typename UsdType, unsigned ArnoldType>
 void _ConvertFaceVaryingPrimvarToBuiltin(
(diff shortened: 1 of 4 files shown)


sebastienblor:
I wondered if this was handled in the procedural, but it looks like it is. 
Not a big deal, but my understanding was that it's more efficient to fill a `std::vector` and make it an `AtArray`  with `AiArrayConvert` rather than allocating the array and then calling `AiArraySet*` for every index.

Issue #210, opened 2020-02-11
arnold-usd does not respect all types of visibility


kreppene:
**Describe the bug**
it looks like not all kinds of visibility is supported in arnold-usd
i run this check before create any arnold geo in my dev branch, and that seams to do the what we want (stolen from the Rodeofx project)
```
bool getPrimVisibility(const UsdPrim& prim, float frame)
{
	// We should also consider standard USD visibility flag.
    UsdGeomImageable imageable = UsdGeomImageable(prim);
    bool invisibleFromUSD = imageable &&
        imageable.ComputeVisibility(frame) == UsdGeomTokens->invisible;

    return invisibleFromUSD;
}
```

**To Reproduce**
render the .ass file provided

**Expected behavior**
We expect only 1 sphere to be visible, but 2 is visible when rendering,
theirs a total of 3 spheres inn the scene, 
so it looks like it respect some kind visibility but not all

**Screenshots**
![visibility](https://user-images.githubusercontent.com/7374105/74256327-79a76980-4cf3-11ea-8716-26d7d07dc618.png)

**Used Software Versions**
 - Arnold 6.0.1.0
 - USD  0.19.11
 - visual studio 2017
 - Windows 10

[visibility_test.zip](https://github.com/Autodesk/arnold-usd/files/4186868/visibility_test.zip)

PR #218, opened 2020-02-16
Prim visibility

kreppene:
changes inn this PR, checks if the usdPrim is visible before we build the arnold geo

this PR fixses this issue, https://github.com/Autodesk/arnold-usd/issues/210


sebastienblor:
Hi Raymond, I hope you don't mind but I've pushed a commit as there were some tabs causing indentation issues. I also did a cosmetic change on `getPrimVisibility`  to make it a bit simpler to read.

This seems to do the right thing, I'm just a bit worried in terms of performance about the comment in the USD repo for this function `UsdGeomImageable::ComputeVisibility` :

    ```
/// If called on each prim in the context of a traversal
    /// we will perform massive overcomputation, because sibling prims share
    /// sub-problems in the query that can be efficiently cached, but are not
    /// (cannot be) by this simple implementation.</b> If you have control of
    /// your traversal, it will be far more efficient to manage visibility
    /// on a stack as you traverse.
```

Could we be introducing a major slowdown here ? and if so, what other solutions would we have ? Computing the visibility along the hierarchy traversal would require a deeper re-writing of the whole traversal.

**Comment by sirpalee** [comment]
If all the logic we are going to implement is to skip translation/node creation when something is invisible, should we add it before we run any of the translators?

How is this going to affect the point instancer? In hydra, if you point instance a hidden mesh, the instances will be visible.

What happens with something like a mesh light?

**Comment by sebastienblor** [comment]
Indeed, there can be primitives that are marked as "invisible" but that must still be exported to arnold because they're needed by other nodes  (point instancer, mesh lights, etc...).
So skipping the export of these primitives won't always work.

In that case we must export the primitives but make it so that they don't appear in the render.
For shapes, we can set `visibility` to 0. For lights the only solution is to set `intensity` to 0.
This is what I've done in latest commit, along with `test_0028`.

Ideally we'd identify when such situations happen and skip the export of the nodes that don't need it. The commited code would remain, but in addition to that, during the stage traversal, we would skip the export of nodes (along the hierarchy) that are hidden. Then, in the last part of the reader that sets connections between nodes, if we need to do a link on a node that doesn't exist in the arnold scene, we would force an export of this node. This way we'd have the best of both worlds.
But that should probably be in another ticket.

@kreppene  could you confirm that the latest code in your branch does what it needs to do in your test scenes ?

**Comment by sebastienblor** [comment]
Now that #227 is addressed I've merged master to this branch, and added support for visibility in volumes. This required to update `test_0030` because we're supposed to see only 3 volumes for the instances (the proto one is hidden). Also, in the point instancer we were missing the fact that instances should have visibility set to `AI_RAY_ALL` otherwise they inherit from the proto visibility (and it might be hidden)

**Comment by kreppene** [comment]
sorry for taking so long, yes this works inn all my tests, perfect.
As you noted above, i'm a bit worried we are introducing a slowdown,  did you do any investigation on alternative ways of doing this @sebastienblor ?

**Comment by sebastienblor** [comment]
Another solution would be that while we're traversing the hierarchy, we check for node's token `visibility`, and skip everything below and invisible node. 
In the last step, when the connections between nodes are made, this might force the export of an invisible node, just like what we're doing in this branch.

But maybe we should first merge this PR and create another ticket to optimize the process

**Comment by kreppene** [comment]
Yes that makes sense.  Well for me this is good to go. I lett you Handel  the rest.

**Comment by sebastienblor** [comment]
So, in the last commit I rewrote the whole thing to avoid the possible slowdowns.
Instead of asking each primitive to compute the visibility (which looks for the `visibility`  token all along the hierarchy over and over), we now handle it during the stage traversal by just getting the value of this attribute. If the primitive is invisible we prune it, which won't even export any of its "children" primitives.

However, now in the second step while processing the connections between nodes, that were previously stacked, we check if some connections point to nodes that don't exist in the arnold world. These nodes are possibly invisible primitives that were skipped from export, but that actually need to be present because they're referenced by another node. 
Only in this not-so-common case, we go through a third step where we force the export of the missing nodes (setting their visibility or intensity to 0).

We should get now the best of both worlds : an optimized export in the most common cases, where invisible primitives should just be skipped, and in the rare cases where the invisible nodes must exist we force their creation and the slowdown would only happen for these specific nodes

**Comment by sirpalee** [comment]
The specification of the visibility parameter on UsdGeomImageable:

```
    token visibility = "inherited" (
        allowedTokens = ["inherited", "invisible"]
        doc = """Visibility is meant to be the simplest form of "pruning" 
        visibility that is supported by most DCC apps.  Visibility is 
        animatable, allowing a sub-tree of geometry to be present for some 
        segment of a shot, and absent from others; unlike the action of 
        deactivating geometry prims, invisible geometry is still 
        available for inspection, for positioning, for defining volumes, etc."""
    )
```

"visible" is not an acceptable value for visibility, so opening test 28 in usdview only shows the cone, because `/World/cube` is hidden due to the invalid visibility token.

sebastienblor:
Fixed `test_0028` in latest commit

PR #223, opened 2020-02-17
Support per-vertex uvs or normals list

### LINKED ISSUE #222: Per-vertex UV coordinates causing errors
**Describe the bug**
Since the change done in #216 , the kitchen scene no longer renders because there are some nodes with per-vertex UVs. In that case there is no `uvidxs` to give to arnold polymeshes, and this causes an error

When that happens, we should do the same as the render delegate does, and manually create a `uvidxs` array (based on `vidxs`)

sebastienblor:
When mesh primvars for uvs or normals are per-vertex, we manually build a `uvidxs` / `nidxs` array by copying `vidxs` .  This is similar to what is done in the render delegate.

While doing this change, I realized we were converting uv primvars to `uvlist` even for non-mesh nodes, so I've fixed this.
I also noticed that a deprecated function `exportPrimvars` had been left around, even though it's not used anymore. It got me confused when I was debugging this issue, so I've removed it.

Fixes #222


sebastienblor:
Fixed the regression that made this test fail, and added a plane geometry with per-vertex UVs and normals.
I realized that the normals primvars code wasn't in the correct place and would have been skipped, I also fixed some  camel-case functions

PR #287, opened 2020-03-19
Author UsdGeomSubset in the usd writer

### LINKED ISSUE #286: Author UsdGeomSubset in the writer
The usd reader now supports UsdGeomSubset primitives, and per-face shading assignments.
We should implement this in the writer, so that arnold polymeshes & curves having per-face / per-strand shading assignments author UsdGeomSubset prims.

sebastienblor:
When writing a usd file, if an arnold polymesh / curves shape has multiple shaders assigned, we now export this as a `UsdGeomSubset`.

This PR comes with a test `test_0036` containing per-face surface and displacement shaders, where the usd scene was manually generated using the changes from this PR. It is not really testing this ticket, since we don't have writer tests yet, but it's actually testing #283 with displacement + surface shaders.

Fixes #286


sirpalee:
Note: handled a merge conflict after merging in #289 .

PR #297, opened 2020-03-24
Prevent crashes when writing USD files with upper-case extensions

### LINKED ISSUE #288: Saving to a .USD file with capitals extension crashes
**Describe the bug**
When using the scene writer with an output filename like ".USD" with extensions in capitals, usd crashes.
Since Arnold scene format plugin is case-insensitive, we should handle this in `scene_write`, either dumping an error or automatically turning it into lower case.

sebastienblor:
When trying to write to a USD file with upper-case extensions, we convert it to lower case to prevent the crashes
 
Fixes #288

### DIFF
--- a/procedural/main.cpp
+++ b/procedural/main.cpp
@@ -21,6 +21,8 @@
 #include "registry.h"
 #include "../utils/utils.h"
 
+#include <pxr/base/tf/pathUtils.h>
+
 #if defined(_DARWIN) || defined(_LINUX)
 #include <dlfcn.h>
 #endif
@@ -156,7 +158,7 @@ procedural_viewport
     applyProceduralSearchPath(filename, universe);
 
     if (!UsdStage::IsSupportedFile(filename)) {
-        AiMsgError("[usd] Filename not supported : %s", filename.c_str());
+        AiMsgError("[usd] File not supported : %s", filename.c_str());
         return false;
     }
 
@@ -246,7 +248,7 @@ AI_SCENE_FORMAT_EXPORT_METHODS(UsdSceneFormatMtd);
 scene_load
 {
     if (!UsdStage::IsSupportedFile(filename)) {
-        AiMsgError("[usd] Filename not supported : %s", filename);
+        AiMsgError("[usd] File not supported : %s", filename);
         return false;
     }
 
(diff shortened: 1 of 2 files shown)


sebastienblor:
I realized the first commit wasn't really addressing the problem, which is that we should ensure the Usd Stage is valid before we read / write on it.

This can be done after the stage is created, by verifying that it's non-null. 
We can also check beforehand if the filename is supported, by calling `UsdStage::IsSupportedFile`.
The second commit is now doing all this, and it errors out if the filename isn't correct, which also prevents the crashes we were facing. I also ported the same tests on the other functions, in the reader and the writer.

I'm wondering what we want to do with the first commit, which forces the extension to be lower case in `scene_write`.... 
So when the users asks to write  `filename.USD` , would we want a warning saying that it's turned into lower case and saving `filename.usd` , or do we just want an error saying that the filename is invalid ?

procedural/main.cpp
```diff
@@ -264,8 +274,28 @@ scene_load
 //                 const AtParamValueMap* params, const AtMetadataStore* mds)
 scene_write
 {
+    std::string filenameStr(filename);
+
+    std::string extension = filenameStr.substr(filenameStr.find_last_of('.'));
+    if (extension == ".USD" || extension == ".USDA" || extension == ".USDC") {
```
sirpalee: Let's check for USDZ too.

translator/reader/reader.cpp
```diff
@@ -165,6 +165,10 @@ void UsdArnoldReader::readStage(UsdStageRefPtr stage, const std::string &path)
 {
     // set the stage while we're reading
     _stage = stage;
+    if (stage == nullptr) {
+        AiMsgError("[usd] Impossible to read USD Stage %s", _filename.c_str());
```
sirpalee: I don't think `Impossible to read USD Stage` is the best phrasing here. We are not reading an USD stage, but we are opening a stage from from a file that USD happens to support. Maybe, `Unable to open %s` or `Unable to create USD stage from %s`

procedural/main.cpp
```diff
@@ -155,6 +155,11 @@ procedural_viewport
 
     applyProceduralSearchPath(filename, universe);
 
+    if (!UsdStage::IsSupportedFile(filename)) {
+        AiMsgError("[usd] Filename not supported : %s", filename.c_str());
```
sirpalee: I think `File not supported` sounds better, that also matches the USD function `UsdStage::IsSupportedFile`.

procedural/main.cpp
```diff
@@ -240,6 +245,11 @@ AI_SCENE_FORMAT_EXPORT_METHODS(UsdSceneFormatMtd);
 // SceneLoad(AtUniverse* universe, const char* filename, const AtParamValueMap* params)
 scene_load
 {
+    if (!UsdStage::IsSupportedFile(filename)) {
+        AiMsgError("[usd] Filename not supported : %s", filename);
```
sirpalee: Same here as above.

procedural/main.cpp
```diff
@@ -264,8 +274,28 @@ scene_load
...
+    if (extension == ".USD" || extension == ".USDA" || extension == ".USDC") {
+        std::transform(extension.begin(), extension.end(), extension.begin(),
+            [](unsigned char c){ return std::tolower(c, std::locale()); });
+        filenameStr = filenameStr.substr(0, filenameStr.find_last_of('.'));
+        filenameStr += extension;
+        AiMsgWarning("[usd] Filename extensions must be lower-case. Saving to %s", filenameStr.c_str());
+    }
+
+    if (!UsdStage::IsSupportedFile(filenameStr)) {
+        AiMsgError("[usd] Filename not supported : %s", filenameStr.c_str());
```
sirpalee: Same here as above.

procedural/main.cpp
```diff
@@ -264,8 +274,28 @@ scene_load
 //                 const AtParamValueMap* params, const AtMetadataStore* mds)
 scene_write
 {
+    std::string filenameStr(filename);
+
+    std::string extension = filenameStr.substr(filenameStr.find_last_of('.'));
```
sirpalee: We can use `TfGetExtension` here from `pxr/base/tf/pathUtils.h`.

Documentation:
```
/// Returns the extension for a file path
///
/// If \p path is a directory path, an empty path, or a dotfile path, return
/// the empty string. Otherwise return \p path 's dot-separated extension as
/// a string(dot not included).
///
/// Examples:
/// 
/// TfGetExtension('/foo/bar')              -> ''
/// TfGetExtension('/foo/bar/foo.baz')      -> 'baz'
/// TfGetExtension('/foo.bar/foo.baz')      -> 'baz'
/// TfGetExtension('/foo/bar/foo.101.baz')  -> 'baz'
/// TfGetExtension('/foo/bar/.foo.baz')     -> 'baz'
/// TfGetExtension('/foo/bar/.foo')         -> ''
TF_API
std::string TfGetExtension(std::string const& path);
```

For example, just by splitting using the '.' character won't handle a bunch of corner cases, like /foor/bar/.USDA .

sirpalee:
One other improvement is to query all the supported file formats, by using `SdfFileFormat::FindAllFileFormatExtensions` and use that to fix the uppercase versions.

Python code to do this:
```
>>> from pxr import Sdf
>>> Sdf.FileFormat.FindAllFileFormatExtensions()
['sdf', 'usd', 'usda', 'usdc', 'usdz']
```

One complication, if there are other registered file formats, like abc, obj, we would be fixing the extension for those as well.

sebastienblor:
In latest commit I'm no longer looping over hardcoded upper case extensions. Instead, if `IsSupportedFile` returns false, I check if it works better with a lower case extension, and change the filename in that case ( along with a dedicated warning). I ported the other suggestions too

sirpalee:
That sounds good to me, way simpler and we don't have to query extensions etc.

Issue #157, opened 2020-01-21
Add tests for the scene format plugin


sirpalee:
**Describe the feature you'd like**
Add tests for the new feature in #151 

**Additional context**
We should do this after #124 , since adding new tests (a new type of tests) will be easier after making more progress with the cmake conversion and the best ways to test this feature is compare generated output from custom commands, which is not trivial with the current test suite.

Tests to add:
- [ ] Option node translation #321 
- [ ] Motion key support #348
- [ ] Per channel connections #352
- [ ] Writing polymesh.subdiv_iterations #353
- [ ] Writing ginstance attributes #363
- [ ] Textured mesh light support #366
- [ ] Writing motion range #369
- [ ] Writing toon shader #377
- [ ] Linking to array attribute elements #381
- [ ] UsdGeomCamera support #388
- [ ] Load namespaced primvars #383
- [ ] Write connections for user-data of type NODE #371 
- [ ] Support motion blur in the scene format loader #347

PR #321, opened 2020-04-01
Translate the options node properly

### LINKED ISSUE #320: Options outputs not translated properly
**Describe the bug**
When writing and reading an arnold scene, we can see that the options attribute `outputs` that lists the output AOVs, is incorrect.
Because of that, when we export an arnold scene to USD and then kick it, there won't be any outputs and the output image(s) won't be saved

sebastienblor:
The problem is twofold :
 1 - in the USD writer, we need a special treatment to translate `options.outputs` to USD. The problem is that the usd primitive names are different from the arnold node names, and `options.outputs` is a string array that contains the name of driver and filter nodes. Therefore we need to modify this strings so that the node names correspond to the usd ones
 2 - In the USD reader, the string array attributes weren't translated properly. Given how the code was written, we were trying to read a string array attribute as a TfToken array, which fails. Plus, later on we were trying to do a memcopy of `TfToken` to `AtString` which obviously cannot work. So we need to do a special treatment for string array attributes (like we already do for string attributes) and do the conversion from string to AtString

Once we can do tests for the scene format that kick usd files (#157 ), they will automatically test this use case

Fixes #320

### DIFF
--- a/translator/reader/prim_reader.cpp
+++ b/translator/reader/prim_reader.cpp
@@ -139,7 +139,7 @@ void UsdArnoldPrimReader::readArnoldParameters(
                             // FIXME: ensure enum is working here
                         case AI_TYPE_ENUM:
                         case AI_TYPE_STRING:
-                            exportArray<std::string, std::string>(attr, node, arnoldAttr.c_str(), time);
+                            exportStringArray(attr, node, arnoldAttr.c_str(), time);
                             break;
                     }
                     {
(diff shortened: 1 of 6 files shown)


sebastienblor:
I could reproduce the same issue on Mac.  The macro cannot expand __AI_FILE__ nor __AI_LINE__ , maybe because this is called from an inlined template function in a .h file. 
My latest commit fixes it on Mac, let me know if it works on linux too.

I made a dedicated function for string array conversions `exportStringArray` , that at least now also handles arrays of `TfToken` and `SdfAssetPath` . The call to `AiArraySetStr`  is now in the .cpp and it builds properly on Mac. Testsuite is still fine.

PR #327, opened 2020-04-01
Read RGB arrays

### LINKED ISSUE #325: RGB Array attributes not converted properly
**Describe the bug**
When reading a USD file where arnold shaders have rgb array attributes (for example in `ramp_rgb`), the attributes aren't translated properly because we're trying to set them as vector arrays instead of rgb

sebastienblor:
I'm adding an optional `type` attribute to `exportArray` so that we can choose a specific arnold type for the reader. We can then read RGB array attributes as expected.

`test_0041` verifies this with a scene containing a `ramp_rgb` shader

Fixes #325


sebastienblor:
Thanks ! I had left VECTOR and RGB attributes in the same switch-case, so we were now trying to convert vector-array attributes as rgb-array. I fixed this in the last commit, and also fixed the indentation of the whole switch block. There were several useless curly brackets around some cases.

PR #341, opened 2020-04-03
Support shader connections to skydome and quad light's color

### LINKED ISSUE #340: Support reading any shading node connected to a Dome light
**Describe the bug**
Arnold supports connecting any shading node to a skydome light color, wheras native usd dome lights only support a determined texture file path.

Currently the usd writer translates this use case with a shading connection, but the reader doesn't read it back to arnold

sebastienblor:
We now look for connections on the light `color` attribute, for both quad and skydome lights.
Note that we were previously setting skydome's "camera" attribute to 0, which is different from arnold default and might not make much sense.  I removed this, so now scenes with a USD Dome light will show up in the background for camera rays.
 
Some tests had to be updated, and I'm adding `test_0042` for this specific use case

Fixes #340


translator/reader/read_light.cpp
```diff
@@ -127,7 +127,16 @@ void UsdArnoldReadDomeLight::read(const UsdPrim &prim, UsdArnoldReaderContext &c
             AiNodeSetStr(node, "format", AtString("angular"));
         }
     }
-    AiNodeSetFlt(node, "camera", 0.f);
+
+    // Special case, the attribute "color" can be linked to some shader
```
sirpalee, line 131: Could you please move this code to a utility function?

sebastienblor: done in latest commit

sirpalee:
Added the missing node parameter to exportLightColorLinks function and used an anonymous namespace instead of the static decorator to specify internal linkage.

PR #347, opened 2020-04-06
Support motion blur in the scene format loader

### LINKED ISSUE #346: Support motion blur in the scene format
**Describe the bug**
Currently, when loading a usd file in a procedural, we tell the reader beforehand what's the motion range (and if motion blur is enabled) based on the main arnold scene settings.

But in the scene format, when kicking a usd file, we need to find the render camera and check its shutter before having this information. So we need an extra step before traversing the stage for that.

sebastienblor:
When the reader is called without any parent procedural (thus from the scene format loader), we look for the render camera to get the shutter range. This will tell us how to load the USD data.

While doing the change, I noticed that before we were adding an additional useless time sample when the arnold data had only 2 motion keys (which is the default use case), so I've also fixed this 

Fixes #346

### DIFF
--- a/translator/reader/reader.cpp
+++ b/translator/reader/reader.cpp
@@ -216,6 +216,40 @@ void UsdArnoldReader::readStage(UsdStageRefPtr stage, const std::string &path)
         rootPrimPtr = &rootPrim;
     }
 
+    // If there is not parent procedural, and we need to lookup the options, then we first need to find the 
+    // render camera and check its shutter, in order to know if we need to read motion data or not (#346)
+    if (_procParent == nullptr) {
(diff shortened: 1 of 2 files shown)


sebastienblor:
> Could you please add a test for this? I would just use a C++ test that reads in a USD file using a procedural and checks the key values and numbers. Rendering motion blurred scene is always messy, because sampler changes could cause false positives.

This code wouldn't be exercised from withing a procedural, this needs to be done through the scene format reader, which doesn't have any tests yet.

sirpalee:
Gotcha! We can add tests later then.

PR #348, opened 2020-04-07
Support motion keys in the USD writer

### LINKED ISSUE #334: Support motion keys in the usd writer
Currently the USD writer doesn't support motion keys in arrays. 
This is required to support motion blur when exporting arnold scenes to usd

sebastienblor:
This PR ensures that we're writing each of the motion keys from arnold array attributes, whether they're saved as USD builtin attributes (mesh matrix, vertices, etc...) or whether they're saved as arnold-specific attributes.

I also fixed another issue related to the amount of keys saved like in #347 , in another part of the reader code.

Fixes #334


sirpalee:
Let's not forget to add tests for this in #157 .

PR #352, opened 2020-04-07
Write per-channel connections by inserting adapter shaders

### LINKED ISSUE #351: Write per-channel connections by inserting adapter shaders
Arnold shaders can have connections on specific attribute channels (e.g. color.r, vector.x, etc...), which isn't currently supported in USD.
#165 will deal with supporting these connections when they're (somehow) present in the USD file.

Meanwhile, we could make it so that the USD writer inserts "adapter" shaders in the shading tree for these use cases. The output USD file could render exactly as the original .ass would, but without any of these connections.

sebastienblor:
When a per-channel connection is found in an arnold scene, we insert a `float_to_rgba` shader in the shading tree. It gets connected to the input attribute, and we can set each of its channels independantly. Note that arnold handles the eventual different types in the connection so always using this RGBA shader works, even for RGB, Vector, Vector2.

Fixes #351

### DIFF
--- a/translator/writer/prim_writer.cpp
+++ b/translator/writer/prim_writer.cpp
@@ -447,6 +447,62 @@ std::string UsdArnoldPrimWriter::getArnoldNodeName(const AtNode* node)
     return name;
 }
 
+// Ensure a connected node is properly translated, handle the output attributes, 
+// and return its name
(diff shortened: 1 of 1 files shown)


sebastienblor:
we'll also need tests for this in #157

PR #363, opened 2020-04-09
Improve how ginstance attributes are written

### LINKED ISSUE #362: Improve how ginstance attributes are written
**Describe the bug**
Usually, we author attribute values when they're different from default. This cannot work for `ginstance` nodes, because of the way they inherit attributes from their instanced node.

For example, if the original geometry has `visibility=0`  and the ginstances has `visibility=255` (which is the most frequent use case), the visibility value won't be written to USD because 255 is the default. Because of that, no instance will be visible

sebastienblor:
We register a new primitive writer for `ginstance` nodes.
It will do a special treatment for attributes `visiblity`, `sidedness`, `matte`, `receive_shadows`, `self_shadows`, `invert_normals`. Note that we don't consider `opaque` which is not used anymore.

For these attributes, we compare the value with the value from the instanced node. If the value is different, we author them. 

Fixes #362

### DIFF
--- a/translator/writer/prim_writer.h
+++ b/translator/writer/prim_writer.h
@@ -35,7 +35,7 @@ PXR_NAMESPACE_USING_DIRECTIVE
  **/
 class UsdArnoldPrimWriter {
 public:
-    UsdArnoldPrimWriter() {}
+    UsdArnoldPrimWriter() : _motionStart(0.f), _motionEnd(0.f) {}
     virtual ~UsdArnoldPrimWriter() {}
     
     void writeNode(const AtNode *node, UsdArnoldWriter &writer);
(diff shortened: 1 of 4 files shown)


sebastienblor:
I had accidentally left a change meant for another ticket in file `prim_writer.h`. I rolled it back in the last commit

sebastienblor:
Note that this behaviour isn't fully exact, but it's the best we can do with the current arnold API.
Example : 
mesh A, and ginstance B pointing at A. 
In arnold, the attribute `receive_shadows` is on by default. But mesh A has it set to off.
If instance B explicitely set `receive_shadows` to ON, then it's supposed to have it enabled. This is what the current patch does.
But if it leaves the attribute untouched, then it will still be ON when we ask arnold for the attribute value. But in arnold, this is supposed to use the value of A, because it's not explicitely set in the instance. So in this case the instance is supposed to have receive_shadows = off, which is not what will happen with the above patch. The problem is that there is no way, with the arnold API, to distinguish between "the attribute is ON because it's the default" and "the attribute is ON because it was explicitely set". So there is no way to find a solution that would work all the time. Note that this only applies to attributes `visibility, sidedness, receive_shadows, self_shadows, matte, opaque`.

The above patch was done to solve the most frequent use case with instances, which is that the original geometry has visibility OFF (0), but the instance has visibility ON (255)

sirpalee:
Let's not forget to add tests for this in #157 .

PR #366, opened 2020-04-10
Support textured mesh lights

### LINKED ISSUE #365: Support textured mesh lights in the procedural
**Describe the bug**
In #340 we added support for textured dome lights and quad lights. But Arnold also supports textured mesh lights. We need to port the same code to mesh lights as well.

sebastienblor:
We just need mesh lights reader to call the function that was introduced in #340 for skydome and quad lights.

I have a test scene ready, but it requires the scene format testsuite first. Loading mesh lights through a procedural doesn't seem to work in arnold, so I cannot test it with the procedural

Fixes #365


sirpalee:
Let's not forget to add a test when doing #157 .

PR #392, opened 2020-04-21
Support tests for the scene format plugin

### LINKED ISSUE #157: Add tests for the scene format plugin

sebastienblor:
Arnold 6.0.3.0 now officially supports scene format plugins., which allow to kick usd files, or to resave ass files to usd and vice-versa.

In this PR I'm adding support for 2 new sorts of tests :
1) if there is no `test.ass` in the test folder, but there instead `test.usd`, `test.usda` or `test.usdc` , we kick this file directly. Adding `test_0050` as a proof of concept. This will exercise the usd reader translator.
2) We had a test param `resaved` , that when set in the README was resaving the file to `test_resaved.ass` and then kick this resaved file. We want to extend this concept to resave to other file formats. So I replaced this boolean `resaved` param by a string `resave` param that provides the file format we want to save to. I fixed `test_0034` that was relying on the previous parameter and I added `test_0051` as a proof of concept. It has a `test.ass` file that is resaved to `test_resaved.usda` and then kicked. This sort of tests is very interesting, because it exercises both the writer (when resaving to usd) and the reader (when kicking the file), so a regression in one of the modules could be revealed by this test.

This implements what is needed for #157
Other PRs will follow with more tests


sebastienblor:
After discussing it, we realized that it's not possible to maintain a testsuite that will work for older versions of Arnold, as it can happen frequently that changes in Arnold require to update the test results.

So for now, we'll just mention in the documentation that the testsuite requires the latest version of Arnold that was released. I mentioned this in the README in the last commits

PR #397, opened 2020-04-24
Adding more scene format tests

### LINKED ISSUE #157: Add tests for the scene format plugin

sebastienblor:
Still related to #157 , this PR adds more tests exercising the scene format

Fixes #157


sebastienblor:
Pushed more tests in latest commit, we now have tests for all the features we've been working on regarding the scene format

PR #401, opened 2020-04-29
Fixing hangs with several usd procedurals

### LINKED ISSUE #400: Multiple usd procedurals causing hangs
**Describe the bug**
When a scene has multiple usd procedurals, it frequently hangs or crashes, because of the procedural node init enabled by default.

This seems to be happening  because of the global registry, that clearly isn't thread-safe for multiple readers in parallel

sebastienblor:
The USD reader wasn't thread-safe in case of multiple readers running in parallel, which happens by default with multiple usd procedurals in the scene.

The problem was related to the global registry that is supposed to be processed once. This PR adds some mutex locks when this is needed (as well as in another static variable `counter` in the code that wasn't protected). Also it ensures that we don't call `RegisterPrimitiveReaders` outside of the lock, when it's the global registry.

To do this properly, I have removed the `mask` field from the registry and I moved it to the prim reader. The registry now knows about all the possible node types, it's just when iterating over the scene that it can decide to skip some nodes or not depending on their type and the existing mask. 

Fixes #400


sebastienblor:
Added a test for the use case of multiple procedurals

PR #436, opened 2020-05-18
Extend viewport representation of nested procedurals

### LINKED ISSUE #435: Improve nested usd procedurals viewport representation
**Describe the bug**

In #424  we added support for nested procedurals in the viewport API. However, this only worked with a single level of nested procedurals, and wouldn't work with multiple levels of nested procedurals.

This ticket is about extending the current code to support for any level of recursion.

sebastienblor:
This ticket improves the work done in #424 and now allows for recursive nested procedurals.
Now we create a procedural in a temporary universe, and we call `AiProceduralViewport` in this procedural, so that it populates the desired universe. If another usd procedural is found, the same will happen recursively, and so on....

Fixes #435


sirpalee:
Can we add tests for this, similar to test 38?

Issue #439, opened 2020-05-20
The Render Delegate crashes when active rendervars are deleted in Solaris


sirpalee:
**Describe the bug**
The Render Delegate crashes when active rendervars are deleted in Solaris.

**To Reproduce**
1. Create a Solaris scene and set up a few rendervars.
2. Delete one of the active rendevars.
3. The Render Delegate crashes.

**Expected behavior**
We shouldn't crash Houdini, but keep on rendering without the deleted rendervar.

PR #445, opened 2020-05-24
HdArnoldRenderBuffer : Fixed uninitialized _mutex variable.

Tuftux:
- Constructor initalize _mutex variable.

When quickly switching between AOVs using UsdImagingGLEngine SetRendererAOV(), I was tripping the debugger on a lock_guard using a "_mutex" member variable that seemingly was unitialized. The following attempt at a fix seems to resolve the issue.

### DIFF
--- a/render_delegate/render_buffer.cpp
+++ b/render_delegate/render_buffer.cpp
@@ -193,7 +193,7 @@ WriteBucketFunctionMap writeBucketFunctions{
 
 } // namespace
 
-HdArnoldRenderBuffer::HdArnoldRenderBuffer(const SdfPath& id) : HdRenderBuffer(id)
+HdArnoldRenderBuffer::HdArnoldRenderBuffer(const SdfPath& id) : HdRenderBuffer(id), _mutex()
 {
     _hasUpdates.store(false, std::memory_order_release);
 }


sirpalee:
According to the c++11 standard 12.6.2/8

> In a non-delegating constructor, if a given non-static data member or base class is not designated by a mem-initializer-id (including the case where there is no mem-initializer-list because the constructor has no ctor-initializer) and the entity is not a virtual base class of an abstract class (10.4), then
> — if the entity is a non-static data member that has a brace-or-equal-initializer , the entity is initialized as specified in 8.5;
> — otherwise, if the entity is an anonymous union or a variant member (9.5), no initialization is performed;
> — otherwise, the entity is default-initialized (8.5)

_mutex is a non-static data member, that has a `brace-or-equal-initializer`, so it should be initialized with it's default constructor.

Example from the standard:

```
struct A {
    A();
};
struct B {
    B(int);
};
struct C {
    C() { } // initializes members as follows:
    A a; // OK: calls A::A()
    const B b; // error: B has no default constructor
    int i; // OK: i has indeterminate value
    int j = 5; // OK: j has the value 5
};
```

See : http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3690.pdf

If this change fixes your issue, it might be just a pure coincidence. I checked the constructor of the `HdRenderBuffer` (I think that's executed before the base members are initialized?), but that's not doing anything nasty either that could cause the issue. Recently we fixed some issues related to deleting/recreating render buffers on the fly (see #439), have you tried using this version? Taking a quick glance at `SetRendererAovs`, which ends up in `HdTaskController::SetRendererOutputs`, the code is going to destroy bprims (including render buffers), and #439 fixed something when destroying bprims.

Tuftux:
Hey Pal, I'm by no means an expert in C++, more of a lighter messing with code, so there's a good chance it could be coincidental !
I'll have a shot at running my app before/after #439 and see if the issue persists.

Thanks !

Issue #280, opened 2020-03-13
Create ArnoldRendererSettingsAPI USD Schema


jhodgson:
Karma defines a KarmaRendererSettingsAPI USD scheme for its render settings shipped in _.../houdini/dso/usd_plugins/usdKarma/resources/usdKarma/schema.usda_ which inherits from the _RendererSettingsAPI_ schema.  In this file the properties are prefixed with _karma:global:_

We should consider creating an ArnoldRendererSettingsAPI USD schema to package our Arnold specific render settings.

PR #454, opened 2020-06-02
Support UsdRender schemas in the reader

### LINKED ISSUE #453: Support UsdRender schemas in the reader
The USD reader should be able to read scenes with `UsdRenderSettings`  , `UsdRenderProduct`, `UsdRenderVar  primitives, as created by Solaris. 
These nodes can represent the arnold options node, as well as the drivers and filters used to perform an arnold rendering

sebastienblor:
This PR adds a prim reader for `UsdRenderSettings` nodes. This reader will be reponsible to set the arnold options node, but also to create all the necessary drivers, filter, and aov shaders.

This reader is more complicated than the other ones, as there is a lot of complicated conversions being done under the hood. We'll need to decide in another ticket if we also want to create a prim writer to author such prims in the USD writer.

Also, we should create an API schema as we've done for other node types., which is discussed in #280 

Fixes #453


sebastienblor:
In the last commit I'm adding a test where we kick a usd file with such a render settings prim.
I'm also supporting different namespaces for the arnold specific attributes in the render settings. I've seen some `arnold:global:` prefixes in some Solaris scenes, so I'm adding this to the list of supported namespaces too.

Issue #496, opened 2020-08-10
Add support for the instancer procedural in the writer


sebastienblor:
Arnold scenes with an `instancer` procedural are currently ignored by arnold. 
After #487 they will be considered as a custom procedural. We could translate it as a `PointInstancer` instead.

sebastienblor:
Note that `test_0137` from #497 would have to be updated when this issue is addressed

PR #497, opened 2020-08-10
Add support for custom procedurals

### LINKED ISSUE #487: Add a schema for custom procedurals
Today we have a schema `ArnoldProcedural` for the arnold node `procedural` (loading .ass files), as well as `ArnoldAlembic` and `ArnoldUsd`. 
But if a studio uses a custom procedural, they would need to define a new usd schema for their specific arnold node.

It could make more sense to define a schema `ArnoldCustomProcedural` with a token that determines what the arnold node type is, in a very similar way to what `UsdShade` does for shaders. So for a procedural called "myProc", instead of having a primitive `ArnoldMyProc`, we would have a primitive `ArnoldCustomProcedural` with an attribute  `nodeType = "myProc"`

sebastienblor:
The procedural now supports reading / writing custom procedurals, as a schema  `ArnoldProceduralCustom` with a string attribute `node_entry` returning the procedural type.
All the other attributes for the procedural are present without any namespace.

I'm adding `test_0137` for this use case, that will probably have to be changed in the future. Since I didn't figure out yet how to get a custom procedural for a test, I'm using the builtin procedural `instancer` that is currently not supported, therefore considered by the translator as a custom proc. This should change with #496 , so the test will have to be updated when it's done

Fixes #487


sebastienblor:
The schemas generation should be updated to support this. Since I saw they were already outdated I created #500 to do a full check of what's and what's not up-to-date

PR #503, opened 2020-08-10
Improve the schemas builds

### LINKED ISSUE #500: Update the schema builds
We should ensure that the schemas generation is still up-to-date with the latest improvements.
Some recent additions to the translator, like cameras, render settings, custom procedurals, etc... were not ported to the schemas build

sebastienblor:
The schemas build was outdated. This PR is about ensuring it's in-line with the behaviour in the translator and render_delegate.

Here's the list of changes :
- Base typed schemas for common types (`ArnoldShape`, `ArnoldOperator`, etc...) now inherit from the base schema typed instead of a corresponding usd builtin schema. They are now declared as being non-instantiable.
- API schemas for common types (`ArnoldShapeAPI`, `ArnoldLightAPI`, `ArnoldCameraAPI`) now define the correct list of attributes ((they ignore the ones already present in the USD builtin schema)
- API schemas for  "leaf" classes (`ArnoldPolymeshAPI`, `ArnoldSkydomeLightAPI`, etc...) inherit from the corresponding base API schema mentioned above. They ignore the attributes already present in the USD corresponding schema
- Set `apiName` in the `customData` section for all attributes, so that the C++ functions are meaningful
- Added support for node attributes
- Added a special case for attribute `operator` (in procedurals), which like `namespace` isn't accepted as a usd token
- Added camera and persp_camera as API schemas (ortho isn't needed as it doesn't define any specific attribute)
- Added typed schema `ArnoldProceduralCustom` for custom procs (see #497)
- Added schema `ArnoldOptions`

Fixes #500

### DIFF
--- a/schemas/createSchemaFile.py
+++ b/schemas/createSchemaFile.py
@@ -71,6 +71,10 @@ def getParameterStr(paramType, paramValue = None, paramEntry = None):
         if paramValue:
             strVal = str(paramValue.contents.STR)
             valueStr = '"{}"'.format(str(strVal))
+    elif paramType ==  ai.AI_TYPE_NODE:
+        typeStr = 'string'
+        strVal = ''
+        valueStr = '""'
 
     # FIXME: matrix parameter is an array
     elif paramType ==  ai.AI_TYPE_MATRIX:
@@ -125,6 +129,9 @@ def getParameterStr(paramType, paramValue = None, paramEntry = None):
 
     return typeStr, valueStr, optionsStr
 
+def makeCamelCase(name):
+    return ''.join(x.capitalize() or '_' for x in name.split('_'))
+
 ''' 
 Convert an Arnold Param Entry to a USD declaration
 and return it as a string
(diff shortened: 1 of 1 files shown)


sebastienblor:
To summarize, all arnold classes are defined as typed schemas and can be instantiated : `ArnoldSetParameter`, `ArnoldGinstance`, etc... in order to group them, we also add schemas for the entry types : `ArnoldShape`, `ArnoldLight` that contain the attributes that are common to all these nodes (e.g. shape defines `matrix`, `shader`, `visibility`, etc...), so the leaf nodes inherit from these classes.

In the translator and render_delegate, shapes, lights and cameras use by default the usd builtin schemas, and look for arnold-specific attributes in the scope `primvars:arnold`. For this we want API schemas that just provide a set of functions to extract these additional attributes. Same as for typed schemas, we first create API schemas for the base classes (`ArnoldShapeAPI`, `ArnoldLightAPI`, `ArnoldCamerAPI`) that inherit from `APISchemaBase`. They contain functions to access the arnold-specific attributes. Leaf nodes have an API schema inheriting from this base one. This way e.g. `ArnoldSkydomeLightAPI` has all the functions to get the arnold attributes, whether they're common to all lights or just for the skydome

PR #515, opened 2020-08-31
Add support for UsdPreviewSurface Usdtransform2d

### LINKED ISSUE #516: Support Transform2d for the Render Delegate
Add support for [Transform2d](https://graphics.pixar.com/usd/docs/UsdPreviewSurface-Proposal.html#UsdPreviewSurfaceProposal-Transform2d) .

arcilat-adsk:
Add support for UsdTransform2d in material network processing. See https://graphics.pixar.com/usd/docs/UsdPreviewSurface-Proposal.html#UsdPreviewSurfaceProposal-Transform2d.

Fixes #516


sirpalee (changes requested):
I tried with a simple file and it does not work for me (but it works as expected using HdStorm), could you please take a look? Just add any png next to the extracted usd file with the name "image.png". After exporting it to ass, it seems that the image shader is not connected to the matrix multiply vector node. I was using arnold 6.0.4.0 and USD 20.8 on CentOS 7.

[transform2d.usda.zip](https://github.com/Autodesk/arnold-usd/files/5154284/transform2d.usda.zip)

arcilat-adsk:
@sirpalee Thanks for your feedback. Two things after investigating a bit:

- I have added a commit to fix the rotation (the previous code was feeding the pxr::GfRotation with radians while it is expecting degrees)
- Changing the file so `string inputs:varname = "st1"` is `token inputs:varname = "st1"` fixes  the issue

For this case, hdStorm seems to be processing strings and tokens the same way in this case.

In hdArnold, the test `if (varnameValue.IsEmpty() || !varnameValue.IsHolding<TfToken>())` in `float2PrimvarRemap` (material.cpp:206) bypasses the conversion of the `UsdPrimvarReader_float2` node. This actually leads to a message (node "UsdPrimvarReader_float2" is not installed). This explains the unconnected graph you saw during your tests.
Note that, because of that, the default case (without any texture transform) is also broken when the primvar is not named `st` and `varname` is a string in the PrimvarReader.

The previewsurface specification defines the `varname` input to be a string, not a token (https://graphics.pixar.com/usd/docs/UsdPreviewSurface-Proposal.html#UsdPreviewSurfaceProposal-PrimvarReader).
If this is fine with you, let's consider this PR complete. I'll submit a new issue regarding the `string varname` vs `token varname` story.

arcilat-adsk:
token vs string issue for varname issue is reported here: https://github.com/Autodesk/arnold-usd/issues/518

sirpalee (approved):

sirpalee:
Thanks for the updates and spotting the issue with std::string and token! This is all good to go now.

PR #545, opened 2020-09-24
Support animated visibility

### LINKED ISSUE #543: Visibility is not checked on the current frame
**Describe the bug**
Visibility is not working correctly, if a prim has its visibility turned off at the start of the time-range of its usd file, it stays off even after the prim is keyed on again.

we fixed this problem on our branch by doing this 
https://github.com/kreppene/arnold-usd/commit/aea7df5a3ff4374c0f55c4d5657b2846658dd149

**To Reproduce**
create a sphere , on the first frame key its visbility of, and then a few frames later key the visibility on.
now export this sphere to usd. and render a frame where the sphere is suppose to be visible. 

**Expected behavior**
the visibility should match what you  se when you scrub the timeline in usdview or another application that can play a usd file

**Screenshots**

**Used Software Versions**
 - Arnold: [6.0.4.0]
 - USD:  0.19.11 and 0.20.05
 - Compiler: VS
 - OS: windows
 - Dcc: maya/mtoa

sebastienblor:
Porting the patch proposed by @kreppene in #543 , that takes into account the desired frame when asking for the visibility of an object. I'm including a test 141 to test this use case.

Fixes #543


sirpalee (approved):
LGTM, thanks for the quick update!

PR #550, opened 2020-09-24
Update the scons version

### LINKED ISSUE #549: Update Scons version
**Describe the bug**
We should update Scons so that it recognizes more recent versions of visual studio.
We also need to fix the way the environment is passed through Scons

sebastienblor:
I'm applying the patch proposed by @ansono , that updates the version of Scons, and adds `MSVC_VERSION` to the build parameters. We also change the way the environment is passed through Scons

Fixes #549


sirpalee (approved):
The change looks good, except modified two lines.

- When configuring the environment, the `tools = ...` line was also deleted, this breaks the documentation build.
- The dependency between the testsuite and the procedural is removed, I assume this was a local debug change accidentally left in. Looking at the core ticket, this wasn't a requirement.

PR #555, opened 2020-09-28
Viewport display of procedurals with overrides but no filename

### LINKED ISSUE #552: Viewport API not working with overrides-only and no filename
**Describe the bug**
A USD procedural can have an empty usd filename, but data written in the `overrides` attribute. This is properly handled for rendering, but not for the viewport API.

sebastienblor:
This PR prevents from doing an early out in the viewport API when there is no usd filename.
If an override is set, we still try to run it.
I'm adding test_0142 that exercises this use case

Fixes #552


sirpalee (changes requested):
The changes look good, but can we change test 142 to a unit test? The behavior of the Viewport API is better tested that way, no need to involve any rendering.

sirpalee (approved):

PR #557, opened 2020-09-28
Support deformation motion blur properly

### LINKED ISSUE #556: Deformation motionblur
**Describe the bug**
motion_start and motion_end is only set if the matrix of the prim is animated, 

this does not work for prims where the matrix is static but the prim is deforming. 

https://github.com/Autodesk/arnold-usd/blob/f20edb553dcc5b2faf29468d5b56f81c415dc125/translator/reader/utils.cpp#L64-L91

**To Reproduce**
Inn maya create a sphere, attach a cluster to the vertex points, and animate the cluster, 
now export the sphere to an usd file.

Bring the usd file back into maya using a standin, and put it next to the orginal sphere,
set these motionblur setting
image

**Screenshots**
image

**Used Software Versions**
 - Arnold: 6.0.4.0
 - USD: [e.g. 0.20.05]
 - Compiler: vs
 - OS: Windows
 - Any 3rd-party app: Maya

sebastienblor:
We're now setting the attributes `motion_start`, `motion_end` if there is deformation on meshes, curves, or points.
Added test_0143 following the repro steps from #556 , this PR's commit does solve the artefacts on such a scene.

Fixes #556


translator/reader/read_geometry.cpp
```diff
@@ -91,7 +91,13 @@ void UsdArnoldReadMesh::Read(const UsdPrim &prim, UsdArnoldReaderContext &contex
     }
 
     // Vertex positions
-    ReadArray<GfVec3f, GfVec3f>(mesh.GetPointsAttr(), node, "vlist", time);
+    UsdAttribute pointsAttr = mesh.GetPointsAttr();
+    ReadArray<GfVec3f, GfVec3f>(pointsAttr, node, "vlist", time);
```
sirpalee, line 95: The return value of ReadArray is not used anywhere in the code, maybe we could return the number of keys instead of array length, and use that to check if the attribute is animated?

Documentation of ValueMightBeTimeVarying:

```
    /// Return true if it is possible, but not certain, that this attribute's
    /// value changes over time, false otherwise. 
    ///
    /// If this function returns false, it is certain that this attribute's
    /// value remains constant over time.
```

sirpalee (changes requested):

translator/reader/read_geometry.cpp
sebastienblor: yep, makes sense. Also, ReadArray not only checks if the prim attribute might be varying, but also if we have motion blur in the reader options. This is more consistent now

sirpalee (approved):

PR #559, opened 2020-10-01
Inheriting from XFormable for shape, light and camera schemas

### LINKED ISSUE #558: Shape, Light and Camera schemas should inherit from Xformable instead of Typed
Shape, Light, and Camera schemas should inherit from Xformable, so they can be hidden in usdview and other USD outliners, and we can write USD imaging adapters for them.

sirpalee:
- Inheriting from XFormable for shape, light and camera schemas.

Fixes #558


sebastienblor (approved):
Now that #627 was implemented and this inheritance is reflected in our translators, we can finally merge this one

PR #566, opened 2020-10-05
Support animation on parent xforms

### LINKED ISSUE #565: Animated transforms not translated properly if set on parent xforms
**Describe the bug**
If a mesh is under a xform hierarchy, the xform has an animated transform but the mesh hasn't, it will render as a static geo.

This happens because  `ReadMatrix` checks if the prim itself is animated with `xformable.TransformMightBeTimeVarying()` but this doesn't account for parent xforms.

sebastienblor:
When we check whether a prim has animated transform or not, we need to check its parents recursively, until one is eventually animated. We only do this if motion blur is enabled.
I tried using `UsdGeomXformCache::TransformMightBeVarying(UsdPrim prim)` but it didn't give me a correct result. It returned false for my example scene, which I found suprising.

It seems more and more clear that we'll need to refactor the way we parse usd files, because USD itself isn't able to give us efficient informations along the hierarchy. We have this problem for visibility, transform, and also primvars (the later isn't fixed yet), and for each of them we're relying on different solutions. We should accumulate informations while traversing the stage, but this will have an impact on multi-threading and we'll probably have to do some deep changes.

I'm adding `test_0144` with a repro scene

Fixes #565


sirpalee (approved):
LGTM

PR #572, opened 2020-10-12
Fixed reading OSL shaders in the procedural

### LINKED ISSUE #570: USD procedural is not picking up osl shader parameters
The USD procedural is setting the code parameter, but not picking up the parameters.

See the example file [test.usda.zip](https://github.com/Autodesk/arnold-usd/files/5355267/test.usda.zip) where the `param_Scale` parameter is being ignored by the procedural.

sebastienblor:
Similarly to what was done on the render delegate for #571 , we're now doing a special case when reading OSL shaders in the translator. We first need to set the attribute "code" and only after we read all the other attributes.
We must also ensure that we update the node entry of this node after "code" was set.

Fixes #570


sebastienblor:
the commit also includes a fix on test_0050 that I spotted when creating a test here. It shouldn't change anything as we're overriding the output filename when running the tests but it's still better this way

sirpalee (changes requested):
test_0122 fails with these changes, however, I believe we are now getting the correct results. Could you please take a look at 122 and confirm it's working well now?

sirpalee (approved):

PR #576, opened 2020-10-15
Removing unnecessary widths query from HdArnoldBasisCurves::sync

### LINKED ISSUE #574: Incorrect display of curve widths in Solaris when changing curve basis
When rendering BasisCurves of type cubic, for proper smooth hair, we loose the per-point widths that are seen with linear curves and in its properties. This happens, deceivingly, after restarting the render.

sirpalee:
- Removing unnecessary widths query from HdArnoldBasisCurves::sync.

Fixes #574


sebastienblor (approved):

jhodgson (commented):
Looks good to me

PR #604, opened 2020-11-26
Add attribute cache_id to the procedural

### LINKED ISSUE #599: Add cacheId attribute in the procedural to load a usdStage in memory
If the procedural was built against a shared usd library, it could get the USD contents through the `UsdStageCache`  using a simple integer "id" value, instead of using a filename.

This would only make sense for shared USD libraries and shouldn't be exposed in any DCC plugin, so we might want to hide this attribute, or only declare it in certain build settings

sebastienblor:
This PR adds an attribute `cache_id` to the procedural. This attribute will have to be hidden from the plugins UI, as it does only make sense if the procedural was built against a shared usd lib, and that another usd plugin is filling the usd stage cache with the same ID.
If we ask to read a usd stage from an ID, then the filename and overrides aren't used. In this context we're not supposed to modify the usd stage as we're loading it from another source.
 
I'm also adding a build parameter `PROC_SCENE_FORMAT` that determines if the scene format plugin needs to be compiled or not (in this case for maya-usd we won't want to have it enabled as it could conflict with the builtin usd procedural).

A second step will be to implement `procedural_update` and to handle changes in specific primitives efficiently.

Fixes #599


sirpalee (approved):
Looks good!

The build time option was missing from cmake, so I quickly added that.

Issue #653, opened 2021-02-08
Support motion blurring of instanced shapes


sirpalee:
At the moment, when translating point instancers in the render delegate, we do not handle motion-blurred attributes, thus lacking transformation motion blur on generated instances.

PR #665, opened 2021-02-17
Initial take on motion blurring of the point instancer

sirpalee:
- Initial take on motion blurring of the point instancer.

**Additional context**
#653 is not fixed by this change, there are still differences when comparing to karma.

### DIFF
--- a/render_delegate/instancer.cpp
+++ b/render_delegate/instancer.cpp
@@ -30,19 +30,13 @@ TF_DEFINE_PRIVATE_TOKENS(_tokens,
 
 namespace {
 
-template <typename T>
-inline const VtArray<T>& _LookupInstancePrimvar(const HdArnoldPrimvarMap& primvars, const TfToken& primvar)
+template <typename IN, typename OUT>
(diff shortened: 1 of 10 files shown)


render_delegate/instancer.cpp
```diff
@@ -76,106 +87,173 @@ void HdArnoldInstancer::_SyncPrimvars()
...
     // TODO(pal): Add support for inheriting primvars from parent instancers.
     VtIntArray instanceIndices;
-    for (const auto& primvar : _primvars) {
-        const auto& desc = primvar.second;
-        if (desc.interpolation != HdInterpolationInstance || !desc.dirtied || primvar.first == _tokens->rotate ||
-            primvar.first == _tokens->translate || primvar.first == _tokens->scale ||
-            primvar.first == _tokens->instanceTransform) {
+    for (auto& primvar : _primvars) {
+        auto& desc = primvar.second;
+        if (!desc.NeedsUpdate()) {
```
sirpalee, line 256: This is causing primvars to not work when there are multiple prototypes per instancer. We can't rely on the new NeedsUpdate function here.

PR #676, opened 2021-02-26
Read velocities from point based schemas

### LINKED ISSUE #221: Create motion keys using the velocity attribute
To implement velocity motion blur when the velocity attribute is present in the usd file, we should create two motion keys based on the velocity and the position as we do for example in the alembic procedural and when exporting geometry in HtoA.

### LINKED ISSUE #564: Changing topology
**Describe the bug**
When rendering a prim with changing topology,  only the first frame of the usd file works, 
on other frames i get messages like this
 Error: [polymesh] p_liquidMilk_001/p_liquidMilk_001Shape: wrong number of vertex indices (found 189872, expected 200200) // 

we should also make sure we use velocities when avalible

**To Reproduce**
Try rendering different frames of a usd file with changing topology, 

**Used Software Versions**
 - Arnold: 6.0.4.0
 - USD: 0.20.05
 - Compiler: vs
 - OS: Windows 10

sebastienblor:
In this PR, for point based schemas (mesh, curves, points), I read the velocity attribute before reading the positions.

If velocities are found, I just read the positions at the current frame, and using velocities I extrapolate it for 2 keys corresponding to shutter start and shutter end.

Otherwise, we read the positions as usual, with eventually multiple motion keys.

As a side change, I also ensure that we don't try to read motion keys for attributes like nsides and vidxs

Fixes #221 
Fixes #564


sirpalee (approved):

bsilvaAX:
Fantastic, thank you for this!

PR #682, opened 2021-03-04
Prevent crashes in the writer by solving invalid naming conventions

### LINKED ISSUE #510: Crashing case when exporting
Attaching the repro code [UsdStapler.cpp.txt](https://github.com/Autodesk/arnold-usd/files/5096345/UsdStapler.cpp.txt) that loads this .ass [stapler.ass.zip](https://github.com/Autodesk/arnold-usd/files/5096340/stapler.ass.zip) and tries to export it to usd.
(I had to rename the files to be uploadable here).
The .ass kicks without problems.
In the cpp, just replace the top defines to match your local paths.

This reproduces what reported in [MAXTOA-834](https://jira.autodesk.com/browse/MAXTOA-834), crashing with the following:

usd_stapler_crash

### LINKED ISSUE #615: Usd Writer crashes when node name contains hyphen character
Exporting and Arnold scene to USD crashes when a node name does not match the USD naming conventions.
There are two specific cases reported, but could be more:

1. Name contains hyphen character (-)
2. Name starts with a number

sebastienblor:
Similarly to what was already done for other non-supported characters (`|`, `:`, `.`, `@`) , we now also replace `-` with undescores to prevent USD from crashing.
Also, we now ensure that the first character after each `/`  is *not* a digit, which is also causing crashes when writing to USD.

Given that we had to add a loop over all characters for the last verification, I've moved all the character replacements in this single loop, to avoid multiple calls to `std::replace` 

Added test 160 with crazy names

Fixes #615 
Fixes #510


sjannuz:
When you change the names like that there is a small change you will have identical names in the end. In my patch for MAXtoA, I also check if the resolved name is not in the universe already, and, in case, add I don't remember what, until the name is not found existent yet.

sirpalee (approved):

PR #689, opened 2021-03-08
Support primvars on xform ancestors #282

### LINKED ISSUE #282: Primvars are not inherited from ancestor primitives
**Describe the bug**
When translating primitives in the procedural inherited primvars are not correctly translated.

**To Reproduce**
1. Create a polymesh with a parent xform primitive
2. Author subdivision settings on the parent xform using primvars
3. Render using the Arnold procedural and see how the primvars do not affect the polymesh

**Expected behavior**
The primvars authored on ancestor primitives should affect the primitive being translated.

**Additional context**
`UsdGeomPrimvarsAPI` provides functions to iterate primvars coming from ancestor prims, like `UsdGeomPrimvarsAPI::FindInheritablePrimvars`.

### LINKED ISSUE #382: Primvars with namespaces aren't loaded
**Describe the bug**
When a user data has column characters, they're not loaded in the usd reader.
Currently, we ignore all user data that have a namespace, because we don't want to load attributes meant for another renderer.

This is causing a problem with the alembic procedural, because it can have user data like `polymesh:subdiv_type` . With such a user data, all the meshes created by the procedural will have `subdiv_type` set to the desired value. 

We could decide to consider all primvars and convert them as user data, with any namespace. At least we won't loose any information when converting to and from USD

### LINKED ISSUE #579: Subdivision settings not being passed to usd procedural
A scene using subdivsion working in Solaris is not displaying subdivisions in the usd procedural.

The example file [subdiv_test.usda.zip](https://github.com/Autodesk/arnold-usd/files/5408381/subdiv_test.usda.zip) is exported from Solaris and contains subdiv settings, but these aren't preserved when converting to .ass file with `kick -resave` for example and the procedural doesn't expand the polymesh with the subdiv settings preserved.

sebastienblor:
This PR adds support for parent primvars. 
We need to change the traversal to be "pre-and-post" meaning that every primitive is found twice. This allows us to keep track of the "current" set of primvars at every point of the hierarchy.
At every xform primitive, we compute the current set of primvars, accounting for its ancestors. In theory this could also happen for other primitive types, but currently happens in practice for xform prims.
When we get the primvars list, we call `UsdGeomPrimvarsAPI::FindIncrementallyInheritablePrimvars`  that takes the parent xforms as input, and avoids excessive duplication of calls to `GetPrimvars` all over the hierarchy.

There will be a small overhead after this PR,  because of the "pre-and-post" traversal, plus the additional calls to get the primvars. On multi-thread loading, this increases the duplicated work across threads which is not ideal. A revamping of the multi-thread traversal strategy will be needed, but we can defer that to a separate task.

test 164 is based on the repro scene from #282 

While implementing it, I faced a crash in test_0131 that I fixed as a separate ticket #688 . **So 688 will need to be merged before this PR.**

Note that this PR is needed to fix #579.
It might also fix #382 , but that will need to be verified.
 
Fixes #282


translator/reader/reader.cpp
```diff
@@ -158,25 +159,74 @@ unsigned int UsdArnoldReader::RenderThread(void *data)
...
+        if (objType.empty())
+            continue;
+
+        // We traverse every primitive twice : once from root to leaf, 
+        // then back from leaf to root. We don't want to anything during "post" visits
+        // apart from popping the last element in the primvars stack for xforms.
+        // This way, the last element in the stack will always match the current 
+        // set of primvars
+        if (iter.IsPostVisit()) {
+            if (objType == "Xform")
```
sirpalee, line 192: I don't think checking if objType is "Xform" is the right approach here, UsdGeomPRimvarsAPI can apply to almost anything, not just Xforms. For example, UsdGeom also defines "Scope" which can also hold primvars and can be used to group objects together. Also, meshes can have other meshes embedded, USD allows for a fully flexible scenegraph.

I think it's easier just to initialize UsdGeomPrimvarsAPI and check if it's valid. (ie cast it to a bool, like `if (primvarsAPI) ...` )

translator/reader/reader.cpp
```diff
@@ -158,25 +159,74 @@ unsigned int UsdArnoldReader::RenderThread(void *data)
...
+            
+            // if the returned vector is empty, we want to keep using the same list as our parent
+            if (primvars.empty())
+                primvarsStack.push_back(primvarsStack.back());
+            else
+                primvarsStack.push_back(primvars); // primvars were modified for this xform
+
+            checkVisibility = true;
+        } else {
+            // Special case for arnold schemas, they don't inherit from UsdGeomImageable
```
sirpalee, line 213: Is this still required even after #558 ?

sirpalee (changes requested):

translator/reader/reader.cpp
sebastienblor: when we'll release the schemas, I guess we'll be able to remove this hack. But for now, I don't think USD can be aware of this inheritance

translator/reader/reader.cpp
sebastienblor: yeah I know.... what I was trying to do here is to only compute this uselessly for non-leaf primitives.
The docs say 
```
" It is not generally useful to call this method on UsdGeomGprim leaf
    /// prims, and furthermore likely to be expensive since *most* primvars
    /// are defined on Gprims."
```
but I haven't found an efficient way to know if a prim has children or not (I need to call it on all prims during traversal).

That being said.... for leaf primitives, we'll usually call the same function again in `_ReadArnoldParameters` to read the `primvars:arnold` attributes. So another approach could be to always do it in the loop, and then in `_ReadArnoldParameters` just read the previously computed set of primvars. I'll try that

sebastienblor: Pushed the change. We now compute the primvars list for every prim type during the traversal.
Then in `ReadArnoldParameters`  (which is called for nearly all primitives) I just reuse this last element of the stack instead of recomputing it.

sirpalee (approved):

PR #700, opened 2021-03-10
Adding a set of common strings to use between the different modules

sirpalee:
- Adding a set of common strings to use between the different modules.
- Replacing static/on-the-fly created TfTokens and AtStrings in the reader module where it makes sense.
- Added a new function to translate Usd/Hydra creases to Arnold.


sebastienblor (approved):
that's awesome ! thanks for starting this

PR #701, opened 2021-03-10
Support curves width vertex interpolation in the procedural

### LINKED ISSUE #244: Curves with vertex interpolation on width (procedural)
**Describe the bug**
As discussed in #242 , Curves with vertex interpolation on width aren't supported in Arnold. 
Arnold curves fail with this message :
```
00:00:00    98MB WARNING |   [curves] /Cubic/Ribbons/VertexWidth: invalid radius specification: found 7, expected 1, 1 or 3 (constant, uniform or varying); just using the first
00:00:00    98MB WARNING |   [curves] /Cubic/Tubes/VertexWidth: invalid radius specification: found 7, expected 1, 1 or 3 (constant, uniform or varying); just using the first

```
We should convert the vertex interpolation into a varying one somehow, so that at least the curves can be rendered.

**To Reproduce**
run test_0033. Two curves are missing (a purple one in the middle and a yellow one at the right), and the error message above appears.

sebastienblor:
This PR extracts the render delegate code to remap the curves primvars, and puts it in the shared utils. The procedural now uses
this same code too, which adds support for curves widths with vertex interpolation.

The code to handle curves primvars is now done through a class `ArnoldUsdCurvesData`, that initializes on demand the remapping between arnold and USD, in the case of varying and vertex interpolations.

I've had to add a test to allow the testsuite to succeed. When the input widths size is equal to the expected one we don't do any remapping. The fact that we have failures without this (we get infs and nans) might imply that there's an issue in the remapping code, so it's maybe something to investigate

```
if (_numPerVertex == original.size()) {
      // The input value size already matches what we're targetting. There's no
      // need to do any remapping
      return true;
}
```
test 171 verifies different combinations of interpolations
in test 33, 2 curves were not showing up because of the vertex interpolation limitation. I've removed them from this test, to verify that the test still succeeds. It's now tested in test 171

Fixes #244


sebastienblor:
I think the problem I had before the early out I added was because the code expects a specific amounts of widths (as per USD standards) and does the remapping for that size, even if that size is different from the one we receive as input.
Maybe we could make it so that the remapping works whatever input we get, since we know the input size and we know the output size.

sirpalee:
Added some changes:
-fix build issues on windows (for the unit tests)
- fixed up test 136 to use the new functions instead of the old ones
- ran clang-format 
- added documentation for the functions in common
- added support for half/half array in the radius set functions

PR #709, opened 2021-03-15
Use a WorkDispatcher for parallel reading of the procedural

### LINKED ISSUE #690: Improve multi-thread reading of the procedural

The strategy for multi-thread traversal is not ideal and could be improved.
Plus, it requires an explicit thread count, and is currently disabled by default.
With #689 we have increased the amount of work duplicated in each thread, since they all need to compute the set of primvars for all primitives, even though that won't be needed.

Among the solutions to investigate, we could eventually use USD's `WorkDispatcher` to send multiple jobs without a hardcoded thread count. If the traversal is always done in the same thread and the actual conversion is sent to the dispatcher, we could also use this to have a single xform cache instead of one per thread.

sebastienblor:
This PR introduces a [WorkDispatcher](https://github.com/PixarAnimationStudios/USD/blob/release/pxr/base/work/dispatcher.h) from usd "work" lib, that is used when the procedural's `threads` is set to 0.
In that case, instead of duplicating the stage traversal over N threads, this will only be done by one thread. Every time a primitive needs to be read, it will run a dispatcher job, that will be picked by another thread when available to read this primitive.
The advantages of this approach are : 
- we don't need to set a specific amount of threads. If this solution is proved successful, we can end up removing the parameter threads to force this behaviour
- the work done during the traversal (accumulating the primvars, checking the visibility, purpose, instanceable, etc...) is only done once and not once per thread. When lots of procedurals are initialized in parallel by arnold, we can't be sure that the threads spawned by the reader will really perform in parallel, so this duplication can end up slowing down things
- there is a single xform cache, not one per thread, the matrices are computed during the traversal.
- this solution should perform well in different use cases (simple and complex usd files, many usd procedurals, a single big one, etc...)
- this solution should perform well on different systems, whatever its amount of cores is.
- this relies on TBB scheduler which is highly optimized

I'm setting the procedural's `threads` to 0 given the good results I'm getting in my tests. For now I'm keeping this parameter, so that the previous algorithm can be used in case of problems, and to be able to keep doing benchmarks on different scenes.
But in the future, we should be able to remove the parameter, and the previous code. This will end up simplifying quite a lot the current code, and we won't need the "Reader Threads" anymore.

Since it's the default value in this PR, all tests are exercising this code. I'm making it so that the kitchen scene does too (threads used to be hardcoded to 10 in this test scene)

Benchmarks I've done in the comment below.

Fixes #690


sebastienblor:
Here are the different timings I've had on several scenes. They all refer to the "node initialization" times, after running multiple times the same renders. I'm adding the precision margins based on the differences I see on consequent renders.

1) Kitchen scene : 1790 of small & simple objects. (Precision around 0.03s)
- 1 thread (previous default) = 0.67s
- 5 threads = 0.35s
- 10 threads = 0.34s
- 20 threads = extremely variable, from 0.5s to 4.0s
- 0 threads (new default with the dispatcher) = 0.3s

2) 30 Robots scene, in one usd file exported from Maya: 900 objects, some of them quite detailed (1309769 triangles for one object) and some much simpler. (Precision = 0.2s)
- 1 thread (previous default) : 9.4s
- 5 threads : 5.5s
- 10 threads : 4.1s
- 20 threads : 4.4
- 0 threads (new default with the dispatcher) : 4.5s

3) 10 usd procedurals, loading each one a usd scene with 1 robot from the scene above. (Precision around 2s)
- 1 thread (previous default) : 26s
- 5 threads : 27s
- 10 threads : 27s
- 20 threads : 31s
- 0 threads (new default with the dispatcher) : 26s

sebastienblor:
Conclusions of the tests above: 
With the previous algorithm, depending on the scene we could get a speedup when increasing from 1 thread to a few threads, then we get a plateau when the amount of threads increase and then at some point the render times become worse, sometimes in a dramatic way. The "ideal" amount of threads is somewhere between 3-6 but it's hard to find a value that works consistently. Also, with lots of procedurals, having more threads doesn't help, and even can slow down things.

Using the dispatcher (with threads=0) seem to be working well consistenly in the different scene, usually around the best timings that we get with the "ideal" amount of hardcoded threads. We're slightly slower in the 30-robots scene that with threads=10, but the difference is acceptable given the benefits and the timing in other different scenes.

Also, interesting to note that 11 procedurals loading a usd file with 1 robot is about 6 times slower than a single usd file with 30 of these robots.

sirpalee:
Pushed back some changes to fix compilation issues on Linux (scons + cmake and two simple c++ changes), could you take a look at it and confirm it's working as expected on your side?

sebastienblor:
it builds and runs well for me on windows, with both static and shared libs

sirpalee (approved):

PR #723, opened 2021-03-24
Fixing build issues on macOS and targeting Houdini USD libraries

### LINKED ISSUE #722: Failing to generate schemas when targeting Houdini on macOS
**Describe the bug**
There are two issues when building arnold-usd using the USD libraries shipped with Houdini.
- usdGenSchema tries to use hython, so we need to add USD_BIN to the PATH when running usdGenSchema
- The Arnold bindings in the PYTHONPATH are causing an issue when using usdGenSchema

sirpalee:
- Clearing PYTHONPATH before running usdGenSchema
- Adding USD_BIN to the PATH before running usdGenSchema

Fixes #722

### DIFF
--- a/schemas/SConscript
+++ b/schemas/SConscript
@@ -80,6 +80,11 @@ FIXME Current issues to address the generation of the schema files :
  - when I build the schemas, I get "Warning C4273: inconsistent linkage" for every function
 '''
 
+myenv = env.Clone()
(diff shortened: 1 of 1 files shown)


sebastienblor (approved):

jhodgson (commented):
Looks good

Issue #775, opened 2021-04-29
Remove deprecated functions, warnings, and fix some bugs


jhodgson:
The OSX build of the render delegate is giving a bunch of warnings which should be addressed, and we want to remove deprecated function use in preparation for more deprecated functions.

jhodgson:
**Bug fixes**
* HdArnoldShape::Sync() should use logical & to test bits
* _GetTokenFromRenderBufferType should not return a local reference
* Use material instead of id  when tracking material changes

**Deprecation**
* Call AiNode() with the universe pointer, enabled by adding the HdArnoldRenderDelegate to the lightSync callbacks and SyncParams function
* Use AtString instead of const char* in API

**Warnings**
* Add override keyword to Material::Reload virtual function
* Comparison should use the same type
* Reorder some class member variables to correspond with constructor initialisation order

**Clean up**
* Adding more strings to the constant_strings.h cache
* Removed unused function _GetStringFromValue

PR #776, opened 2021-04-29
Remove deprecated functions, warnings, and fix some bugs (#775)

jhodgson:
See #775

**Bug fixes**
* HdArnoldShape::Sync() should use bitwise & to test bits
* _GetTokenFromRenderBufferType should not return a local reference
* Use material instead of id  when tracking material changes

**Deprecation**
* Call AiNode() with the universe pointer, enabled by adding the HdArnoldRenderDelegate to the lightSync callbacks and SyncParams function
* Use AtString instead of const char* in API

**Warnings**
* Add override keyword to Material::Reload virtual function
* Comparison should use the same type
* Reorder some class member variables to correspond with constructor initialisation order

**Clean up**
* Adding more strings to the constant_strings.h cache
* Removed unused function _GetStringFromValue


common/shape_utils.h
```diff
@@ -78,7 +78,7 @@ class ArnoldUsdCurvesData {
         }
 
         const auto& original = value.UncheckedGet<VtArray<T>>();
-        if (_numPerVertex == original.size()) {
```
sirpalee, line 81: Please use c++ style casts. See https://github.com/Autodesk/arnold-usd/blob/master/docs/conventions.md

render_delegate/light.cpp
```diff
@@ -184,14 +184,14 @@ AtString getLightType(HdSceneDelegate* delegate, const SdfPath& id)
 }
 
 auto spotLightSync = [](AtNode* light, AtNode** filter, const AtNodeEntry* nentry, const SdfPath& id,
-                        HdSceneDelegate* delegate) {
```
sirpalee, line 187: Please use camelCase instead of snake_case when naming parameters. See https://github.com/Autodesk/arnold-usd/blob/master/docs/conventions.md

render_delegate/light.cpp
```diff
@@ -400,9 +401,9 @@ class HdArnoldGenericLight : public HdLight {
     AtNode* _light;                    ///< Pointer to the Arnold Light.
     AtNode* _texture = nullptr;        ///< Pointer to the Arnold Texture Shader.
     AtNode* _filter = nullptr;         ///< Pointer to the Arnold Light filter for barndoor effects.
+    bool _supportsTexture = false;     ///< Value indicating texture support.
```
sirpalee, line 404: Please change the constructor initialization order instead of changing the parameter order on the class. I generally order parameters by size on structs like this.

render_delegate/light.cpp
```diff
@@ -641,7 +642,6 @@ class HdArnoldSimpleLight : public HdLight {
     HdDirtyBits GetInitialDirtyBitsMask() const override;
 
 private:
-    HdArnoldRenderDelegate* _delegate; ///< Pointer to the Render Delegate.
```
sirpalee, line 644: Let's just get rid of simple light, we don't need it anymore. This is just a leftover from older times, nowadays scene delegates should supply us with correct, UsdLux light types.

render_delegate/render_buffer.h
```diff
@@ -110,8 +110,8 @@ class HdArnoldRenderBuffer : public HdRenderBuffer {
 
     /// Utility class for storing render buffers.
     struct BufferDefinition {
-        HdAovSettingsMap settings;              ///< Filter and AOV settings for the Render Buffer.
         HdArnoldRenderBuffer* buffer = nullptr; ///< HdArnoldRenderBuffer pointer.
+        HdAovSettingsMap settings;              ///< Filter and AOV settings for the Render Buffer.
```
sirpalee, line 114: Same about here about the parameter ordering vs fixing the ctor.

render_delegate/shape.cpp
```diff
@@ -44,7 +44,7 @@ void HdArnoldShape::Sync(
         param.Interrupt();
         _SetPrimId(rprim->GetPrimId());
     }
-    if (dirtyBits | HdChangeTracker::DirtyCategories) {
+    if (dirtyBits & HdChangeTracker::DirtyCategories) {
```
sirpalee, line 47: Nice catch!

sirpalee (changes requested):
The change looks good! Thanks for the bug fixes and cleanup.

I had a few small requests to better align with existing code and our coding conventions, but otherwise, we are good to go.

render_delegate/material.h
```diff
@@ -76,7 +76,7 @@ class HdArnoldMaterial : public HdMaterial {
     ///
     /// Currently does nothing.
     HDARNOLD_API
-    void Reload() {}
+    void Reload() override {}
```
sirpalee, line 79: I remember now, this was marked non-override on purpose. Newer USD versions don't provide this function anymore. It is missing in 20.11 (https://github.com/PixarAnimationStudios/USD/blob/v20.11/pxr/imaging/hd/material.h) but = 0 in 20.08 https://github.com/PixarAnimationStudios/USD/blob/v20.08/pxr/imaging/hd/material.h. We can wrap in #if PXR_VERSION < 2011 , that should solve your warning and still work with newer USD versions.

PR #782, opened 2021-05-03
Write multiple frames to a single usd file

### LINKED ISSUE #777: Writer could append multiple frames to a usd file
Arnold universes only handle a single frame, so currently we can only export one arnold "frame" to a single usd file through the writer.

It would be useful to append another universe (for another frame) to an already existing usd file, so that we could generate a full sequence on a single usd file.

sebastienblor:
This PR is a first support to append several frames in a single usd file.
In the write parameters, we can specify a float "frame" value that will allow to set the desired frame in the usd file.
This information is currently stored in the options "frame" attribute, which is meant for that currently in arnold scenes.

When an arnold universe for a following frame is saved in the same file, in addition to the frame we also need a boolean parameter "append". It means that instead of creating a fresh new usd file, we want to reuse the existing one and append data to it. 

When a parameter animated, this will end up as time samples on this usd prim. Constant attributes should appear as a simple value.
The logic is that as long as the attribute value remains unchanged, it's authored as a constant value, but as soon as it changes it relies on time samples. When that happens for the first time, we set a time sample for the previously exported "surrounding frames" with the previous value, so that the animation is correct. We get the list of previously exported frames in "options.frame" that contains them all. In this version, as soon as an attribute is animated, it will be stored as time samples, whatever the values are. In the future we could improve this, and avoid setting sequential time samples with identical values.

To do all this, we now use a wrapper function `UsdArnoldWriter::SetAttribute` that will take care of all of the above. 

test_0178 sequentially loads 3 ass files representing a sequence and then verifies that loading it at different frames return the proper value.

Fixes #777


sirpalee (changes requested):
The commits fail to build on Linux with the following error (and there is one for every use of the function):
```
translator/writer/write_geometry.cpp: In member function ‘virtual void UsdArnoldWriteMesh::Write(const AtNode*, UsdArnoldWriter&)’:
translator/writer/write_geometry.cpp:44:78: error: no matching function for call to ‘UsdArnoldWriter::SetAttribute(arnoldPxrInternal__pxrReserved__::UsdAttribute, const arnoldPxrInternal__pxrReserved__::TfToken&)’
 writer.SetAttribute(mesh.GetOrientationAttr(), UsdGeomTokens->rightHanded);
                                                                          ^
In file included from translator/writer/prim_writer.h:27:0,
                 from translator/writer/write_geometry.h:24,
                 from translator/writer/write_geometry.cpp:14:
translator/writer/writer.h:100:10: note: candidate: template<class T> void UsdArnoldWriter::SetAttribute(arnoldPxrInternal__pxrReserved__::UsdAttribute&, const T&, float*) const
     void SetAttribute(UsdAttribute &attr, const T& value, float *subFrame = nullptr) const
          ^~~~~~~~~~~~
translator/writer/writer.h:100:10: note:   template argument deduction/substitution failed:
translator/writer/write_geometry.cpp:44:48: note:   cannot convert ‘arnoldPxrInternal__pxrReserved__::UsdGeomGprim::GetOrientationAttr() const()’ (type ‘arnoldPxrInternal__pxrReserved__::UsdAttribute’) to type ‘arnoldPxrInternal__pxrReserved__::UsdAttribute&’
     writer.SetAttribute(mesh.GetOrientationAttr(), UsdGeomTokens->rightHanded);
```
This happens because the first argument of the function is a reference, and using the generic GetXXX functions on schema classes returns an UsdAttribute by value. (like `UsdAttribute GetOrientationAttr() const;`) I assume that MSVC is letting you get away with this, but it shouldn't.

We can either take UsdAttributes in the function by value (which means trivial copying of a `UsdObjType`, `Usd_PrimDataHandle`, `SdfPath` and `TfToken` types), or by const reference. I suspect copying is avoided in both cases because it's a templated function defined in a public header (so it's very likely to be inlined). Maybe const reference is the safer option, I'm not sure.
A super simple example here: https://godbolt.org/z/vYKeh5jsf, based on this both approaches seem to generate the same code.

It's a bit counter-intuitive to pass UsdAttribute by const ref, but in USD all the UsdAttribute functions are const since UsdAttribute only holds a handle/path to the attribute (which can't be changed), not the actual value.

Most of the USD functions with a UsdAttribute parameter seem to take them by const ref, most importantly sparseValueWrite is doing the same as well: https://github.com/PixarAnimationStudios/USD/blob/090ef0d849ced875b3b99a15e454b69148ccd8e1/pxr/usd/usdUtils/sparseValueWriter.h#L237

sebastienblor:
Right, I changed the function to pass a const ref, let me know if it builds this way.

By the way, I forgot to mention the sparseValueWrite.... It's the first thing I tried in my wrapper function, but it didn't do the job unfortunately. If I understand properly, it's meant to be created once for an attribute, to then write values at different frames. 
In the current framework, each frame is appended in a completely different process, so using a sparseAttrValue in the wrapper would just create a time sample for each frame.

sirpalee (approved):

sirpalee:
Thanks for the updates!

Issue #783, opened 2021-05-04
Use the render session API to support multiple render sessions


jhodgson:
Use the render session API to allow concurrent render sessions of different universes.  This will also prevent conflicts with other host DCCs that may have been also using the main universe.

PR #784, opened 2021-05-04
Translate to a new universe and use a new render session (#783)

jhodgson:
See issue: Use the render session API to support multiple render sessions #783.

Use the render session API to allow concurrent render sessions of different universes. This will also prevent conflicts with other host DCCs that may have been also using the main universe.


sirpalee (approved):
The change looks good, as discussed we are holding back merging it.

kikou:
Sorry to be late to the party but we should not define `AI_MULTIPLE_RENDER_SESSION` in the code as it might collide with core definitions. The `AI_` prefix is reserved for Arnold core defs.

PR #786, opened 2021-05-05
Fixing compilation issues when using USD 21.02 and later introduced in #775

sirpalee:
- Fixing compilation issues when using USD 21.02 and later.

**Additional context**
Additional fix for #775


sirpalee:
#771 included the same fixes as in this PR. We can close this PR without merging.

PR #806, opened 2021-06-16
Python 3 build for Render Delegate

jhodgson:
Change requires to build the render delegate using python 3


jhodgson:
Closing this for now as this is work in progress and are looking to use CMake moving forwards.

PR #813, opened 2021-06-22
Using a custom AOV to store hydra prim IDs instead of the built-in ID AOV

### LINKED ISSUE #812: Use a hydra specific prim ID buffer instead of the built-in ID AOV
We should use a hydra-specific prim ID buffer instead of the built-in ID AOV when filling out the primID render buffer. This has the advantage of being able to overwrite IDs inside a procedural, which is not uniformly supported for the ID AOV.

sirpalee:
- Using a custom AOV to store hydra prim IDs instead of the built-in ID AOV.
- Moving the string definitions from `render_delegate/utils.cpp` to the shared library.
- Exposing the function to declare user parameters on shapes.

Fixes #812


jhodgson (approved):
Looks good to me.

PR #826, opened 2021-07-05
Read aov name from RenderVars for merged AOVs

### LINKED ISSUE #816: The procedural does not use "driver:parameters:aov:name"
**Describe the bug**
When setting up rendervars in Solaris, `driver:parameters:aov:name` on a rendervar is used to overwrite the AOV name outputted to the Render Product. This way it's possible to use two Render Vars that point to the same source.

Attaching an [example archive](https://github.com/Autodesk/arnold-usd/files/6731017/duplicate_aov_bug.zip), that includes a USDA file, and the EXR outputted through husk/render delegate and kick.

sebastienblor:
The procedural was reading the aov name (driver:parameters:aov:name) only when a RenderVar had a light path expression.
This PR now always reads this parameter and uses it if we have more than one render var for the same driver.
Added test_0180 that verifies the same render vars than the repro scene. Since we can't test merged exr images, I'm explicitely checking what options.outputs we read from the input usd file, and failing the test if it's not as expected.

Fixes #816


sirpalee (approved):
Look good!

PR #863, opened 2021-08-11
Cleaning up AiOutputIteratorGetNext changes

### LINKED ISSUE #845: Update AiOutputIteratorGetNext usage
AiOutputIteratorGetNext will now take AtString instead of const char* arg.

sirpalee:
- Cleaning up AiOutputIteratorGetNext changes.

Fixes #845

### DIFF
--- a/render_delegate/nodes/driver_main.cpp
+++ b/render_delegate/nodes/driver_main.cpp
@@ -117,7 +117,7 @@ driver_process_bucket
 #if ARNOLD_VERSION_NUMBER > 60201
         return outputName == name;
 #else
-        return strcmp(outputName, name.c_str()) == 0;
+        return strcmp(outputName, name) == 0;
 #endif
     };
     while (AiOutputIteratorGetNext(iterator, &outputName, &pixelType, &bucketData)) {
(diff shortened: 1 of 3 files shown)


render_delegate/nodes/driver_main.cpp
```diff
@@ -113,18 +113,25 @@ driver_process_bucket
     ids.clear();
     const void* colorData = nullptr;
     const void* positionData = nullptr;
+    auto checkOutputName = [&outputName](const AtString& name) -> bool {
+#if ARNOLD_VERSION_NUMBER > 60201
+        return outputName == name;
+#else
+        return strcmp(outputName, name.c_str()) == 0;
```
ThiagoIze, line 120: I think you might be able to do the following simplification since `AtString` is usually automatically converted to `char*`:
```
        return strcmp(outputName, name) == 0;

ThiagoIze (approved):
Looks good. If you want, I think there might be one more cleanup you can do, but I wouldn't say it's mandatory -- your choice.

render_delegate/nodes/driver_main.cpp
sirpalee: Good point, thanks for the advice! Changed.

PR #864, opened 2021-08-11
Validating facevarying indices of primvars.

### LINKED ISSUE #853: Missing indices for facevarying primvars in Hydra abort renders
**Describe the bug**
When indices are missing from facevarying primvars the render delegate does not validate the incoming data and can potentially put Arnold in an "aborted" state, which will stop renders from working.

sirpalee:
- Validating facevarying indices of primvars.

Fixes #853


sebastienblor (approved):
not urgent, but the function _ConvertValueToArnoldParameter looks like something that could perhaps be in the common libs, so that maybe we could reuse the logic in the procedural ?

Issue #462, opened 2020-06-15
Default lighting not showing up in Solaris


sirpalee:
**Describe the feature you'd like**
When using the render delegate in Solaris, the default camera light does not affect the scene. This is causing confusion, as assigning a standard surface to any object will result in black renders without a light, yet in other render delegates, like karma, there is a default light.

sirpalee:
Usdview and Solaris handle default lighting differently.

Usdview has two options in the menu, enabling a camera and a dome light. When these lights are enabled, a vector of GlfSimpleLights is passed to UsdImaging, which will be converted to Hydra primitives by the HdxTaskController. However, the `_simpleLightTask` is used to convert these lights to hydra primitives. However, this task is only created when HdStorm is the rendering backend, which means we won't get this information in the Arnold render delegate.

In Solaris, when the camera light is enabled, there is no hydra primitive created, but a Houdini specific render setting is passed to the render delegate via `HdRenderDelegate::SetRenderSetting`. It's called `houdini:headlight`, which tells the render delegate if a headlight should be displayed or not. Based on this value, the render delegate can create a distant light aligned with the active camera. This is relatively straightforward to implement, however, updating a distant light when the camera moves, would disable the fast node updates in Arnold and would introduce a significant delay to the first pixel when using the GPU backend.

PR #867, opened 2021-08-12
Changing the fallback shader to standard surface.

### LINKED ISSUE #861: Improve fallback shader
At the moment the fallback shader in case of missing shader assignments is a utility shader with ambient occlusion and reading from the displayColor user data. While this works well if we want to preview assets in an environment without lights, it can make scenes lighter which has a mixture of objects with materials, objects without materials and lights.

It would be good to use a default standard_surface shader, that falls back to the previous ambient occlusion shader if no lights are present in the scene. Alternatively, we could just use a simple standard surface shader and find a solution to support the default lights in usdview / solaris.

sirpalee:
- Changing the fallback shader to standard surface.

Fixes #861

**Additional context**
Forcing the fallback shader on the ALAB scene.
Using the current utility shader:
alab_utility
After the PR:
alab_standard_surface
We might want to wait until #462 is fixed, so in case of no lighting in the scene, we can still see something.


sebastienblor (approved):
that's consistent with what the procedural does

PR #870, opened 2021-08-13
Avoid updating the main driver matrices in case of camera movement

### LINKED ISSUE #869: Don't update the main driver matrices if only the camera moves
We shouldn't update the matrix values on the main camera, so Arnold can use a faster update path for camera only movement.

sirpalee:
- Avoid updating the main driver matrices in case of camera movement.

Fixes #869

### DIFF
--- a/render_delegate/nodes/driver_main.cpp
+++ b/render_delegate/nodes/driver_main.cpp
@@ -31,8 +31,9 @@
 #include <memory>
 
 #include <constant_strings.h>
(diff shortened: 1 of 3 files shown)


ThiagoIze (approved):
If the matrices are not used in node_update, then sidestepping AiNodeSetMatrix by directly modifying the matrices will let arnold know that when just moving the camera no other nodes need a node_update, and so we can do the camera IPR fast path.

PR #872, opened 2021-08-18
Enforce writing multiple frames when no default is authored

### LINKED ISSUE #871: Enforce writing multiple frames when no default is authored
When writing multiple frames to the same usd file, we can get missing data if the first frame was not authored.
This happens for example if the matrix of the first frame is the default identity matrix. It won't be authored in the first place, and then for the second frame it won't manage to get the previous value and won't write the new one.
This will lead to an empty transform attribute.

sebastienblor:
When appending matrix attributes to existing usd files, if no value was previously authored, it means that it was originally equal to identity matrix. In that case we ensure that a default value is set before calling `writer.SetAttribute`.
In this later function, we also ensure that we set the attribute value if no value had been previously authored.

Fixes #871


sirpalee (approved):
LGTM

PR #875, opened 2021-08-23
Expose additional path for schemas generation

### LINKED ISSUE #874: Allow to prepend PATH folders when building schemas
When we build the arnold-usd schemas, we need to run `usdGenSchema` that links against python and also needs to run the python command. If we need to use a specific python distribution that is not set by default in the system's PATH, this fails.

A way to fix this would be to expose a new variable with a folder that is meant to be prepended to the PATH before running usdGenSchema

sebastienblor:
This PR adds an optional build variable `SCHEMA_PREPEND_PATH`. When set, the given folder is prepended to the environment PATH  before running `usdGenSchema`. 
This allows to run usdGenSchema with a specific python distribution that is different from the system's default one.
I also had to add USD_LIB to the path, so that the usd libs can be found. I'm not sure why the current code currently adds instead USD_BIN to the path. I first wanted to remove it and replace it with USD_LIB, but in case this is needed for HtoA I kept the existing one.

Fixes #874

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -74,6 +74,7 @@ vars.AddVariables(
     PathVariable('TBB_INCLUDE', 'Where to find TBB headers.', os.getenv('TBB_INCLUDE', None)),
     PathVariable('TBB_LIB', 'Where to find TBB libraries', os.getenv('TBB_LIB', None)),
     BoolVariable('TBB_STATIC', 'Whether we link against a static TBB library', False),
+    StringVariable('SCHEMA_PREPEND_PATH', 'Folder to prepend to the PATH environment to run usdGenSchema', None),
     # Google test dependency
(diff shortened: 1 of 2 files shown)


sebastienblor:
Actually, while it works on windows, it's probably not final on linux/mac. We need 2 things:
1) make it so that the proper python executable is ran
2) allow to load the python3 dso, which is needed by the usd python modules.

On windows, setting the PATH environment works for both, but not on linux/mac. We also need it in LD_LIBRARY_PATH

sebastienblor:
In the above commit, I extended the logic for it to work on linux and mac as well, plus some comments for clarity

sebastienblor:
Actually, when testing this on linux for MayaUSD builds, I realized that we're actually supposed to run `mayapy usdGenSchema` and this automatically sets the right environment. So I no longer need to set anything fancy in the environment for this to work. I'll close this PR and maybe I'll add a new one to override the usdGenSchema command

PR #881, opened 2021-09-14
Fixing testsuite issues using 21.02+ and supporting new and old light parameter naming conventions.

### LINKED ISSUE #772: Testsuite fails when using USD 21.05/21.02
**Describe the bug**
There are failures when running the testsuite using USD 21.05 or 21.02, I suspect these happen due to the changes in light token names, as most of the failures due to lighting differences

**To Reproduce**
1. Build/run the testsuite using USD 21.05.

**Expected behavior**
Testsuite runs clean.

### LINKED ISSUE #880: Scenes using pre-21.02 light parameters are not rendering correctly
**Describe the bug**
USD files exported from applications using an older version of USD won't work correctly in the procedural, due to the light parameter name changes in 21.02 and 21.05.

sirpalee:
- Supporting both new and old light parameter naming conventions, without using the UsdLux APIs.
- Changing tests 140 and 181, so we make sure the new parameters are preferred over the old ones.
- Fixing test 15 with 21.08.
- Adding support for animated color temperature.

Fixes #880
Fixes #772

**Additional context**
I benchmarked the new code vs the old code, using the worst possible case: when no parameters are set we are checking both the new and old parameter names. The results were: 1-3 seconds when translating 1 million UsdLux lights to Arnold primitives on a single thread. This means about 1-3 microseconds of overhead for each light, which seems acceptable.

In scenes with many lights (above 10k), we have significant other overheads than just 10ms+ introduced in this PR.

### DIFF
--- a/translator/reader/read_light.cpp
+++ b/translator/reader/read_light.cpp
@@ -21,6 +21,7 @@
 
 #include <pxr/usd/usd/prim.h>
 #include <pxr/usd/usdGeom/mesh.h>
+#include <pxr/usd/usdLux/blackbody.h>
 #include <pxr/usd/usdLux/diskLight.h>
 #include <pxr/usd/usdLux/distantLight.h>
 #include <pxr/usd/usdLux/domeLight.h>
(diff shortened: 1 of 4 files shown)


sebastienblor (approved):

translator/reader/read_light.cpp
```diff
@@ -36,63 +37,118 @@
...
+{
+    if (attr.HasAuthoredConnections()) {
+        return attr;
+    } else {
+        auto oldAttr = light.GetPrim().GetAttribute(oldName);
+        return oldAttr && oldAttr.HasAuthoredConnections() ? oldAttr : attr;
+    }
+}
+
+#if 1
```
jhodgson, line 97: We can remove the unused case here.

PR #883, opened 2021-09-17
Support render tags in the render delegate.

sirpalee:
- Adding support for render tags in the render delegate.

Fixes # 843


sirpalee:
I made this PR draft, as there are a few things I want to test, like to make sure the behavior matches other render delegates, and check what happens with point instancers.

sebastienblor (approved):

sirpalee:
Fixed an issue regarding purpose support for point instancers and reorganized the code.

PR #885, opened 2021-09-17
Only call AiBegin/AiEnd in the delegate if no arnold session is active

### LINKED ISSUE #884: Render delegate shouldn't call AiBegin/AiEnd if Arnold is already active
Currently, the render delegate always calls AiBegin in its constructor, and also calls AiEnd in its destructor.
This is causing crashes in plugins that expose it (e.g. MtoA) because this ends up conflicting with the active arnold session in the plugin.
There is no fully robust way to be fully sure of when the session should be closed or not, but the best solution is probably to only call AiEnd if AiBegin was called, and only do that if arnold is not active in the render delegate constructor

sebastienblor:
As it's being done in other parts of the code, we only call AiBegin/AiEnd if no Arnold session is active.
Note that this arnold API will be renamed as AiArnoldIsActive, in order to avoid confusions with universes that are something different. That's why the variable I'm adding is called isArnoldActive.

Fixes #884


sirpalee (approved):
LGTM

PR #889, opened 2021-09-20
Improving conversions between integer types.

### LINKED ISSUE #887: Int parameters are not converted to unsigned in shader parameters in the render delegate
Shader parameters of type unsigned int are represented by integer parameters in Houdini, and these are currently translated as integer primitives attributes in usd. 

However the render delegate is not converting these integer parameters since there is a type mismatch.

The best way to fix this is to tag the houdini parameters with their usd type using the sidefx::shader_parmtype tag since this will result in the correctly translated usd types.  But the Houdini material translation python scripts don't support the Sdf.ValueTypeNames.UInt type (see HTOA-1841) so there is an exception setting the scalar values.

sirpalee:
- Converting integer parameters to unsigned int.
- Converting unsigned int to int.
- Converting unsigned int to byte.

Fixes #887


jhodgson (approved):
Looks good to me

PR #901, opened 2021-10-28
Avoid calling AiRenderBegin when render is already running

### LINKED ISSUE #900: Avoid calling AiRenderBegin if render is already running
**Describe the bug**
Calling AiRenderBegin while a render is already running could potentially lead to a crash with Arnold 7, so we should avoid doing it.

sirpalee:
- Avoid calling AiRenderBegin when render is already running.

Fixes #900


jhodgson (approved):
LGTM

PR #907, opened 2021-11-04
Setting the returned version to default in the Sdr plugin

### LINKED ISSUE #906: Version is not set as default when returning SDR definitions
**Describe the bug**
When returning Sdr definitions, the version is not set as the default version, which makes shader lookups by name more complicated.

sirpalee:
- Setting the returned version to default in the Sdr plugin.

Fixes #906


jhodgson (approved):
LGTM

PR #909, opened 2021-11-04
Merging fix-7.0.0 to master

### LINKED ISSUE #900: Avoid calling AiRenderBegin if render is already running

### LINKED ISSUE #902: Render session is not passed AiDeviceAutoSelect
**Describe the bug**
The internal render session pointer is not passed to AiDeviceAutoSelect if targeting Arnold 7, which permits switching to the GPU backend correctly.

### LINKED ISSUE #905: arnold-usd int[] primvars
int[] primvars are not working with renderVar in solaris
we can see float and color ones no prob, but not int ones

[aov_int.zip](https://github.com/Autodesk/arnold-usd/files/7459248/aov_int.zip)

### LINKED ISSUE #906: Version is not set as default when returning SDR definitions

sirpalee:
- Fixing the use of AI_TYPE_INT AOVs.
- Removing a warning when setting blend_opacity.
- Setting the returned version to default in the Sdr plugin.
- Avoid calling AiRenderBegin when render is already running.
- Passing render session to AiDeviceAutoSelect.

Fixes #905
Fixes #906
Fixes #900
Fixes #902

### DIFF
--- a/ndr/discovery.cpp
+++ b/ndr/discovery.cpp
@@ -62,14 +62,14 @@ NdrNodeDiscoveryResultVec NdrArnoldDiscoveryPlugin::DiscoverNodes(const Context&
         TfToken filename("<built-in>");
         prim.GetMetadata(_tokens->filename, &filename);
         ret.emplace_back(
-            NdrIdentifier(TfStringPrintf("arnold:%s", shaderName.GetText())), // identifier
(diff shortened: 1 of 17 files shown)


jhodgson (approved):
LGTM

PR #920, opened 2021-11-12
Using GetVolumeShader if native rprim is volume

### LINKED ISSUE #915: Volume shader is not applied to the ArnoldVolume primitive in Hydra
**Describe the bug**
The Native Rprim hydra primitive currently only applies the surface shader. It should use the volume shader if it's a volume procedural/primitive.

sirpalee:
- Using GetVolumeShader if native rprim is volume.

Fixes #915


jhodgson (approved):
Looks good, fixes the ArnoldVolume volume shader assignment in Solaris

PR #926, opened 2021-11-15
Remap curves primvars

### LINKED ISSUE #854: Primvars is ignored when rendering basisCurves as bspline
**Describe the bug**
When rendering BasisCurves using basis bspline, the primvars is ignored. and we get messages like this

```
/Root/RootOffset/g_geometry/groom/my_nice_groom/my_nice_groomShape: varying user-data count wrong! (found 651014 expected 550858) -- disabling user-data 'look_color'

/Root/RootOffset/g_geometry/groom/my_nice_groom/my_nice_groomShape: varying user-data count wrong! (found 651014 expected 550858) -- disabling user-data 'look_tip'
```

**To Reproduce**
1. Extract the attached zip file
2. Edit the .ass file and update the path to the usdfile (currently its pointing to my local harddriver)
3. kick -l ../shaders -l ../procedurals -l ../plugins C:/qvisten_pipeline/primvarsBug/renderme.ass

**Expected behavior**
If the primvars is working fine when rendering the BasisCruves using linear basis, it should also work when using bspline basis

**Screenshots**
render trough the arnold procedural  basis bspline
image

Render trough the arnold procedural  basis linear
image

Rendered with Arnold Hydra in solaris
image

**Used Software Versions**
 - Arnold: 6.2.1.1
 - OS: windows 10
 - Maya 2022.1

**Additional context**
[primvarsBug.zip](https://github.com/Autodesk/arnold-usd/files/6945016/primvarsBug.zip)

### LINKED ISSUE #921: Remap curves primvars to avoid "wrong data count" errors
For curves, the procedural currently remaps the width in case of vertex or varying interpolation. But this should also be done for primvars that are translated as arnold user data.
The render delegate currently does this properly, it should be implemented in the procedural too

sebastienblor:
As it's being done in the render delegate, the procedural now remaps the curves primvars in certain interpolations and curve bases. This was already done for the curves width, so we had to extend it to other primvars.

The function `UsdArnoldPrimReader::ReadPrimvars` already had an optional argument `MeshOrientation` to eventually modify some primvars indexes for polymeshes. Here we needed a similar argument to remap the curves so instead of adding yet another argument, I've replaced the mesh orientation by a more generic `PrimvarsRemapper` that is specialized in both the mesh and the curves code to remap the primvars accordingly.

Fixes #921 
Fixes #854


sirpalee (approved):
LGTM

PR #930, opened 2021-11-17
Merging Fix 7.0.0 to master

### LINKED ISSUE #918: Rendervars do not show up in Houdini 19
**Describe the bug**
Rendervars don't show up in the viewport anymore in Houdini 19.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a rendervar, render product and render settings.
2. Set the rendervar to raw and RGBA with the name of rendervar1

### LINKED ISSUE #928: Support Additional transforms on PointInstancer
Currently transforms applied on top of a PointInstancer are not supported see attached example USD files

This works in the render delegate but not in the procedural

[instance_with_xform.usda.zip](https://github.com/Autodesk/arnold-usd/files/7545510/instance_with_xform.usda.zip)
l

sirpalee:
- Changing the default descriptor returned for AOVs.
- For motion blur, we were already just considering the instance xforms in the first and last keys. So I'm doing the same for the point instancer matrix as well.
- Adding test_0185 with the scene provided in the ticket, it shows that the instances are 1 unit away in the Y axis.

Fixes #918
Fixes #928


jhodgson (approved):
Looks good to me.

PR #941, opened 2021-11-30
Avoid linking against hdx for older USD versions

### LINKED ISSUE #937: Do not link against Hdx when it's not required
**Describe the bug**
At the moment the render delegate links against Hdx, even when the OpenGL blitting functions are not used. This is causing potential dso loading issues in Houdini 19 and husk.

sirpalee:
- Avoid linking against hdx for older USD versions.

Fixes #937

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -243,6 +243,7 @@ else:
 # Get USD Version
 header_info = get_usd_header_info(USD_INCLUDE) 
 env['USD_VERSION'] = header_info['USD_VERSION']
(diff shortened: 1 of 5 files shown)


sebastienblor (approved):

jhodgson (approved):
After this change in Houdini 19 on Linux:
* Husk is running correctly from the USD render ROP with the Arnold option chosen
 (HTOA-1874)
* The Arnold renderer option is available when loading old Houdini scenes in the Solaris viewport again

jhodgson:
And hdArnold.so is no longer linking with hdx as required.

PR #958, opened 2021-12-08
Support USD 21.11

sebastienblor:
**Changes proposed in this pull request**
In order to compile arnold-usd against USD 21.11, I've just had to do a few changes in the translator code regarding the removal of `UsdLuxLight` 
The code is still compatible with older versions of USD
Testsuite seems to be ok, so we don't need to update test scenes apparently

**Issues fixed in this pull request**
Fixes #957

PR #961, opened 2021-12-09
Fix build of usd_imaging with USD 21.11 #957

### LINKED ISSUE #957: Support USD 21.11
**Describe the bug**
USD 21.11 comes with incompatible changes, in particular regarding the UsdLux API.
We need to port it to the code, and still make it backwards compatible

sebastienblor:
This PR fixes the build of usd_imaging with USD 21.11.

Fixes #957

### DIFF
--- a/usd_imaging/usd_lux_light_filter_adapter.cpp
+++ b/usd_imaging/usd_lux_light_filter_adapter.cpp
@@ -13,7 +13,7 @@
 // limitations under the License.
 #include "usd_lux_light_filter_adapter.h"
 
-#include <pxr/usd/usdLux/light.h>
+#include <pxr/usd/usdLux/lightAPI.h>
 #include <pxr/usd/usdShade/material.h>
 
 #include <pxr/usdImaging/usdImaging/indexProxy.h>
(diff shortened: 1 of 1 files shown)


jhodgson (approved):
Looks good with the version switch

SimonCSmith:
Not sure if this is the right place to post this, but I grabbed the latest master code (to get this fix) as i'm building against 21.11 and I'm getting compiler errors as the translator library still uses the `UsdLuxLight` class in **read_light.cpp**.

I'm currently working around it by passing a `UsdPrim` in (rather than `UsdLuxLight`) and changing the callers of those 3 functions to use the prim they have passed in instead of their `UsdLuxXXX` light that they were previously.

(Side note: Compiling under VS2019 I had to change `ImagingArnoldDelegate::GetTaskRenderTags` to explicitally `return {}` else that fails to compile too).

Hope this is useful for someone here, and apologies if i've posted this in the wrong place!

sebastienblor:
Hi Simon, this is supposed to have already been addressed by #958 which should be in master.

SimonCSmith:
I'm not sure that fix has made its way into Master - however it does seem to be in the fix-7.0.0 branch comparing `read_light.cpp` between the 2 branches on github.

PR #995, opened 2022-01-06
Fix arnold node graph adapter for USD 21.08

### LINKED ISSUE #994: Background shaders networks are not traversed correctly
ArnoldNodeGraphAdapter::GetMaterialResource is not traversing the ArnoldNodeGraph correctly in Houdini 19, since the UsdShadeConnectableAPI instance is not detecting the primitive has a connectable API.

jhodgson:
The ConnectableAPIBehaviour is only available from USD 21.11 see https://github.com/PixarAnimationStudios/USD/blob/v21.11/pxr/usd/usdShade/plugInfo.json

The early out relies on this from apiSchemaBase.h (UsdAPISchemaBase):

/// \note When the bool-conversion operator is invoked on an applied API 
/// schema, it evaluates to true only if the application of the API schema has
/// been recorded on the prim via a call to the auto-generated Apply() method.

To work around this we can skip the early and if the connectable behaviour is not available we will still return an empty HdMaterialNetworkMap since there will be no outputs.

Fixes #994

### DIFF
--- a/usd_imaging/node_graph_adapter.cpp
+++ b/usd_imaging/node_graph_adapter.cpp
@@ -70,9 +70,6 @@ VtValue ArnoldNodeGraphAdapter::GetMaterialResource(
     // since this is simplified.
     HdMaterialNetworkMap materialNetworkMap;
     UsdShadeConnectableAPI connectableAPI(prim);
-    if (!connectableAPI) {
-        return VtValue{materialNetworkMap};
-    }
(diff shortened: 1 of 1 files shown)


sebastienblor (approved):
In theory this test isn't strictly needed since GetOutputs below should return an empty list if the schema isn't valid.
If we end up needing it, we can reintroduce it with a version ifdfef

PR #997, opened 2022-01-07
Support point instancer with a cache_id in the procedural

### LINKED ISSUE #996: Point instancer not working in the procedural with cache_id
**Describe the bug**
When using the procedural with a cache_id, point instancers will try to create a child usd procedural with a filename. They should instead create a procedural with a cache_id. This is preventing point instancers from working properly in Maya with a proxy shape

sebastienblor:
Ensuring that point instancers can work with the "cache" procedural that accesses the usd stage through a cache id in the UsdStageCache. Whenever the point instancer needs to create a child procedural to instantiate a given hierarchy, it might need to create a cache proc and pass the cache id, instead of always creating a "usd" procedural and relying on the filename.

Fixes #996

### DIFF
--- a/common/constant_strings.h
+++ b/common/constant_strings.h
@@ -170,6 +170,7 @@ ASTR(bottom);
 ASTR(box_filter);
 ASTR(bucket_scanning);
 ASTR(bucket_size);
+ASTR(cache_id);
 ASTR(camera);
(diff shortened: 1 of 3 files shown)


krishnancr (approved):
Looks good to me.

PR #1007, opened 2022-01-20
Use arnold instancer node in the procedural

### LINKED ISSUE #938: PointInstancer of 100K cubes hangs procedural
**Describe the bug**
The attached USD file cannot be rendered with the Arnold usd procedural. The render will hang, usually before the rendering starts.
[scatter-cube.zip](https://github.com/Autodesk/arnold-usd/files/7610020/scatter-cube.zip)

**To Reproduce**
1. Edit scatter_cubes.usda and update the reference to cube.usda
2. kick scene.ass, or in Maya, use an aiStandIn to load scatter_cubes.usda
3. Render

```
kick -v 6 -dp C:\Users\blairs\Downloads\_support\aa\scatter-cube\scene.ass
00:00:03   160MB         | initializing 10 nodes ...
00:00:06   269MB   
      |  [proc] /aiStandIn/aiStandInShape: loaded 100005 nodes (100003 objects, 0 shaders)
```

**Expected behavior**
The instances should render, like they do in usdview or if you kick the scatter_cubes.usda file.

**Screenshots**

**Used Software Versions**
 - Arnold: 7.0.0.1
 - USD: 
 - OS: Windows 10
 - Any 3rd-party app: Maya, Houdini

**Additional context**
https://answers.arnoldrenderer.com/questions/36820/render-not-starting-with-mtoa-usd-point-instancer.html

### LINKED ISSUE #949: Use arnold point_instancer in the procedural
When reading a usd `PointInstancer`  primitive, we're still creating several `ginstance` arnold nodes. 
But arnold now has a similar `point_instancer` node, that could be used transparently. The render delegate already relies on it, so we should make it consistent with the procedural.

### LINKED ISSUE #956: PointInstancer double transformations
**Describe the bug**
When translating a prim that contains an instancer with the prototypes as children, both the points and the prototypes are being translated resulting in double transformations.

**To Reproduce**
1. render the attached point_instance_offset.usd file with the usd procedural

**Expected behavior**
When translating an instancer I expect only the points to be moved (and the prototypes being scattered on those new translated points) the prototype offset from the point should be relative to the parent of the prototype rather than relative to world space.
For the attached scene the two instanced cubes (screen-left) should be touching the normal cube (screen right).

**Screenshots**
Current:
instace_offset_actual

Expected:
instace_offset_expected

**Used Software Versions**
 - Arnold: 7.0.0.1
 - USD: 0.21.08
 - Compiler: gcc 6.3.1
 - OS: CentOS 7

**Additional context**

[point_instance_offset.zip](https://github.com/Autodesk/arnold-usd/files/7676033/point_instance_offset.zip)

sebastienblor:
With this PR, we read usd PointInstancer primitives as arnold "instancer" nodes (instead of creating lots of ginstance nodes).
We now set the primvars similarly to what usd does. 
I'm adding test_0196 where the primvars are set on the parent instancer, that was already working in the render delegate. All the other tests are working correctly. 
I noticed the performance is better than in the previous implementation

Fixes #949 
Fixes #938 
Fixes #956


sebastienblor:
In the latest commit I'm also fixing the double transformation issues with point instancer reported in #956. 
I could have done it in a separate PR but it required this one to be merged first so I'm putting everything under the same hat.

There were 2 problems to be addressed, both happening when an instancer prototype is a nested usd procedural (which we need to do e.g. when the prototype is a xform in the usd scene) :

1) when this prototype has a xform, it was being applied twice. This happened because an instance inherit_xform only applies to the child procedural's matrix, not to its root primitive's matrix. So we need to exclude the prototype matrix in that use case.

2) these child usd procedurals have a root primitive (object_path set in the procedural) so that it only loads the expected prototype. But when we compute primitive's matrices in this procedural, we consider the ones from the root prim's ancestors, which shouldn't happen

jhodgson (approved):

PR #1011, opened 2022-01-28
 Fix shadow/light linking when some lights don't have any collection

### LINKED ISSUE #1010: Issue in light linking when some lights don't have any collection
**Describe the bug**
When some lights have light/shadow linking info and others don't, the later get excluded when they shouldn't

sebastienblor:
There was an issue with the existing light linking code. When lights weren't part of any collection they were supposed to be applied to all shapes, but because of a bracket mismatch in the code, this wasn't applied properly. Adding test_0200 to exercise this use case.

Fixes #1010


jhodgson (commented):
LGTM

jhodgson (approved):

PR #1017, opened 2022-02-01
Allow to specify a given RenderSettings prim when kicking a usd file

### LINKED ISSUE #1016: Control which RenderSettings prims is used when reading a usd file
The USD procedural converts USD RenderSettings primitives into Arnold options. However, there can only be a single arnold options whereas nothing prevents from having many RenderSettings in the USD stage.
We should add an argument when reading a usd file so that we can choose which render settings to use

sebastienblor:
We add a possible paramMap "render_settings" when loading a usd file. If set, the reader will only pick a RenderSettings prim with this given name. If not set, only the first RenderSettings prim will be considered (currently all would end up being translated to the same arnold options node)

Fixes #1016

### DIFF
--- a/common/constant_strings.h
+++ b/common/constant_strings.h
@@ -342,6 +342,7 @@ ASTR(region_max_y);
 ASTR(region_min_x);
 ASTR(region_min_y);
 ASTR(render_device);
+ASTR(render_settings);
 ASTR(repeat);
 ASTR(rotation);
 ASTR(roughness);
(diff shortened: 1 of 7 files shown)


sebastienblor:
In the second commit I added a test (203) that opens several times the same usd file with different params for `render_settings` .
The test verifies the translated arnold options node has the proper attribute values and camera. It also verifies that when no render settings are provided we exclusively read the first one found during the traverse, instead of reading all of them as before.

jhodgson (approved):
Looks good to me

PR #1018, opened 2022-02-02
Get proper instancer matrices when some instances are hidden #1005

### LINKED ISSUE #1005: InivisibleIds property not interpreted correctly on the PointInstancer
When instances are hidden through the invisibleIds[] property, the solaris viewport which is expected behaviour but the USD procedural is hiding the wrong instances
[test.usda.txt](https://github.com/Autodesk/arnold-usd/files/7896534/test.usda.txt)

Left : Procedural.  Right, Delegate.
image

sebastienblor:
When we get the point instancer matrices for each time key, we need to set the argument `IgnoreMask` so that the returned data doesn't skip invisible instances. This way we can set the whole data properly along with visibilties in the arnold scene in a non-destructive way
Added test 202 with the repro scene from the ticket
Note that the image is slightly different than what was posted, the file is probably not exactly the same one. I verified in Solaris that we do get the same result

Fixes #1005


jhodgson (approved):
Looks good to me.

PR #1020, opened 2022-02-03
Support light linking with nested collections

### LINKED ISSUE #1019: Support nested collections with light-linking
As per the UsdCollectionAPI  SDK, it is possible for a collection includes list to reference another collection with an "instanceName".
In this case there can be a single "Collection" primitive in the stage, having multiple namespaces each representing a single collection "instance".

We need to take this into account in the procedural

sebastienblor:
When the light linking collection points to another collection, we now propagate the information properly. USD uses a special format for that , which now supported (and commented).
As we need to remove test 190, which would need too much disk space to be render, I'm replacing it with a test for the current ticket
Fixes #1019


jhodgson (approved):
Looks good.

PR #1025, opened 2022-02-14
Fix random result with instancer and multiple referenced files

### LINKED ISSUE #1021: USD Instanceable Bug
**Describe the bug**
When creating a basic scene with 2 cubes, 2 cones, and 2 torus' that all have the Instanceable Metadata set to True, every time we render with Arnold inside of Maya, the instanceable prototype groups that get generated are being randomly assigned to each instance group. Every time you render, the instances swap places. For instance, the first render may be (Cube, Cube, Cone, Cone, Torus, Torus) and the next may be (Torus, Torus, Cube, Cube, Cone, Cone). Loading this same USD into Houdini and rendering with Arnold seems to render everything fine consistently.

**To Reproduce**
1. Download 
[cube_cone_torus.zip](https://github.com/Autodesk/arnold-usd/files/8037126/cube_cone_torus.zip)
2. Open Maya with Arnold and mayaUSD plugins active.
3. Load the zipped cube_cone_torus USD as a stage.
4. Add some sort of light so you can see the objects.
5. Render the scene and note the placement of the objects.
6. Render again, and note the placement of the objects.
7. Render a few more times to be sure that the objects are swapping places.

**Expected behavior**
I expect the Instanceable objects to render in the same spot every time.

**Screenshots**
render1
render2

**Used Software Versions**
 - Arnold: 6.2.1.1
 - MtoA: 4.2.4
 - mayaUSD: v0.13.0
 - OS: Windows 10
 - Houdini (for comparison testing): Houdini v19.0.383

**Additional context**
None.

### LINKED ISSUE #1022: Point Instancer gives random results with procedural
**Describe the bug**

**To Reproduce**
1. Extract the attached zip. It includes an ASS file with a usd procedural
2. Kick the ass file
[00029310_test_setdress.zip](https://github.com/Autodesk/arnold-usd/files/8043811/00029310_test_setdress.zip)

**Expected behavior**
All the instances should be the same in each render.

**Screenshots**
Three kick renders of the same ass file
point_instancer_random_results

**Used Software Versions**
 - Arnold: 7.x
 - USD: 
 - Compiler: 
 - OS: Windows 10
 - Any 3rd-party app: 

**Additional context**
USD was set up in Solaris. Renders ok in Solaris. Looks ok in the Maya viewport too.

sebastienblor:
For instanceable primitives that are referencing an external usd file, need to create a nested USD procedural with the same filename, but with an object path set to the name of the prototype primitive. But in this case, the prototype primitive is created on the fly by USD to represent the file reference, and can be called `/__Prototype1`, or `/__Prototype2`, `/__Prototype3`, depending on the order in which usd references are evaluated. This order can change randomly when we load the same usd file over and over. Because of that, we cannot set such a name in the nested procedural "object path" if we want to have predictable results.
What I'm doing to address this limitation, is to store a list of these instanceable file references, to map the computed primitive name with its filename. Then, the nested usd procedural will point directly to this file. 

Adding test_0204 for this use case

Fixes #1021 
Fixes #1022


krishnancr (approved):
Looks good to me.

PR #1026, opened 2022-02-15
Ensure the lights normalize attribute is authored in usd files

### LINKED ISSUE #1024: Resaving a distant_light from .ass to usd does not preserve the normalize parameter
Resaving an ass file with `kick test.ass -resave test.usda` does not preserver the `normalize` parameter.

The .ass file has `distant_light.normalize on`, but the resaved usd file has no entry `bool inputs:normalize = 1`.

[test.ass.txt](https://github.com/Autodesk/arnold-usd/files/8060084/test.ass.txt)

The arnold default is on but the usd default is off so it may be we are checking the default value of the arnold parameter instead of the usd parameter on writing to usd.

sebastienblor:
When we write distant or skydome lights, we were not authoring the normalize attribute. This is causing issues, since the default is different between usd and arnold. With this PR, we're now ensuring we're always dump it for these "non-boundable" lights

Fixes #1024

**Additional context**

### DIFF
--- a/common/constant_strings.h
+++ b/common/constant_strings.h
@@ -66,6 +66,7 @@ ASTR2(houdiniFps, "houdini:fps");
 ASTR2(hydraPrimId, "hydra_primId");
 ASTR2(inputs_code, "inputs:code");
 ASTR2(primvars_arnold_subdiv_type, "primvars:arnold:subdiv_type");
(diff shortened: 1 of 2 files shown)


jhodgson (approved):
Looks good, and this is fixing the arnold-usd tests in HtoA using distant lights.

PR #1031, opened 2022-02-16
Fix for element connection in arnold node graph

jhodgson:
Currently we have a warning in the render delegate code for connecting array elements

```
render_delegate/node_graph.cpp:830:54: warning: multi-character character constant [-Wmultichar]
            size_t elemPos = outputAttr.find_last_of(':i');
                                                     ^
render_delegate/node_graph.cpp:830:54: warning: implicit conversion from 'int' to 'std::basic_string<char>::value_type' (aka 'char') changes value from 14953 to 105 [-Wconstant-conversion]
            size_t elemPos = outputAttr.find_last_of(':i');
```
This doesn't amount to a bug in the render delegate, but is in fact only searching for the last :, not the last :i since we pass a char not a string. If we has used a string it would pick the last of : OR i which is not what we want either.

We can fix by using rfind("i0") and updating the indices.

Note we only get different behaviour for strings such as input[3].i which we would represent as input:i0:i, but we would never have such connections, only input[3].x or similar (i.e. components x,yz,r,g,b,a).

But this fix will make the code tighter and remove the compile warning.


jhodgson:
Test_0293 is still passing in HtoA.

Issue #1001, opened 2022-01-14
Changing signature for AiMsg* calls 


krishnancr:
AiMsg* API calls used to require a render session. This is being changed to require a universe.

Issue #998, opened 2022-01-11
UsdPreviewSurface opacity translated incorrectly in the procedural


sebastienblor:
When the opacity of a UsdPreviewSurface shader is defined in a usd procedural scene, it won't get translated correctly because arnold expects a color while usd defines it as a scalar.

Issue #987, opened 2021-12-28
Light/shadow links consider ignoreRoots as true by default


sebastienblor:
**Describe the bug**
In the `CollectionAPI` for light links and shadow links, the procedural is currently assuming that when `ignoreRoots` is disabled, it is true by default.
That's wrong based on the collection api schema, it should default to false. This is leading to wrong results (as opposed to the render delegate which is correct)

PR #1032, opened 2022-02-16
Update fix-7.0.0

sebastienblor:
Updating fix-7.0.0 with the following changes :

Ensure preview surface opacity is set properly #998
#986 Read usd builtin normals in meshes
#977 Support Curves UVs
#951 Extend the skinning baking interval to surrounding integer frames
#944 Instance primvars are ignored in the procedural
#939 Add writer API to skip material bindings
#970 Support connections to array elements when translating hydra material networks
#987 Lights and shadow links default ignoreRoots to false
#873 Add support for atmosphere and background shaders in the delegate
#983 Fix compilation issues with Visual Studio 2019
Fixes on NodeGraph schema to enable it with USD 21.08 #995
Support nested instance primvars in the render delegate #482
Changing signature for AiMsg calls #1001
Replacing more deprecated arnold APIs #1003
Use arnold instancer node in the procedural #949
Support point instancer with a cache id #996


jhodgson:
The arnold node graph and https://github.com/Autodesk/arnold-usd/issues/873 (Add support for atmosphere and background shaders in the delegate) are new features, so do we want them in the fix-7.0.0 branch?

PR #1035, opened 2022-02-18
Propagate thread count to nested procedurals for instancing 

### LINKED ISSUE #1034: Nested usd procedurals for instancing aren't passing the proper thread count
**Describe the bug**
When instanced primitives require to create a nested usd procedural, we're not passing the current amount of threads requested by the user. This is preventing e.g. from properly using a single thread for usd traversal in this case

sebastienblor:
When we need to create a nested usd procedural for instancing purpose, we now also propagate the thread count to this procedural, so that we can properly debug scenes with a single thread

Fixes #1034


krishnancr (approved):
Looks good to me !

Issue #1037, opened 2022-02-22
Pcp missing from shared libraries list


sebastienblor:
A recent commit to the translators started to rely on the pcp libs from USD.
When building against shared usd libs, we need to add pcp to the link list, both with Scons and Cmake

PR #1038, opened 2022-02-22
Add pcp to the list of shared USD libs #1037

sebastienblor:
This PR adds the usd lib `pcp` to the list of shared libraries the translator & procedural needs to link against.
We get build errors otherwise, when we build the procedural against shared usd libs

Fixes #1307


krishnancr (approved):
Looks good to me !

PR #1039, opened 2022-02-22
Use velocity/acceleration primvars for mesh/curves in the render delegate

### LINKED ISSUE #911: Use the velocity/acceleration primvar to extrapolate curve positions when shutter range is non-zero
We should use the velocity/acceleration primvars to extrapolate curve positions.

### LINKED ISSUE #912: Use the velocity/acceleration primvar to extrapolate polymesh positions and normals when shutter range is non-zero
We should use the velocity/acceleration primvars to extrapolate polymesh positions and normals.

sebastienblor:
The render delegate already had some code to support velocity/accelerations when computing the positions (in `_ExtrapolatePositions` ), but a few things were missing :
- meshes and basis curves need to provide the primvars list to `HdArnoldSetPositionFromPrimvar` so that they can be used in `_ExtrapolatePositions`
- We were only considering velocities if positions were not animated. This isn't consistent with what the procedural is doing, and wuold still leads to issues. If the positions are animated and have varying topologies, then we wouldn't read the velocities, and this would still lead to errors. Instead, as soon as velocities/accelerations are provided, we should always consider them  instead of reading positions at different times.
- We were not necessarily considering the positions at the "current" frame, as we were just reading the first time sample. In case of varying topologies, this could lead to reading positions from the previous frame, where the amount of elements doesn't match the amount of velocities. So we need to find a time sample that can be used with velocities/accelerations

I verified that the test scene from test_0157 that exercises this with meshes & varying topologies. This change should also apply to curves, but I don't have a test scene yet. So it will need to be verified

Fixes #912
Fixes #911 

**Additional context**

### DIFF
--- a/render_delegate/basis_curves.cpp
+++ b/render_delegate/basis_curves.cpp
@@ -73,7 +73,7 @@ void HdArnoldBasisCurves::Sync(
                                (*dirtyBits & HdChangeTracker::DirtyPrimvar);
     if (_primvars.count(HdTokens->points) == 0 && HdChangeTracker::IsPrimvarDirty(*dirtyBits, id, HdTokens->points)) {
         param.Interrupt();
-        HdArnoldSetPositionFromPrimvar(GetArnoldNode(), id, sceneDelegate, str::points, param());
+        HdArnoldSetPositionFromPrimvar(GetArnoldNode(), id, sceneDelegate, str::points, param(), GetDeformKeys(), &_primvars);
     }
 
     if (HdChangeTracker::IsTopologyDirty(*dirtyBits, id)) {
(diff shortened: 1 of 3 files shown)


jhodgson:
Working well for velocity blur on mesh in our test scene, although it seems necessary to restart the render when changing LOP selection.

For changing topology of mesh I'm seeing an error

00:00:07  2019MB WARNING |  [polymesh] /var_trans/mesh_0: discarded 2 duplicate deformation keys
00:00:07  2019MB ERROR   |  [polymesh] /var_trans/mesh_0: out-of-range vertex index

when no velocity is defined.

sebastienblor:
in the last commit, I'm raising an error for varying topologies without velocities ( for some reason these errors are not appearing in the log window for me, but that's another story). And I'm setting a single static key, so that we see at least the static geometries if abort on error is turned off

jhodgson:
Yes theses case seem good now.

LGTM

jhodgson (approved):
LGTM

PR #1040, opened 2022-02-23
Check python 2.7 and 3.7 folders for HtoA libraries

### LINKED ISSUE #955: Loading the HtoA openvdb conversion functions have hardcoded paths for the Python 2 build of Houdini
**Describe the bug**
When loading the `_htoa_pygeo` dso in the hydra volume primitive, our search paths are hardcoded to the Python 2 version of Houdini.

```
            const auto dsoPath = path + ARCH_PATH_SEP + "python2.7libs" + ARCH_PATH_SEP + "_htoa_pygeo" +
//. HTOA sets this library's extension .so on MacOS.
#ifdef ARCH_OS_WINDOWS
                                 ".dll"
#else
                                 ".so"
#endif
                ;
```

jhodgson:
Look for the _htoa_pygeo.so in either of the two possible python directories.

Fixes #955

### DIFF
--- a/render_delegate/volume.cpp
+++ b/render_delegate/volume.cpp
@@ -118,7 +118,7 @@ struct HtoAFnSet {
             if (path == "&") {
                 return false;
             }
-            const auto dsoPath = path + ARCH_PATH_SEP + "python2.7libs" + ARCH_PATH_SEP + "_htoa_pygeo" +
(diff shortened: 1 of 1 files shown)


sebastienblor (approved):

sebastienblor (approved):
Approving for now, but we really need to find a better way to support these HtoA specific stuff. It shouldn't be in the render delegate code

PR #1041, opened 2022-02-23
Setup color manager on the options node from the render settings LOP

### LINKED ISSUE #946: Setup color manager on the options node from the render settings parameters
In Solaris we need to set up the color manager options for the render and narrow color space and chromaticities on the render settings LOP, and pass them through to the color manager set on the options node via the render delegate.

jhodgson:
Set the render/narrow color space on the options node from the render settings LOP

Fixes #946


jhodgson:
We need to verify if the color space is set before the render starts, or can be set afterwards.

jhodgson:
We should also check for the config string if set on the render settings. 

We can either use the default color manager and set the config explicity, or we can always use a new color manager but default to the default colorspace OCIO file shipped with the core.

This will replace the getOrCreate behaviour.

jhodgson:
Moving to a new sprint.

PR #1044, opened 2022-02-23
Add version checks for older versions of Arnold

### LINKED ISSUE #1043: Cannot build against Arnold 6.2
Some new APIs are being used in the repository without checking the arnold version. This currently prevents from building arnold-usd with Arnold 6.2.x

sebastienblor:
Adding version checks, so that we only call `AiArnoldIsActive` for Arnold 7.x versions

Fixes #1043


jhodgson (approved):
LGTM

PR #1046, opened 2022-02-24
Fix crashes with invalid indexed primvars

### LINKED ISSUE #1045: Crashes with invalid indexed primvars and left-handed orientations
When a mesh has left-handed orientations, we need to remap the indexed primvars. But if the primvars has invalid data (ii.e its size doesn't match the mesh topology) then this can lead to crashes

sebastienblor:
This PR verifies that we're not setting out-of-bounds vector elements in case primvars have invalid indices. 
I'm also verifying we're not trying to declare a user data for an attribute that already exists to prevent warnings that I've seen while debugging this.
I'm not managing to reproduce the crash with a simpler scene than the one provided by the customer, so I couldn't add a test  

Fixes #1045

### DIFF
--- a/translator/reader/prim_reader.cpp
+++ b/translator/reader/prim_reader.cpp
@@ -598,8 +598,10 @@ void UsdArnoldPrimReader::ReadPrimvars(
 
         // Declare the user data
         AtString nameStr(name.GetText());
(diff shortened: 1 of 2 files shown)


jhodgson (approved):
Looks good.

PR #1050, opened 2022-02-25
Fix instanceable primitives without prototype

### LINKED ISSUE #1049: Instanceable primitives without prototype aren't rendered
While improving the support for instanceable primitives in the procedural, we introduced a regression when these instances don't have any prototype. They should still be rendered instead of skipped

sebastienblor:
When instanceable primitives didn't have any prototype, we were skipping all the primitive export. Now we still try to render as any regular primitive.
Adding test_0207 with the scene provided by the user.

While running the testsuite, I found an issue with the test scene for 157, just a `,` to remove. Not sure when this was introduced, but I fixed it in this PR

Fixes #1049


jhodgson (approved):
I see, we carry on reading the prim if no proto is found now.

PR #1053, opened 2022-03-03
Default mesh smoothing to true in the delegate

### LINKED ISSUE #24: Invert smoothing default

sebastienblor:
Just like it's done in the procedural, we're now setting the polymesh smoothing to be enabled by default in the delegate.
We were already doing similar in `HdArnoldMesh` for the `subdiv_iterations` that needed to be set to 0 by default.

Note that the `HdArnoldRprim`  constructor is different for usd versions older than 21.02. We had ifdefs for it, but they were containing the whole contructor function. I'm changing them to just change the function signature, which is enough and avoids code duplication on ech ifdef case.

Fixes #24


jhodgson (approved):
Works as required

PR #1055, opened 2022-03-03
Control scene_format_load thread count through an env variable

### LINKED ISSUE #1054: Control threads through an env variable for kick
It's currently not possible to pass custom arguments to scene_format_load, as when we kick a usd file.
It's therefore not possible to control the amount of threads used by the arnold-usd reader, as opposed to using a procedural 
that has such a parameter.
Until this is implemented in Arnold, we should allow to control this through an environment variable. When set, it would define the 
reader's thread count in scene_format_load

sebastienblor:
In `scene_format_load`, we check if an environment variable is set, and in that case we use this value to control the amount of threads used by the reader. This allows to control this thread count when kicking a usd file

Fixes #1054


procedural/main.cpp
```diff
@@ -286,6 +287,15 @@ scene_load
...
     float frame = AiNodeGetFlt(AiUniverseGetOptions(universe), "frame");
     int threadCount = 0;
+
+    // It is currently not possible to pass any custom arguments when we kick a usd file.
+    // In order to support any amount of threads in that use case, we're checking if an 
+    // environment variable is set, and use that value in that case    
+    const static std::string envThreads = "ARNOLD_USD_READER_THREADS";
+    if (ArchHasEnv(envThreads)) {
+        std::string envThreadsVal = ArchGetEnv(envThreads);
+        threadCount = std::max(0, std::atoi(envThreadsVal.c_str()));
```
jhodgson, line 297: Could use std::stoi(envThreadsVal) here instead of atoi

jhodgson (approved):

jhodgson (approved):
Looks good to me

PR #1064, opened 2022-03-10
Set the defaultPrim in the usd writer

### LINKED ISSUE #1063: Set defaultPrim when authoring usd files
The arnold-usd writer isn't setting any defaultPrim.
We should set one by default, so that these files can be referenced properly

sebastienblor:
It's been requesting several times that we should set a defaultPrim when authoring usd files, as this can be required for file referencing. This PR sets one value for that. If an explicit value is provided we use it, otherwise if a scope is set we use it as the defaultPrim. If none of the params is set, then we use the top parent primitive, created by `UsdArnoldWriter::CreateHierarchy`

Fixes #1063

### DIFF
--- a/common/constant_strings.h
+++ b/common/constant_strings.h
@@ -194,6 +194,7 @@ ASTR(crease_idxs);
 ASTR(crease_sharpness);
 ASTR(curves);
 ASTR(cylinder_light);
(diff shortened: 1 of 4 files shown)


jhodgson (approved):
Looks good

PR #1068, opened 2022-03-11
Write shading trees under their material's scope

### LINKED ISSUE #1067: Write shaders under their material's scope
The USD norm is to author shaders under the hierarchy of the materials.
This is a bit complicated to do when converting from arnold to usd, since materials don't exist in arnold. It means that we might have to duplicate shading trees where the same surface shader can be associated to different displacement shaders

sebastienblor:
In this PR we're now authoring shaders under their respective material's scope. For that we've needed to do the following changes :
- we split the loop through arnold nodes to be written in 2. First for all nodes except shaders, and a second one with only shaders
- in `processMaterialBinding` we create a material based on the surface (and eventually displacement) shader. And we then set the material's path as the new writer scope while we trigger the export of surface/displacement shaders. This way the shaders will be written under the material's hierarchy, and it will propagate to the whole shading tree
- when we check if a node was already written (to avoid doing it multiple times), we ensure the scope is taken into account. Therefore the same shader can now be authored multiple times, once for every different material it's assigned to
- in the second loop to write all shaders, we only consider those that weren't exported in the previous step, i.e. which aren't part of any shading tree assigned to a geometry. Only those will be authored with a regular naming (not under any material)
- Fixed an issue when testing if a shape was assigned to arnold's default shader. The previous code would have failed when a scope was being used
- Fixed test_0138 , as the shader names are now different
- Added test_0208 that exercises this

Fixes #1067

### DIFF
--- a/testsuite/test_0138/data/test.cpp
+++ b/testsuite/test_0138/data/test.cpp
@@ -27,7 +27,7 @@ int main(int argc, char **argv)
         AiMsgError("/beautiful/scope/my/cone/name doesn't exist");
         success = false;
     }
-    if (!AiNodeLookUpByName("/beautiful/scope/lambert1")) {
+    if (!AiNodeLookUpByName("/beautiful/scope/lambert1/lambert1")) {
         AiMsgError("/beautiful/scope/lambert1 doesn't exist");
         success = false;
     }
--- a/testsuite/test_0208/README
+++ b/testsuite/test_0208/README
@@ -1,0 +1,6 @@
+Write shaders under materials
+See #1067
+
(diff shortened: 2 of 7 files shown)


jhodgson:
Looks good as is but a couple of comments, perhaps for another issue.

1. We should look at collecting and naming the materials using the material_{surface,displacement,volume} metadata when it's available
2. Hanging shaders that aren't part of a material could still be collected under a Material primitive, named after the terminal shader.

jhodgson (approved):
Looks good.

sebastienblor:
Created #1069 for the user data

PR #1108, opened 2022-04-01
Add aov_shaders after primvar AOVs.

### LINKED ISSUE #1107: Add aov_shaders after primvar AOVs.
In the procedural we need to make sure we add the global aov_shaders after the primvar AOVs, currently they may overwrite them.  Let's also change the ordering to match this in the render delegate.

jhodgson:
Fixes #1107

In the procedural we need to make sure we add the global aov_shaders after the primvar AOVs, currently they may overwrite them. Let's also change the ordering to match this in the render delegate.


sebastienblor (approved):
LGTM

PR #1110, opened 2022-04-01
Remove deprecated API usage and compiler warnings in the procedural

### LINKED ISSUE #1109: Remove deprecated API use in procedural

jhodgson:
Fixes #1109

### DIFF
--- a/translator/reader/reader.cpp
+++ b/translator/reader/reader.cpp
@@ -323,7 +323,7 @@ void UsdArnoldReader::ReadStage(UsdStageRefPtr stage, const std::string &path)
             txt += " for procedural ";
             txt += AiNodeGetName(_procParent);
         }
-        AiMsgWarning(txt.c_str());
+        AiMsgWarning("%s", txt.c_str());
     }
     // If this is read through a procedural, we don't want to read
     // options, drivers, filters, etc...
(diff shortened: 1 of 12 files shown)


translator/reader/reader.cpp
```diff
@@ -966,42 +947,30 @@ void UsdArnoldReaderThreadContext::SetReader(UsdArnoldReader *r)
...
+    AtNode *node = AiNode(_reader->GetUniverse(), AtString(type), AtString(name), _reader->GetProceduralParent());
     // All shape nodes should have an id parameter if we're coming from a parent procedural
     if (_reader->GetProceduralParent() && AiNodeEntryGetType(AiNodeGetNodeEntry(node)) == AI_NODE_SHAPE) {
         AiNodeSetUInt(node, str::id, _reader->GetId());
     }
-    
-    if (_createNodeLock)
-        AiCritSecEnter(&_createNodeLock);
+
+    std::lock_guard<AtMutex> guard(_createNodeLock);
```
sebastienblor, line 971: We're just missing a if(dispatcher) to lock `_createNodeLock` , as for the 2 other locks. This way we won't be locking in the other modes.

sebastienblor (changes requested):
Overall it's looking great ! I just added a comment about `_createNodeLock` being locked even when there's no dispatcher. The other 2 locks are checking if dispatcher is null or not, it would be better to do the same with `_createNodeLock` too.

Issue #1042, opened 2022-02-23
Read render and narrow color space settings in the reader


jhodgson:
We'd like to define render and narrow color spaces on the render settings LOP, and have them converted to color space parameters on the options node when we read usd files or load a usd procedural.

Example usd file snippet

```    
    def RenderSettings "rendersettings2"
    {
        custom int arnold:global:AA_seed
        int arnold:global:AA_seed.timeSamples = {
            1042: 1042,
        }
        custom string arnold:global:color_space_config = "/dev/OpenColorIO-Configs/aces_1.0.3/config.ocio"
        custom string arnold:global:color_space_linear = "scene_linear"
        custom string arnold:global:color_space_narrow = "texture_paint"
        uniform token aspectRatioConformPolicy = "expandAperture"
        rel camera = </cameras/CAM>
        uniform float4 dataWindowNDC = (0, 0, 1, 1)
        uniform token[] includedPurposes = ["default"]
        uniform bool instantaneousShutter = 0
        uniform token[] materialBindingPurposes = ["full", "allPurpose"]
        uniform float pixelAspectRatio = 1
        rel products
        uniform int2 resolution = (2048, 1080)
    } 
```

Note the `color_space_config` is also set here so the color spaces make sense, and we should also consider supporting the `linear_chromaticities`.

PR #1111, opened 2022-04-01
Set color manager in reader (#1042)

jhodgson:


sebastienblor (approved):
Looking good.
We really need a test though, at least for the use case without the env variable being set. We can do it in a separate PR eventually, but we should do it asap  before we forget about it.

PR #1113, opened 2022-04-02
Build fix for procedural

### LINKED ISSUE #1112: Build issue on OSX

jhodgson:
Fixes #1112 

std::vector<bool> does not have .data() and is represented internally as a bitfield so let's just use C arrays.


sebastienblor (approved):
looks good

PR #1123, opened 2022-04-08
Fix build issue in procedural against USD 21.11

sebastienblor:
Recent commits are not building in USD 21.11. It's because one of the versions of `UsdGeomImageable::ComputeVisibility` has been removed. I'm computing it manually now, with the simple rule that visibility can either be `inherited` or `invisible`, and that a primitive is hidden if any of its ancestors is imageable and has visiblity set to `invisible`


compso (approved):
LGTM

PR #1126, opened 2022-04-11
Fix builds for older versions of USD

sebastienblor:
More fixes on the mesh light adapter, but for older versions of USD (20.11, 21.02). Some APIs were different back at the time.


compso (approved):
LGTM

PR #1127, opened 2022-04-13
Support instanceable prims with variant & transform overrides

### LINKED ISSUE #1122: Support instanceable prim with variant & transform overrides
An instanceable primitive can add overrides to the original primitive data, like changing the variant, etc...
This won't be supported in the procedural at the moment, since we'll create a nested procedural with an object_path, and these overrides will be lost.
We'd need to find a solution for variant overrides. Usd will create a different "prototype" prim when the variants are different, we could add an "over" statement in that case to modify the current variants.

Transformations in the instanceable prims are also not being taken into account properly when the prototype also has transforms.

sebastienblor:
When an instanceable primitive modifies some variants, we are now adding an "overrides" string to the nested procedural, so that it loads the prototype and then applies the desired changes.
This relies on the fact that USD will create different prototype prims for each instanced variant in that case ( e.g. __Prototype1, __Prototype2, etc...). So we can associate each of them with a set of variants, and then eventually add them to the nested procedural if they're actually different from the actual prototype

Adding test 221 with references to other prims. Also test 222 with references to external usd files

Fixes #1122


sebastienblor:
Note that test_0218 had to be updated as the new fixes revealed that the previous reference image was incorrect

jhodgson (approved):

PR #1134, opened 2022-04-26
Set outputs layer names only if required #1133

### LINKED ISSUE #1133: Only set output layer names if required
Currently, when we read a usd file with a render settings prim, as soon as several AOVs are found for the same driver, we give them a "layer name" (last token in each output string).
But this is actually required only if there are several AOVs of the same name in the same file.
And this is currently revealing an issue in cryptomatte : as soon as this layer name is present in the outputs cryptomatte fails to parse it and modify the outputs as needed, as it can be seen [here](https://github.com/Psyop/CryptomatteArnold/blob/master/cryptomatte/cryptomatte.h#L985).

In order to workaround this issue in cryptomatte, we could avoid setting a layer name if not explicitely required

sebastienblor:
Instead of always writing the layer name to the outputs string when we read a renderSettings prim, we only set it if there are several AOVs with the same name in the same driver

Fixes #1133


jhodgson (approved):

jhodgson:
We could also add HALF to the option string if the format is half, half2, half3, half4, but this could be a separate issue.

PR #1137, opened 2022-04-28
Fix uninitialized UV coords with leftHanded meshes

### LINKED ISSUE #1136: Out of range UVs with leftHanded meshes
We're facing random errors about out-of-range UV coordinates for meshes that are leftHanded (as with the scene attached).
Sometimes we need to change attributes interactively (e.g. subdiv parameters) for issues to show up.

The issue only happens with versions of USD higher than 21.05, and seems to be related to the flag
`USD_HAS_SAMPLE_INDEXED_PRIMVAR` which makes us read indexed primvars in a different way. The re-indexing done for leftHanded doesn't seem to be working well here

sebastienblor:
For recent versions of USD where we use indexed primvars differently, there was a typo in the code that reorders primvars indices for left-handed meshes.
When checking for out-of-bound indices in the primvar index arrays (in this case for `uvidxs` ), we were not comparing the indices correctly and skipping the values from the last polygon (there was a `>=` instead of a `>`).
Because of that, we were leaving the last few indices uninitialized, and that would randomly cause errors.

In the same vein, I'm also ensuring that we're not leaving uninitialized values in `nsides` when faces have no vertices, even though this is not what caused the original issue.

Fixes #1136

### DIFF
--- a/render_delegate/mesh.cpp
+++ b/render_delegate/mesh.cpp
@@ -228,6 +228,7 @@ void HdArnoldMesh::Sync(
             for (auto i = decltype(numFaces){0}; i < numFaces; ++i) {
                 const auto vertexCount = _vertexCounts[i];
                 if (Ai_unlikely(_vertexCounts[i] <= 0)) {
+                    nsides[i] = 0;
                     continue;
                 }
                 nsides[i] = vertexCount;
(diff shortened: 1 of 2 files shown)


jhodgson (approved):
LGTM

PR #1138, opened 2022-04-29
Support per-instance velocity blur in the delegate

### LINKED ISSUE #1092: Support instances with velocities and varying instance counts in the delegate
A Point instancer can have varying amount of instances, and velocity attributes to define their motion. We need to support this use case in the render delegate (see attached test scenes from #1090 , there is one meant for instances with this use case)

sebastienblor:
In the render delegate, when an instancer has a velocity primvar, we now use the per-instance velocities instead of checking the variations of the instance transforms. We also do the same if the accelerations primvar is present.
This is now matching what the procedural does when it calls `UsdGeomPointInstancer::ComputeInstanceTransformsAtTimes`.

Fixes #1092

### DIFF
--- a/render_delegate/instancer.cpp
+++ b/render_delegate/instancer.cpp
@@ -158,22 +158,45 @@ void HdArnoldInstancer::CalculateInstanceMatrices(HdArnoldRenderDelegate* render
         return;
     }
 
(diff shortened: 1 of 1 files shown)


jhodgson (approved):
LGTM

PR #1159, opened 2022-05-13
Improve and generalize Hydra dependency graph #1158

### LINKED ISSUE #1158: Improve Hydra dependency graph
**Describe the bug**
The render delegate needs to track the dependencies between translators, in order to properly update nodes that are dependent on others. This is currently used in particular for materials & node graphs, but could be extended to other use cases (e.g. geometry lights depending on the source mesh).
But this isn't currently tracked properly, as the data is stored in different parts of the code (some references are stored in the render delegate, whereas others are in each mesh nodeGraphTracker).
This causes refresh issues, or crashes, when tweaking shading trees.

sebastienblor:
This PR does quite a few changes on the previous "tracking" of node graphs. The previous system was a dependency graph, but hardcoded for specific use cases, and not fully reliable. It's important the the list of references & back-references is stored in a centralized place, so that every time something changes, all the lists are properly updated.  In this case, the "node graph tracker" which was part of the meshes, was not updated when the render delegate changed its dependencies, and we could end up with crashes or refresh issues when shading nodes were deleted / reconnected / etc...

Here's a list of what this PR changes :
- "node graph"  is now generalized to "dependencies", as we just really want to track the dependencies between some prims and others, for proper refresh. Both use case for meshes and lights are now unified
- node_graph_tracker is removed. Everything is now centralized in the render delegate
- did some cleanup on the functions to track new dependencies, remove existing ones, etc... 
- ensure the light shaders are only updated when the light is "dirty"
- removed the skydome attribute "shader" that we don't need to support for now
- whe na Sprim is created, we need to check if there's already a link to its path (can happen when we delete or disconnect / reconnect)
- when a node graph is deleted, we ensure we mark all its source prims as dirty, so that they re-evaluate and set their connections properly in arnold (we used to have dangling links to nodes that were destroyed)

Fixes #1158


render_delegate/light.cpp
```diff
@@ -620,6 +620,49 @@ void HdArnoldGenericLight::Sync(HdSceneDelegate* sceneDelegate, HdRenderParam* r
...
         }
         AiNodeSetDisabled(_light, !sceneDelegate->GetVisible(id));
+    
+        // get the sdf path for the light shader arnold node graph container
+        SdfPath lightShaderPath;
+        ArnoldUsdCheckForSdfPathValue(sceneDelegate->GetLightParamValue(id, TfToken("primvars:arnold:shaders")),
+                                      [&](const SdfPath& p) { lightShaderPath = p; });
+
+        AtNode *color = nullptr;
+        AtNode *shader = nullptr;
```
jhodgson, line 630: `AtNode *shader = nullptr;` is unused

render_delegate/render_delegate.cpp
```diff
@@ -1221,80 +1239,88 @@ bool HdArnoldRenderDelegate::ShouldSkipIteration(HdRenderIndex* renderIndex, con
...
+    // sources as being dirty, so that they can update their 
+    // new reference properly
+    while (_dependencyRemovalQueue.try_pop(id)) {        
+        auto it = _targetToSourcesMap.find(id);
+        if (it != _targetToSourcesMap.end()) {
+            skip = true; // this requires a render update
+            for (const auto& source : it->second) {
+                // for each source referencing the current target
+                // we need to remove the target from its list
+                auto &sourceIt = _sourceToTargetsMap.find(source);
```
jhodgson, line 1273: This should be a const reference `const auto &sourceIt = _sourceToTargetsMap.find(source);`

jhodgson (commented):
```
    HDARNOLD_API
    AtNode* HdArnoldRenderDelegate::GetSubdivDicingCamera(HdRenderIndex* renderIndex);
```
We need to remove `HdArnoldRenderDelegate::` in render_delegate.h line 453.

The PR is working well and is fixing crashed deleting light filters.

sebastienblor:
Fixed the build issues in the last commit

jhodgson (approved):

PR #1166, opened 2022-05-24
Support double array primvars in the procedural

### LINKED ISSUE #1165: Double array primvars not supported in the procedural
If a shape has double array primvars, they're properly read in the render delegate, but not in the procedural

sebastienblor:
When reading double array attributes in the procedural, we read them by converting them to float array attrs, like the render delegate does.
Added test 231

Fixes #1165


jhodgson (approved):
LGTM

PR #1169, opened 2022-05-29
Remove attributes reset in the lights delegate

### LINKED ISSUE #1168: Regression in lights colors in the delegate
Arnold 7.1.2 introduced a regression in the lights colors, when rendered in Hydra.

sebastienblor:
Before checking the shaders connections to lights, we call `AiNodeResetParameter` on the light color to ensure IPR updates behave properly. But this is now overriding the code setting the light color attribute.
After looking more in depth at the code, there is already a `AiNodeReset(_light)` that resets the whole light before any attribute is set, so we don't actually need to reset any attribute in particular.
This PR removes the 3 attribute resets from this code, which fixes the issue with the lights colors

Fixes #1168


jhodgson (approved):
LGTM

PR #1173, opened 2022-05-31
Support dataWindowNDC

### LINKED ISSUE #1161: Support overscan with dataWindowNDC
The render settings primitive can have an attribute dataWindowNDC meant for overscan (or region).
We should support it in both the render delegate and the reader

sebastienblor:
This PR adds support for dataWindowNDC. When set to a non-default value, we use the options region min/max x/y.
Something special as compared to arnold though, is that we don't want the output resolution to change. In other words, the desired resolution must match the window.
This means we need to have  `region_max_x - region_min_x = xres - 1`  and `region_max_y - region_min_y = yres - 1` .
In order to make this work properly, we must then adjust `options.xres options.yres`  accordingly.

For example, in the case of overscan, we'll set the options xres, yres, to a resolution that is lower than the expected one, but since the window will go beyond [0,0,1,1] the actual amount of pixels will be equal to the desired res.

This also means that if someone sets a very small window, e.g. [0.5, 0.5, 0.6, 0.6], then we might need to increase dramatically xres, yres (x10 in this example). 

Fixes #1161


sebastienblor:
In order to match correctly the exact window required by the user, I'm now setting the pixel_aspect_ratio accordingly, so that the full frustum matches the expected window. This way, the image can be "strecthed" if the requested window is higher in one dimension

jhodgson (approved):
Looks good to me, it's now working in the non-isotropic case.

PR #1216, opened 2022-07-15
Don't resample per-vertex data when translating pinned curves.

AndrewSidwell:
Normally a curve with n vertices has n-2 primvar values (e.g. radius, texture coords etc.), so it is necessary to resample per-vertex primvars to achieve the correct number of values. However with pinned curves, the renderer internally adds an extra two positions at the start and end of the curve so that curve passes through the original start and end position. Therefore it is not necessary to resample per-vertex data to achieve the correct number of primvar values (the renderer recognizes this as a special case when dealing with pinned curves).


translator/reader/read_geometry.cpp
```diff
@@ -362,14 +362,14 @@ void UsdArnoldReadMesh::Read(const UsdPrim &prim, UsdArnoldReaderContext &contex
 class CurvesPrimvarsRemapper : public PrimvarsRemapper
 {
 public:
-    CurvesPrimvarsRemapper(bool remapValues, ArnoldUsdCurvesData &curvesData) : 
+    CurvesPrimvarsRemapper(bool remapValues, bool pinnedCurve, ArnoldUsdCurvesData &curvesData) :
```
jhodgson, line 365: We need to set the member var `_pinnedCurve(pinnedCurve), `

jhodgson (changes requested):

translator/reader/read_geometry.cpp
sebastienblor: indeed, test_0184 was failing and after the proposed change it works again

sebastienblor:
Pushed a fix for @jhodgson comment regarding the uninitialized variable. I also updated test_0239 that now shows a slightly different result at the tip of the hair (as expected)

PR #1219, opened 2022-07-15
Fix build issue with older versions of Arnold 

### LINKED ISSUE #1217: Build fails with Arnold version under 7.1.3 because of AiMaterialxGetOslShaderCode error
**Additional context**
We need to build Arnold-Usd at commit 4a2a44fa69210eef562a84614360c191b2975832 in order to make some AOV features work with htoa in Houdini.

**Describe the bug**
When we try to build arnold-usd with a version of Arnold that is under 7.1.3 (in that case, 7.1.2.2), we get the following error:
```
render_delegate/node_graph.cpp: In member function 'AtNode* pxrInternal_v0_21__pxrReserved__::HdArnoldNodeGraph::GetMaterialxShader(const AtString&, const AtString&)':
render_delegate/node_graph.cpp:746:82: error: 'AiMaterialxGetOslShaderCode' was not declared in this scope
         AtString oslCode = AiMaterialxGetOslShaderCode(nodeType.c_str(), "shader");
```

This is due to the fact that the ```AiMaterialxGetOslShaderCode``` function doesn't exist in Arnold if the version is under 7.1.3.
It seems that you tried to handle this version compatibility in the ```node_graph.cpp``` file at line ```line 734```. There is a preprocessor command that tests if the Arnold version is under 7.1.3. If it is the case, it returns a null pointer:

```
AtNode *HdArnoldNodeGraph::GetMaterialxShader(const AtString &nodeType, const AtString &nodeName)
{
#if ARNOLD_VERSION_NUMBER < 70103
    return nullptr;
```

But, in the following lines, instead of putting the code for the Arnold >= 7.1.3 version under an ```#else``` preprocessor command, there is an ```#endif``` command that stops the condition. As a consequence, this part of the code is still compiled when Arnold version < 7.1.3:

```
#endif
    const char *nodeTypeChar = nodeType.c_str();
    if (nodeType == str::ND_standard_surface_surfaceshader) {
        AtNode *node = AiNode(_renderDelegate->GetUniverse(), str::standard_surface, nodeName);
        return node;
    } else if (nodeType.length() > 3 && nodeTypeChar[0] == 'N' && nodeTypeChar[1] == 'D' && nodeTypeChar[2] == '_') {
        // Create an OSL inline shader
        AtNode *node = AiNode(_renderDelegate->GetUniverse(), str::osl, nodeName);
        // Get the OSL description of this mtlx shader. Its attributes will be prefixed with 
        // "param_shader_"
        AtString oslCode = AiMaterialxGetOslShaderCode(nodeType.c_str(), "shader");
        // Set the OSL code. This will create a new AtNodeEntry with parameters
        // based on the osl code
        AiNodeSetStr(node, str::code, oslCode);
        return node;
    }
    return nullptr;
}
```

This behavior doesn't seem to have been changed in the next commits on the master branch.

**To Reproduce**
Build arnold-usd with an Arnold version under 7.1.3.

**Expected behavior**
Unless I'm missing something, the MaterialX part of the code should not be built.

**Used Software Versions**
 - Arnold:: 7.1.2.2
 - USD: 21.08
 - Compiler: GCC 6.3.1
 - OS: CentOS Linux release 7.6.1810 (Core)

**Patch File**
Here is the patch file we use for now:

```
--- render_delegate/node_graph.cpp.orig	2022-07-15 10:33:50.932998000 +0200
+++ render_delegate/node_graph.cpp	2022-07-15 10:36:10.535882295 +0200
@@ -733,7 +733,7 @@
 {
 #if ARNOLD_VERSION_NUMBER < 70103
     return nullptr;
-#endif
+#else
     const char *nodeTypeChar = nodeType.c_str();
     if (nodeType == str::ND_standard_surface_surfaceshader) {
         AtNode *node = AiNode(_renderDelegate->GetUniverse(), str::standard_surface, nodeName);
@@ -750,6 +750,7 @@
         return node;
     }
     return nullptr;
+#endif
 }
 bool HdArnoldNodeGraph::ClearUnusedNodes()
 {
```

Thank you

sebastienblor:
As described in #1217, the ifdefs weren't correctly placed. Fixing it so that we don't try to call APIs that don't exist for a given version of arnold

Fixes #1217

### DIFF
--- a/render_delegate/node_graph.cpp
+++ b/render_delegate/node_graph.cpp
@@ -733,7 +733,7 @@ AtNode *HdArnoldNodeGraph::GetMaterialxShader(const AtString &nodeType, const At
 {
 #if ARNOLD_VERSION_NUMBER < 70103
     return nullptr;
-#endif
+#else
     const char *nodeTypeChar = nodeType.c_str();
     if (nodeType == str::ND_standard_surface_surfaceshader) {
         AtNode *node = AiNode(_renderDelegate->GetUniverse(), str::standard_surface, nodeName);
@@ -755,6 +755,7 @@ AtNode *HdArnoldNodeGraph::GetMaterialxShader(const AtString &nodeType, const At
     }
 
     return nullptr;
+#endif
 }
 bool HdArnoldNodeGraph::ClearUnusedNodes()
 {


cpichard (approved):
lgtm !

PR #1222, opened 2022-07-18
Disable MaterialX shaders unless variable is defined

sebastienblor:
This PR temporarily comments out the code needed to read materialx shaders, both in the procedural and in the delegate, as the USD static libs currently shipped with Arnold don't include the materialx plugin. An environment variable can restore it.

In the Cmake files, I'm removing the previous materialx build variables that are no longer relevant.

I'm commenting out the tests based on the arnold version since they're not needed with the new build variable. But I'm keeping them as comments so that we don't forget to restore them when the build variable is no longer needed

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -101,6 +101,10 @@ if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
     endif ()
 endif ()
 
+if (ENABLE_MATERIALX)
+    add_compile_options(-D_ARNOLD_MATERIALX=1)
+endif ()
+
 # Common directory includes.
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -101,6 +101,10 @@ if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
     endif ()
 endif ()
 
+if (ENABLE_MATERIALX)
+    add_compile_options(-D_ARNOLD_MATERIALX=1)
```
cpichard, line 105: there is an underscore before ARNOLD_MATERIALX, is it on purpose ?

PR #1227, opened 2022-07-25
re-export core usd libraries symbols on macOS

### LINKED ISSUE #1226: procedural dynamic link issue on macOS
**Describe the bug**
On macOS the procedural does not export the weak external symbols which creates duplicated objects at runtime and can potentially crash an application, specially when type_info objects are compared to determine which type a VtValue holds.
This is similar to https://github.com/PixarAnimationStudios/USD/issues/1475
This issue should be addressed at some point by Pixar, in the meantime we need to find a workaround.
Also filled internally as MTOA-837

cpichard:
- on macOS it allows to export only the symbols belonging to USD. Most of them are weak external symbols.

Fixes #1226

### DIFF
--- a/procedural/macos_export_list
+++ b/procedural/macos_export_list
@@ -1,2 +1,5 @@
 _NodeLoader
 _SceneFormatLoader
+*pxrInternal_*Gf*
+*pxrInternal_*Vt*
+*pxrInternal_*Tf*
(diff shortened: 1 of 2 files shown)


sebastienblor (approved):
looks good !

PR #1234, opened 2022-07-28
Issue 1224

### LINKED ISSUE #1224: Investigate better strategies to handle udims with relative paths.
Today, possibly because we're still relying on ArResolver v1, udims with relative paths can't be resolved properly by USD.
Because of that, we try to identify which layer they come from, in order to resolve the path ourselves, but to do that we expect the 1001 udim to be present on disk (see https://github.com/Autodesk/arnold-usd/blob/master/translator/reader/utils.h#L562 )

However, this is not a good strategy as this udim token might not exist on disk. We should investigate if there are better strategies to handle these use cases

cpichard:
- Use UsdAttribute instead of UsdPrim to find the layer relative to a SdfAssetPath stored in an attribute. It reduces the number of layer searched.
- Test if the directory of the resolved global path exists without checking if the file itself exists. It should allow to have tags in the filename. 

Fixes #1224


translator/reader/read_geometry.cpp
```diff
@@ -1044,7 +1044,7 @@ void UsdArnoldReadVolume::Read(const UsdPrim &prim, UsdArnoldReaderContext &cont
         VtValue vdbFilePathValue;
 
         if (vdbAsset.GetFilePathAttr().Get(&vdbFilePathValue, time.frame)) {
-            std::string fieldFilename = VtValueGetString(vdbFilePathValue, &prim);
+            std::string fieldFilename = VtValueGetString(vdbFilePathValue, nullptr);
```
sebastienblor, line 1047: shouldn't this have the attribute `vdbAsset.GetFilePathAttr()` instead of nullptr ?

translator/reader/read_options.cpp
```diff
@@ -416,13 +413,11 @@ void UsdArnoldReadRenderSettings::Read(const UsdPrim &prim, UsdArnoldReaderConte
...
                 // we found the same aov name multiple times, we'll need to add the layerName
                 useLayerName = true;
-                // store the list of aov names that were actually duplicated
-                duplicatedAovs.insert(aovName);
             }
             VtValue aovNameValue;
             // read the parameter "driver:parameters:aov:name" that will be needed if we have merged exrs (see #816)
             std::string layerName = (renderVarPrim.GetAttribute(_tokens->aovSettingName).Get(&aovNameValue, time.frame)) ? 
-                VtValueGetString(aovNameValue, &prim) : renderVarPrim.GetPath().GetName();
+                VtValueGetString(aovNameValue, nullptr) : renderVarPrim.GetPath().GetName();
```
sebastienblor, line 420: same as above, I guess we need to pass `renderVarPrim.GetAttribute(_tokens->aovSettingName)` ?

sebastienblor:
I think we need to merge the master branch back to this `issue_1224 `branch, I see the inverted diffs of one of the last commit from  1221

translator/reader/read_geometry.cpp
cpichard: It should ! I'll push an update.

cpichard:
I'll also change the test number to avoid a conflict with the current pull request related to cylinder light.

sebastienblor (approved):
All good, and passing the testsuite

PR #1239, opened 2022-08-01
Fix crash with varying instances #1238

### LINKED ISSUE #1238: Crash with varying instance count
In scenes where the amount of instances changes accross the shutter, we can get into crashes inside the procedural code.
Attaching a scene that reproduces the issue
[test.zip](https://github.com/Autodesk/arnold-usd/files/9236745/test.zip)

sebastienblor:
When a node has varying instance count and we export a primvar, we ignore the animation and store a single key. But we weren't reallocating the attribute array, as compared to the shutter start. This PR fixes the issue and adds a test for this ticket 

Fixes #1238


cpichard (commented):
I think that should work if I understood correctly the code. There are 2 things bothering me. First in case of different array sizes, the function chooses the first one that is different, which might not be the most appropriate. If the samples have the following sizes [100,100,100,0], because of some bad export, the function will choose the array with the empty value. Which brings to the second question, should we also test for empty arrays ?

sebastienblor:
no, it won't pick the first that is different, it will pick the "current" time (for shutter=0), which is the time at which the other attributes were read (points count, etc...). 
The variable `timeVal` represents the motion time at different shutter values, but `time.frame` is the current "base" time.

If we pick a different time, then we can end up with a mismatch between e.g. the points count and the primvars count, and this can easily crash. That's why we need to do all this mess with the resizing, etc... We could pick this base time in the first place, instead of the min shutter, but then we could be reading one additional sample which could make it less efficient.
So here we just have an overhead when we get to a situation where one of the samples count is different

sebastienblor:
that's right, we could add an additional test for empty arrays

sebastienblor:
Added verification for null-size arrays

cpichard (approved):
Right, it's using `time.frame` for the fallback value, my bad.
Looks good to me !

PR #1248, opened 2022-08-03
Compute material assignments in the "full" purpose

### LINKED ISSUE #1245: Read Material bindings in the full purpose
When using the material binding API to compute the bound material, we're currently using the default "allPurpose" purpose. 
This will work in most cases, but if a scene has different material assignments for rendering ("full") and preview, then we might not render the correct material.

sebastienblor:
We're currently reading material assignments in the "allPurpose" purpose. But if different materials are assigned in "full" or in "preview" we might not render the correct ones. This PR computes the material binding in the "full" purpose as explained in https://graphics.pixar.com/usd/dev/api/class_usd_shade_material_binding_a_p_i.html#a145195b04bd6367016edf7986f1120bd

I've ran the testsuite and verified that the existing material assignments in the all purpose still render properly.
Added test_1245 with such an example

Fixes #1245


translator/reader/utils.cpp
```diff
@@ -186,7 +186,8 @@ AtArray *ReadLocalMatrix(const UsdPrim &prim, const TimeSettings &time)
 static void getMaterialTargets(const UsdPrim &prim, std::string &shaderStr, std::string *dispStr = nullptr)
 {
 #if PXR_VERSION >= 2002
-    UsdShadeMaterial mat = UsdShadeMaterialBindingAPI(prim).ComputeBoundMaterial();
+    // We want to get the material assignment for the "full" purpose, which is meant for rendering
```
cpichard, line 189: is this code executed in the render delegate as well ? I am wondering if this change can/will affect the interactivity in the viewport, any idea ?

cpichard (approved):

PR #1249, opened 2022-08-03
Fix adaptive sampling in the render delegate

### LINKED ISSUE #1006: Render settings not passed to the render delegate via husk

sebastienblor:
When rendering through Solaris, the "beauty" output was provided by solaris as a Light Path Expression `C.*` for an aov named `color` . Since this lpe refers to the beauty, arnold was rendering this correctly. Unfortunately, algorithms like adaptive sampling rely on an explicit beauty aov (called `RGBA` in arnold), and since this AOV wasn't present through Solaris, adaptive sampling couldn't work. 
This PR fixes it by simply considering the beauty as arnold would expect. 

Fixes #1006


jhodgson (approved):
Looks good to me, HtoA tests are passing

PR #1267, opened 2022-08-18
Add env variable in the delegate to save the arnold scene before rendering it

### LINKED ISSUE #1266: Add env variable in the delegate to dump the arnold scene before rendering it
For debugging purpose, we often need to dump the arnold scene just before it gets rendered, in order to understand how the render delegate translates it.
Since this is just for internal debugging and we don't want to expose a user parameter for this, we could look at a defined environment variable that does it

sebastienblor:
When the env variable `HDARNOLD_DEBUG_SCENE` is defined, the render delegate will save it out before each render iteration. 
This allows to debug how the delegate translates hydra data into arnold.

Fixes #1266

### DIFF
--- a/render_delegate/render_param.cpp
+++ b/render_delegate/render_param.cpp
@@ -28,10 +28,12 @@
 #include "render_param.h"
 #include "render_delegate.h"
 #include <constant_strings.h>
-
+#include <pxr/base/tf/envSetting.h>
 #include <ai.h>
 
 PXR_NAMESPACE_OPEN_SCOPE
+
+TF_DEFINE_ENV_SETTING(HDARNOLD_DEBUG_SCENE, "", "Optionally save out the arnold scene before rendering.");
 
 #ifdef ARNOLD_MULTIPLE_RENDER_SESSIONS
 HdArnoldRenderParam::HdArnoldRenderParam(HdArnoldRenderDelegate* delegate) : _delegate(delegate)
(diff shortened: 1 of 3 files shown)


cpichard (commented):
I think it would be great to use the usd TfDebug facilities for this features https://graphics.pixar.com/usd/release/api/class_tf_debug.html
using a debug code TF_DEBUG_CODE
https://graphics.pixar.com/usd/release/api/group__group__tf___debugging_output.html#ga798d0b335c62eb79f5cd3ca35361932f

Activating or deactivating a variable can be triggered by environment variable, but also at runtime which could come handy, there is a UI in usdview which allows to specifically do that, and probably one in Houdini, but I haven't checked.

one example here https://github.com/PixarAnimationStudios/USD/blob/dc710925675f7f58f7a37f6c18d046a687b09271/pxr/imaging/hgiGL/diagnostic.cpp#L40

render_delegate/render_param.cpp
```diff
@@ -40,6 +41,11 @@ HdArnoldRenderParam::HdArnoldRenderParam()
 {
     _needsRestart.store(false, std::memory_order::memory_order_release);
     _aborted.store(false, std::memory_order::memory_order_release);
+    // If the HDARNOLD_DEBUG_SCENE env variable is defined, we'll want to 
+    // save out the scene every time it's about to be rendered
+    const char *HDARNOLD_DEBUG_SCENE = std::getenv("HDARNOLD_DEBUG_SCENE");
+    if (HDARNOLD_DEBUG_SCENE) 
+        _debugScene = std::string(HDARNOLD_DEBUG_SCENE);
```
cpichard, line 48: I am not sure it is safe to construct a string with a nullptr, which would happen if `HDARNOLD_DEBUG_SCENE` is not set

sebastienblor: if the variable is not set, then the const char* `HDARNOLD_DEBUG_SCENE`would be null, and so we wouldn't contruct a string from it, right ? We have the same code for OCIO config files variables

cpichard: I don't know why I didn't see the condition, all good.

cpichard (dismissed):

sebastienblor:
I'm now registering the env variable using `TF_DECLARE_ENV_SETTING` and querying the value with `TfGetEnvSetting`

PR #1283, opened 2022-09-01
Avoid crashes when invalid primitive path is set for a volume's field

### LINKED ISSUE #1281: Crash when volume density prim is invalid
When a volume prim has a density field pointing to an invalid prim path, the procedural crashes. We should ensure we don't crash but just dump an error

sebastienblor:
In the volume reader, we were testing if the field density primitive had the proper type (UsdVolOpenVDBAsset) but we were simply not testing if the primitive existed.  So if the path was invalid we were crashing. 

Fixes #1281


cpichard (approved):
Looks good to me

PR #1303, opened 2022-09-23
adds plugin resources when INSTALL_USD_PLUGIN_RESOURCES is set

### LINKED ISSUE #1302: Install the resources of the plugins compiled with the static usd lib
In the arnold procedural we can embed plugins statically, however to properly work, they need to access their resources files.
We need to also install those files when we build the procedural with plugins

cpichard:
- Setting INSTALL_USD_PLUGIN_RESOURCES will trigger the copy of the plugin/usd folder onto the install destination 

Fixes #1302

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -617,7 +617,6 @@ if PROCEDURAL:
     INSTALL_PROC += env.Install(os.path.join(PREFIX_HEADERS, 'arnold_usd'), ARNOLDUSD_HEADER)
     if env['USD_BUILD_MODE'] == 'static':
         INSTALL_PROC += env.Install(PREFIX_PROCEDURAL, usd_input_resource_folder)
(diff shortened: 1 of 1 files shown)


SConstruct
```diff
@@ -607,6 +617,7 @@ if PROCEDURAL:
     INSTALL_PROC += env.Install(os.path.join(PREFIX_HEADERS, 'arnold_usd'), ARNOLDUSD_HEADER)
     if env['USD_BUILD_MODE'] == 'static':
         INSTALL_PROC += env.Install(PREFIX_PROCEDURAL, usd_input_resource_folder)
+        INSTALL_PROC += env.Install(PREFIX_PROCEDURAL, usd_input_resource_folder)
```
sebastienblor, line 620: isn't it the same line that is being duplicated ?

cpichard: it is, I didn't notice it, thanks !

sebastienblor (approved):

Issue #1232, opened 2022-07-28
Set output to half precision for RenderVars with half types


jhodgson:
If we create renderVars with half precision (we currently use the dataType and then override with the format) we need to add HALF at the end of the output string for that AOV in read_options.cpp.  

The usual case is that the format is set to color3h for the RGB case, but there may be other scenarios such as as color4h (RGBA), half3 (RGB), half4 (RGBA), half2 (VECTOR2) and half (FLOAT) to support.

jhodgson:
[aov_setup_manual_add_RGBA_from_subnet_v3.usd_rop1.usda.zip](https://github.com/Autodesk/arnold-usd/files/9209209/aov_setup_manual_add_RGBA_from_subnet_v3.usd_rop1.usda.zip)

jhodgson:
Example file with half float AOVs

PR #1308, opened 2022-10-02
Set driver's half_precision attribute

sebastienblor:

This PR is just a derived form of the changes proposed in #1292 , it's only partially addressing #1232 .
The main difference is that it just handles the driver's attribute `half_precision`. 

The proposed rationale is that if the beauty AOV is half, we set `half_precision` to true. 
If the driver doesn't have any beauty AOV, and they're all set to half, then we also set this attribute to true.

What is still missing is the use case where, in a given driver, some AOVs are half but others aren't. This requires to set the HALF token in the outputs string for the corresponding AOVs, which is a feature that is supported in some arnold plugins but not all. 

The reason we're not including this yet, is because it triggers an arnold bug in kick that is being addressed separately. 
However, just having a single "half" status for the whole driver is already a great improvement that we want to include asap.

I'm adding test_1232 with the scene provided in the ticket. I verified that removing the half status in the beauty does break the test.


translator/reader/read_options.cpp
```diff
@@ -458,7 +476,38 @@ void UsdArnoldReadRenderSettings::Read(const UsdPrim &prim, UsdArnoldReaderConte
                 outputs[j + prevOutputsCount] += std::string(" ") + layerNames[j];
             }
         }
+        // For exr drivers, we need to set the attribute "half_precision"
+        if (!isHalfList.empty() && driverType == "driver_exr") {
```
cpichard, line 480: tif should also be able to handle half precision, do we want to add it as well ?

cpichard (approved):

translator/reader/read_options.cpp
sebastienblor: not in Arnold. 
Here we're setting the attribute "half_precision" in the arnold driver node type "driver_exr". We don't have such attribute in "driver_tiff".

PR #1310, opened 2022-10-03
Fix default width in rendervar's filters

### LINKED ISSUE #1309: RenderVar's filter width not translated properly in the reader
A renderVar prim can have custom attributes to determine the arnold filter, and filter width.

But when the filter width isn't set, it's forced to a default value of 1, which doesn't match the arnold defaults

sebastienblor:
We were setting the render var's filters width to 1 by default, in the USD reader. Now we're only setting it if the attribute is defined in the renderVar prim, so we rely on the arnold filter's default values.

Fixes #1309


cpichard (approved):
Looks good to me !

PR #1312, opened 2022-10-03
Support AOV names called "color" for the beauty

### LINKED ISSUE #1311: Support AOVs called "color" for beauty in the reader
Render var's aov names can be called "color" for the beauty, when they come from Houdini Solaris.
This is supported in the delegate but not in the translator, we should support it there too

sebastienblor:
In this PR we check for AOV names called "color" and assume this is the beauty. We translate this as a "RGBA" AOV in Arnold.
I'm adding test_1311 with such a scene

Fixes #1311


jhodgson (approved):
LGTM

PR #1322, opened 2022-10-14
use SdfPathVector for aov_shaders

### LINKED ISSUE #1321: Allow list of AOV ArnoldNodeGraph primitives in RenderSettings
The aov_shader string on the RenderSettings refers to an ArnoldNodeGraph, and this is used in the render delegate and the reader.
However sometimes two ArnoldNodeGraphs need to be referenced, for example when using the Arnold Cryptomatte AOV.  This creates one ArnoldNodeGraph, but other custom AOVs need to be created using another ArnoldNodeGraph.
The solution is to gather the AOV shaders from more than one ArnodNodeGraph primitive.  The change will be in the arnold-usd repository, and will be required in both the reader and the delegate.

cpichard:
- aov_shaders primvar stores paths as a string. With this change, we can now add multiple paths separated by space. Internally they are stored as SdfPathVector.
- 

Fixes #1321


jhodgson (dismissed):
Nice, looks good for the delegate.

sebastienblor (approved):

PR #1337, opened 2022-10-27
Ensure empty product names don't render as ".exr" 

### LINKED ISSUE #1336: Empty ProductName renders ".exr"
If a render product has productName set to an empty string, this will end up with Arnold trying to render a file ".exr".
Whereas, if the attribute isn't set at all, it uses the render product prim name as a basename for the output image.
Both use cases should behave the same way

sebastienblor:
When product name isn't set at all, we're using the render product prim name as an output filename.
This PR checks if the product name is set to an empty value, and does the same in that case

Fixes #1336


cpichard (approved):
lgtm

PR #1350, opened 2022-11-14
Declare metadatas in Sdr registry

### LINKED ISSUE #55: Handle metadata in the Ndr plugin and store arnold specific metadata
We should handle arnold metadata and store additional information in the ndr plugins.

This include:
- metadata on arnold nodes
- metadata on arnold parameters
- metadata storing the arnold type of the parameter (there are arnold types that map to the same sdf type)

We need to extend `ndr/utils.cpp` to deal with metadata. Add additional USD metadata to store the raw arnold type on the primitives in the Usd Stage returned by `NdrArnoldGetShaderDefs`. And pass metadata to Usd in the Discovery and Parser plugins.

sebastienblor:
This PR declares some arnold metadatas as Sdr info in the registry.
The list of supported metadatas are the following one
USD builtin metadatas :
- page :  gives the UI group to put this attribute under. It's based on the arnold metadata "ui.groups"
- label : UI label of the parameter. Based on arnold attr metadata "label"
- help : description of the parameter. Based on existing "desc" arnold metadata
- connectable : whether this parameter is connectable or not. Based on existing "linkable" metadata. Note that in Arnold it defaults to true, so it won't be set for a lot of parameters that are actually connectable. To be verified if this is similar in USD or if we should always declare it
- role (returns the shader classification/category)

Custom metadatas (declared as "hints")
- uimin : based on arnold "min" metadata
- uimax : based on arnold "max" metadata
- uisoftmin : based on arnold "softmin" metadata
- uisoftmax : based on arnold "softmax" metadata
 
Enums:
enum attributes now declare all the allowed enum values in the "options" field

Fixes #55


ndr/parser.cpp
```diff
@@ -99,6 +105,86 @@ NdrArnoldParserPlugin::NdrArnoldParserPlugin() {}
...
+    VtValue v;
+    attr.Get(&v);
+    // The utility function takes care of the conversion and figuring out
+    // parameter types, so we just have to blindly pass all required
+    // parameters.
+    VtDictionary customData = attr.GetCustomData();
+    NdrTokenMap metadata;
+    NdrTokenMap hints;
+
+    // For enum attributres, all enum fields should be set as "options"
```
cpichard, line 130: typo `attributres` -> `attributes`

cpichard (approved):

PR #1363, opened 2022-11-26
Skip curves without any width

### LINKED ISSUE #1334: Skip curves with empty radius
When curves have empty radius in a usd scene, the procedural could skip them instead of creating such curves in Arnold (which errors out). In a USD file they could be present for other reasons, so it shouldn't cause an error

sebastienblor:
In this PR we ignore curve primitives which don't have any width attribute set.

Fixes #1334


cpichard (dismissed):
lgtm !
Just one note, on line 493 the widths are queried again, I think it's not necessary anymore to query them back again and removing the line might save some memory operation

sebastienblor:
> lgtm ! Just one note, on line 493 the widths are queried again, I think it's not necessary anymore to query them back again and removing the line might save some memory operation

fixed

cpichard (approved):

PR #1373, opened 2022-12-09
Read primitives with empty purpose

### LINKED ISSUE #1372: Render prims with purpose set to empty string
We have scenes generated by MaxUSD where primitives have `purpose` set to an empty string (instead of leaving the attribute untouched or set to the default value "default").

In usd_view, these objects show up, but in Arnold they don't because we only check if the purpose is "default" or the expected one ("render" in this case). We should also account for empty values to match usd_view

sebastienblor:
In this PR, when we find a primitive with `purpose` set to an empty string, we consider this to be equivalent to the default value `default` .

Fixes #1372

### DIFF
--- a/translator/reader/reader.cpp
+++ b/translator/reader/reader.cpp
@@ -250,13 +250,16 @@ void UsdArnoldReader::TraverseStage(UsdPrim *rootPrim, UsdArnoldReaderContext &c
             UsdGeomImageable imageable(prim);
             bool pruneChildren = false;
             attr = imageable.GetVisibilityAttr();
-            if (attr && attr.HasAuthoredValue())
+            if (attr && attr.HasAuthoredValue()) {
+
                 pruneChildren |= (attr.Get(&visibility, frame) && 
                         visibility == UsdGeomTokens->invisible);
+            }
 
             attr = imageable.GetPurposeAttr();
             if (attr && attr.HasAuthoredValue()) {
                 pruneChildren |= ((attr.Get(&purpose, frame) && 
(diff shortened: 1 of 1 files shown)


translator/reader/reader.cpp
```diff
@@ -250,13 +250,16 @@ void UsdArnoldReader::TraverseStage(UsdPrim *rootPrim, UsdArnoldReaderContext &c
...
+            if (attr && attr.HasAuthoredValue()) {
+
                 pruneChildren |= (attr.Get(&visibility, frame) && 
                         visibility == UsdGeomTokens->invisible);
+            }
 
             attr = imageable.GetPurposeAttr();
             if (attr && attr.HasAuthoredValue()) {
                 pruneChildren |= ((attr.Get(&purpose, frame) && 
+                        !purpose.IsEmpty() &&
```
cpichard, line 262: it looks like the only allowed values are "default, render, proxy, guide", so the scene shouldn't have an authored empty purpose. Testing it is ok, but it might be the case where the scene with empty purpose won't work with other usd compatible applications or will have other meaning. 
Should we also add a warning ligne in debug mode ? just in case we have to debug scenes in the future ?

cpichard (approved):

translator/reader/reader.cpp
sebastienblor: actually, the reason we're doing this change is because other tools (usd_view, mayausd, etc..) are showing these primitives.
It might not be intentional, as the docs don't mention it, but at least we'll now be consistent.

PR #1381, opened 2022-12-15
Interactive GPU rendering crashes with MaterialX shader #1380

### LINKED ISSUE #1380: Interactive GPU rendering crashes with MaterialX shader
**Describe the bug**
Interactive GPU rendering crashes when tweaking MaterialX shader with an Image input.

**To Reproduce**
1. Extract the attached to a drive "IPRGPU_Mtlx_crash.rar"
2. Open "IPRGPU_Crash2.hip"
3. Make sure IPR is set to Arnold and GPU is enabled
4. Note the MaterialX shader base color is connected to an Image node.
5. Tweak the base slider in the mtlx shader or try changing the texture file in the Image node.
6. Result: Crash

**Expected behavior**
Tweaking the base slider or changing the texture file should not result in a crash.

**Used Software Versions**
 - Arnold 7.1.5
 - USD: e.g. 0.22.08
 - HtoA 6.1.5
 - Houdini 19.5.368

juandejoya:
When we sync `HdArnoldNodeGraph` we clear out unused nodes that have not been used during material translation, including nodes OSL nodes such as `mtlximage1_texturesource`. What's happening is that, once the OSL node `mtlximage1_texturesource` exists as a parameter for a Material node `mtlximage1`, we do not flag it be used in the next render session. This results in a crash where the texture can't be found and the shader is not compiled, which leads to a OOM crash.  

If the OSL node already exists as a node, we would need to update the node data's used status so that it doesn't get cleared out as an unused node.  

Fixes #1380

**Additional context**
https://user-images.githubusercontent.com/2394669/207842569-c7fdaf58-e031-4cfe-bbd7-0bd6efd71a23.mp4


sebastienblor (dismissed):

cpichard (approved):
It also fixes the issue on mac and I guess linux.

render_delegate/node_graph.cpp
```diff
@@ -675,12 +675,19 @@ AtNode* HdArnoldNodeGraph::ReadMaterialNode(const HdMaterialNode& node)
                 if (resourceNodeIt != _nodes.end())
                     oslSource = resourceNodeIt->second->node;
 
+                // If the OSL node does not exist, create the node and the node data. Othewise, update the
```
cpichard, line 678: Othewise -> Otherwise

render_delegate/node_graph.cpp
```diff
@@ -675,12 +675,19 @@ AtNode* HdArnoldNodeGraph::ReadMaterialNode(const HdMaterialNode& node)
...
+                // If the OSL node does not exist, create the node and the node data. Othewise, update the 
+                // node data to so that it's assumed to be unused and is therefore not cleared out.
                 if (oslSource == nullptr) {
                     oslSource = AiNode(_renderDelegate->GetUniverse(), str::osl, AtString(resourceNodeName.c_str()));
                     AiNodeSetStr(oslSource, str::code, tx_code);
                     auto resourceNodeData = NodeDataPtr(new NodeData(oslSource, true));
                     _nodes.emplace(resourceNodePath, resourceNodeData); 
-                }                
+                } else {
+                    auto it = _nodes.find(resourceNodePath);
```
cpichard, line 686: isn't `it` already the same as `resourceNodeIt` ? so we might not have to call find again ?

juandejoya: Thanks, that's a good catch. I did a test with the suggested optimization and it looks like it's work the same, so can optimize in the next commit.

render_delegate/node_graph.cpp
juandejoya: Thanks, updated in the next commit.

sebastienblor (approved):

PR #1388, opened 2023-01-11
Ensure arnold_usd.h is included when needed

### LINKED ISSUE #1387: Deprecated arnold APIs still being used
We are still invoking AiUniverseIsActive which is deprecated. This is happening because the version check isn't working as it should

sebastienblor:
We were testing ARNOLD_VERSION_NUMBER to use `AiArnoldIsActive` instead of `AiUniverseIsActive` , but without including `../arnold_usd.h` that properly defines this variable

Fixes #1387

### DIFF
--- a/translator/reader/registry.cpp
+++ b/translator/reader/registry.cpp
@@ -26,7 +26,7 @@
 #include "read_options.h"
 #include "read_shader.h"
 #include "utils.h"
-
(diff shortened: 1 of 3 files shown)


ThiagoIze (approved):
This looks good to me.

cpichard (approved):

PR #1409, opened 2023-01-30
Always output layer names when explicitely set

### LINKED ISSUE #1408: Always read RenderVar aov:name as an output layer
Currently, when a RenderVar has a parameter `driver:parameters:aov:name` , it's considered to be the layer name of the AOV. But this only happens if there are duplicated AOVs for the same render product. 
Instead, we'd want the layer name to be always set if this attribute is explicitely set (and if it's different from the AOV sourceName

sebastienblor:
- 
- 

Fixes #1408


translator/reader/read_options.cpp
```diff
@@ -255,8 +255,12 @@ AtNode * DeduceDriverFromFilename(const UsdRenderProduct &renderProduct, UsdArno
         // The product name is supposed to return the output image filename.
         // If none is provided, we'll use the primitive name
     VtValue productNameValue;
-    std::string filename = renderProduct.GetProductNameAttr().Get(&productNameValue, time.frame) ?
-    VtValueGetString(productNameValue, nullptr) : renderProduct.GetPrim().GetName().GetText();
+    std::string filename = renderProduct.GetPrim().GetName().GetText();
```
cpichard, line 258: filename is it the name of the render product by default ?

cpichard (approved):

PR #1410, opened 2023-01-31
Imagers only updates in the render delegate 

### LINKED ISSUE #1320: Imager only updates in the render delegate
We need to detect when changes in the render delegate are coming only from the imager changes and not interrupt the render.

sebastienblor:
We're adding a flag to the HdArnoldNodeGraph class to specify if this is an imager graph. 
When it's modified interactively, instead of pausing and restarting the render, we update directly the arnold scene, and then notify arnold that the imagers were update (through the AiRenderHints APIs).

This allows the imagers to be updated without necessarily restarting the render

Fixes #1320

### DIFF
--- a/common/constant_strings.h
+++ b/common/constant_strings.h
@@ -389,6 +389,7 @@ ASTR(region_min_y);
 ASTR(render_device);
 ASTR(render_settings);
 ASTR(repeat);
+ASTR(request_imager_update);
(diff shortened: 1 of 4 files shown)


cpichard (approved):

sebastienblor:
closing for now, as we need to do more complex changes for older versions of Solaris

PR #1418, opened 2023-02-02
Declare filename attributes as asset Paths

### LINKED ISSUE #1163: Support asset attributes for filenames when no tokens are present
Currently, filename attributes are expressed as strings in arnold-usd, in order to support tokens that can only be resolved at render time.
This doesn't fit well the USD logic, where filenames should be expressed as assets.
We should check if a custom file resolver could allow us to definitively switch to asset attributes, otherwise we could have 2 usd attributes, one string and one asset, that would be authored or not depending on whether a token is present in the filename.

sebastienblor:
In this PR, we're authoring "filename" arnold attributes as SdfAssetPath in USD. This no longer requires changes in the usd reader, since we now handle the conversion at read-time, and support not resolving the usd paths.

In order to find out if an attribute is a filename, we look for its metadata `file` and check if it's set to `path`
The change also needs to be done in the Sdr registry, since we should now declare these attributes as being SdfAssetPath

Fixes #1163

### DIFF
--- a/ndr/utils.cpp
+++ b/ndr/utils.cpp
@@ -465,6 +465,20 @@ void _ReadArnoldShaderDef(UsdStageRefPtr stage, const AtNodeEntry* nodeEntry)
                 attr.Set(conversion->f(array));
             }
         } else {
+
+            // Some arnold string attributes can actually represent USD asset attributes.
+            // In order to identify them, we check for a specific metadata "path" set on this attribute
+            if (paramType == AI_TYPE_STRING) {
+                AtString pathMetadata;
+                if (AiMetaDataGetStr(nodeEntry, paramName, str::path, &pathMetadata) && 
+                        pathMetadata == str::file) {
+
+                    attr = prim.CreateAttribute(TfToken(paramName.c_str()), SdfValueTypeNames->Asset, false);
+                    SdfAssetPath assetPath(AiParamGetDefault(pentry)->STR().c_str());
+                    attr.Set(VtValue(assetPath));
+                    continue;
+                }
+            }
             const auto* conversion = _GetDefaultValueConversion(paramType);
             if (conversion == nullptr) {
                 continue;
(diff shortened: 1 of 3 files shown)


ndr/utils.cpp
```diff
@@ -465,6 +465,20 @@ void _ReadArnoldShaderDef(UsdStageRefPtr stage, const AtNodeEntry* nodeEntry)
...
+            // In order to identify them, we check for a specific metadata "path" set on this attribute
+            if (paramType == AI_TYPE_STRING) {
+                AtString pathMetadata;
+                if (AiMetaDataGetStr(nodeEntry, paramName, str::path, &pathMetadata) && 
+                        pathMetadata == str::file) {
+
+                    attr = prim.CreateAttribute(TfToken(paramName.c_str()), SdfValueTypeNames->Asset, false);
+                    SdfAssetPath assetPath(AiParamGetDefault(pentry)->STR().c_str());
+                    attr.Set(VtValue(assetPath));
+                    continue;
```
cpichard, line 479: I think the `continue` here will skip the copy of the metadata for this attribute, is it the expected behaviour ?

cpichard (approved):

JGamache-autodesk:
Sorry to barge in this late, but I think something is missing at shader node discovery level. The problem is that "Asset" is not in the set of valid SdrProperty property types.

This is reflected all over the USD code base with functions like [_GetShaderPropertyTypeAndArraySize()](https://github.com/PixarAnimationStudios/USD/blob/dev/pxr/usd/usdShade/shaderDefUtils.cpp#L228) being used to convert `SdfValueTypeNames->Asset->` into `SdrPropertyTypes->String`. And if you look at the code that [creates the SdrShaderProperty](https://github.com/PixarAnimationStudios/USD/blob/dev/pxr/usd/usdShade/shaderDefUtils.cpp#L325) for the UsdFileTexture node you will see that it adds `SdrPropertyMetadata->IsAssetIdentifier` to the attribute.

If you look elsewhere in the code base, you will see the same actions taken for MaterialX, OSL, and RMan shader nodes.

PR #1423, opened 2023-02-07
Generalize support for render delegate products

### LINKED ISSUE #1422: Support render delegate products for all arnold drivers
At the moment, we can set the render product "type" to "deep". In this case, instead of handling the drivers on the scene delegate side, it's up to the render delegate to handle the render product. In this case, during batch renders we create a `driver_deepexr` instead of a hydra driver. It would be useful to extend this to other arnold driver types

sebastienblor:
Support for Render delegate products is now generalized, and no longer hardcoded for `deep` . If one sets the name of an arnold driver in the product type field, then the arnold render delegate will create such a driver instead of relying on Hydra drivers.

For example, if we set `driver_exr` in the render product in Solaris, and then render with husk, the render rely on a driver_exr and therefore render as if it was using kick-usd. 

Still missing : pass on the arnold render product parameters (i.e. the list of driver attributes) to the driver. At the moment they're ignored and only the deep exr params are supported.

Fixes #1422

### DIFF
--- a/render_delegate/render_delegate.cpp
+++ b/render_delegate/render_delegate.cpp
@@ -63,12 +63,14 @@ TF_DEFINE_PRIVATE_TOKENS(_tokens,
     (arnold)
     (openvdbAsset)
     ((arnoldGlobal, "arnold:global:"))
+    ((arnoldDriver, "arnold:driver"))
     (percentDone)
     (delegateRenderProducts)
     (orderedVars)
     ((aovSettings, "aovDescriptor.aovSettings"))
     (productType)
     (productName)
+    (driver_exr)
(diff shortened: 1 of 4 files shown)


sebastienblor:
I changed the way we use the the product type. Now, every time we want an arnold driver to be used, we'll set the render product type to "arnold" (instead of the default "raster"). This will tell hydra that the render delegate will own the driver, and in that case we'll create a driver based on the render product parameter "arnold:driver". This now supports the custom render driver parameters (e.g. arnold:driver_exr:append").

cpichard (dismissed):

cpichard (approved):

PR #1439, opened 2023-03-01
Surface models from PBR nodes

### LINKED ISSUE #1438: Creating surface models from PBR nodes doesn't work
Creating a surface model using the PRB nodes doesn't work as expected, where the OSL nodes are created in Arnold, but they are not connected.

This needs a fix in Arnold, but parameters flagged as non-editable in the MaterialX shader generator will follow a different naming convention.

I verified that the following patch fixes the issue for the procedural (same things needs to happen in the render delegate)
```
diff --git a/translator/reader/read_shader.cpp b/translator/reader/read_shader.cpp
index 7018813..cdac10e 100644
--- a/translator/reader/read_shader.cpp
+++ b/translator/reader/read_shader.cpp
@@ -128,19 +128,23 @@ void UsdArnoldReadShader::Read(const UsdPrim &prim, UsdArnoldReaderContext &cont

         // Loop over the USD attributes of the shader
         for (const auto &attribute : attributes) {
-            std::string attrName = attribute.GetBaseName().GetString();
             // only consider input attributes
             if (attribute.GetNamespace() != str::t_inputs)
                 continue;

             // In order to match the usd attributes with the arnold node attributes,
             // we need to add the prefix "param_shader_"
-            attrName = "param_shader_" + attrName;
-
+            std::string attrName = "param_shader_" + attribute.GetBaseName().GetString();
             const AtParamEntry *paramEntry = AiNodeEntryLookUpParameter(nodeEntry, AtString(attrName.c_str()));
             if (paramEntry == nullptr) {
-                // Couldn't find this attribute in the osl entry
-                continue;
+                // If we failed to find the attribute, try without the shader prefix
+                // this is needed for non editable (BSDF/EDF/VDF) MaterialX node inputs
+                attrName = "param_" + attribute.GetBaseName().GetString();
+                paramEntry = AiNodeEntryLookUpParameter(nodeEntry, AtString(attrName.c_str()));
+                if (paramEntry == nullptr) {
+                    // Couldn't find this attribute in the osl entry
+                    continue;
+                }
             }
             uint8_t paramType = AiParamGetType(paramEntry);
```

See the attached repro [surface_model.zip](https://github.com/Autodesk/arnold-usd/files/10805801/surface_model.zip)

sebastienblor:
This PR checks for attributes prefixed with `param_` instead of `param_shader_` , which can happen with PBR nodes.
It's implemented both in the procedural and the delegate.

Added test_1438

Fixes #1438


cpichard (changes requested):
We noticed something weird on mac, investigating now

cpichard (approved):

PR #1462, opened 2023-03-24
Ensure shader scope doesn't appear twice in the hierarchy

### LINKED ISSUE #1457: Writer duplicates scopes in shader names
When we convert a .ass file to .usd, the surface shader hierarchy will show up first once above the material, and then a second time between the material and the shader. It should just be there once

sebastienblor:
When we write a shader primitive, we force it to be under the hierarchy of the material it's assigned to.
But because of that, when the whole USD scene is exported under a given scope, then we will find the scope twice under the shader.
This PR strips this scope before writing the usd prim, so that we get the expected scene hierarchy

Fixes #1457

### DIFF
--- a/translator/writer/prim_writer.cpp
+++ b/translator/writer/prim_writer.cpp
@@ -557,6 +557,15 @@ std::string UsdArnoldPrimWriter::GetArnoldNodeName(const AtNode* node, const Usd
         }
     }
 
+    // If we need to strip a hierarchy from the arnold node's name,
+    // we need to find if this node name starts with the expected hierarchy
+    // and do it before prefixing it with the scope
+    const std::string &stripHierarchy = writer.GetStripHierarchy();
+    if (!stripHierarchy.empty()) {
+        if (TfStringStartsWith(name, stripHierarchy)) {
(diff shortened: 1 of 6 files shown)


translator/writer/prim_writer.cpp
```diff
@@ -1157,8 +1157,18 @@ static void processMaterialBinding(AtNode* shader, AtNode* displacement, UsdPrim
 
     TfToken arnoldContext("arnold");
     if (shader) {
-        // write the surface shader under the material's scope
+        // Write the surface shader under the material's scope.
+        // Here we only want to consider the last name in the prim 
+        // hierarchy, so we're stripping the scope here
+        size_t npos = shaderName.rfind('/');
```
cpichard, line 1163: I am wondering if, instead of using the latest name component, we should just strip the scope that is going to be added later on ? this would allow to have scopes within the current material scope.
I haven't tested but I am thinking something like the following:
```
if (TfStringStartsWith(shaderName, materialName)) {
  std::string nodeLastName = shaderName.substr(materialName.size());
  ....
}
```

cpichard (approved):

PR #1493, opened 2023-04-06
USD-1492 | Add verbosity to render stats in the delegate 

TomMinor:
Added render resolution as an annotation
- Added elapsed time to render stats
- Added render resolution to render stats
- Estimated remaining render time now visible as a render annotation


render_delegate/render_param.cpp
```diff
@@ -217,4 +245,44 @@ void HdArnoldRenderParam::WriteDebugScene() const
...
+}
+
+void HdArnoldRenderParam::StartRenderMsgLog()
+{
+    _msgLogCallback = AiMsgRegisterCallback(MsgStatusCallback, AI_LOG_STATUS, this);
+}
+
+void HdArnoldRenderParam::StopRenderMsgLog()
+{
+    AiMsgDeregisterCallback(_msgLogCallback);
```
sebastienblor, line 264: we might just want to set _msgLogCallback to nullptr here, just to be sure we don't try to deregister the same callback multiple times.

TomMinor: Whoops, good spot

sebastienblor (approved):
LGTM

PR #1503, opened 2023-04-14
Ensure we don't map/unmap empty arrays

### LINKED ISSUE #1502: Render delegate crashes with empty arrays
Some scenes having empty primvar arrays end up crashing arnold when rendered in Hydra.

sebastienblor:
This PR ensures that we don't call AiArrayMap / AiArrayUnmap with empty arrays, which can happen sometimes when primvars are empty arrays.

Fixes #1502

### DIFF
--- a/render_delegate/basis_curves.cpp
+++ b/render_delegate/basis_curves.cpp
@@ -116,11 +116,13 @@ void HdArnoldBasisCurves::Sync(
         }
         const auto numVertexCounts = vertexCounts.size();
         auto* numPointsArray = AiArrayAllocate(numVertexCounts, 1, AI_TYPE_UINT);
-        auto* numPoints = static_cast<uint32_t*>(AiArrayMap(numPointsArray));
-        std::transform(vertexCounts.cbegin(), vertexCounts.cend(), numPoints, [](const int i) -> uint32_t {
-            return static_cast<uint32_t>(i);
-        });
-        AiArrayUnmap(numPointsArray);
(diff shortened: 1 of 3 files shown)


render_delegate/basis_curves.cpp
```diff
@@ -198,12 +200,14 @@ void HdArnoldBasisCurves::Sync(
                     } else if (desc.value.IsHolding<VtVec3fArray>()) {
                         const auto& v = desc.value.UncheckedGet<VtVec3fArray>();
                         auto* arr = AiArrayAllocate(v.size(), 1, AI_TYPE_VECTOR2);
-                        std::transform(
-                            v.begin(), v.end(), static_cast<GfVec2f*>(AiArrayMap(arr)),
-                            [](const GfVec3f& in) -> GfVec2f {
-                                return {in[0], in[1]};
-                            });
-                        AiArrayUnmap(arr);
+                        if (!v.empty()) {
```
cpichard, line 203: question ... would `if (AiArrayGetDataSize(arr)) {` also work ? in case the allocation failed ...

cpichard (dismissed):
lgtm, an additional note in the changelog would be fantastic

render_delegate/basis_curves.cpp
sebastienblor: maybe, but since we explicitely set 1 key, the issue should only happen if v.size is 0, so it should be equivalent

cpichard (approved):

PR #1510, opened 2023-04-25
Attempt to fix testsuite build issues

sebastienblor:
Simple change in the testsuite Sconscript, attempting to fix some issues in the CI


cpichard (approved):

sebastienblor:
> Hi @sebastienblor, should this change have a line in the changelog ?

I guess it's not needed here, I didn't even create an issue since it's really just an internal testsuite change, and the title wouldn't be very explicit neither as we're not sure what is causing it.

PR #1513, opened 2023-05-02
Remove deprecated ENABLE_MATERIALX build variable

### LINKED ISSUE #1512: Remove deprecated ENABLE_MATERIALX build variable
We should now be able to remove the build variable ENABLE_MATERIALX and only rely on the Arnold version

sebastienblor:
We no longer need the ENABLE_MATERIALX build variable, since we can just check for arnold versions being >= 7.1.4.0.

Fixes #1512

### DIFF
--- a/render_delegate/node_graph.cpp
+++ b/render_delegate/node_graph.cpp
@@ -650,11 +650,9 @@ AtNode* HdArnoldNodeGraph::ReadMaterialNode(const HdMaterialNode& node, const Co
     const auto* nodeTypeStr = node.identifier.GetText();
     bool isMaterialx = false;
     const AtString nodeType(strncmp(nodeTypeStr, "arnold:", 7) == 0 ? nodeTypeStr + 7 : nodeTypeStr);
-#ifdef ARNOLD_MATERIALX
     if (node.identifier != str::t_ND_standard_surface_surfaceshader && strncmp(nodeTypeStr, "ND_", 3) == 0) {
         isMaterialx = true;
     }
-#endif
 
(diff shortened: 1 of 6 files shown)


render_delegate/node_graph.cpp
```diff
@@ -814,11 +812,8 @@ HdArnoldNodeGraph::NodeDataPtr HdArnoldNodeGraph::GetNode(const SdfPath& path, c
 
 AtNode *HdArnoldNodeGraph::GetMaterialxShader(const AtString &nodeType, const AtString &nodeName, AtParamValueMap *params)
 {
-    // Disable materialx support until it's included in the USD libs
-    // that are shipped with Arnold
-#ifdef ARNOLD_MATERIALX
-    
-//#if ARNOLD_VERSION_NUM < 70103
+    // MaterialX support in USD was added in Arnold 7.1.4
+#if ARNOLD_VERSION_NUM < 70104
```
cpichard, line 816: should that be superior or equal to 70104 ?

sebastienblor: oups

sebastienblor:
I'm just realizing that there have been 2 steps, first in 7.1.4 the AiMaterialxGetOslCode function was introduced, and later in the next version we added an argument to this function. This is why we need 2 tests

render_delegate/node_graph.cpp
sebastienblor: fixed

cpichard (approved):

PR #1517, opened 2023-05-03
Add a "mtl" scope when authoring usd materials

### LINKED ISSUE #1486: Add a "Mtl" scope for materials when authoring usd files
We want to add a "Mtl" scope for materialx when we save an arnold file as USD. This is what is being done in other Autodesk tools, thus we want to port the same logic in arnold-usd

sebastienblor:
By default, now when we write usd files we add a "mtl" scope for materials (along with their shading trees). This can still be customized, by providing a specific flag "mtl_scope" to set that to a custom value (e.g. "/Looks" as it used to be in mayaUSD).

Some of the tests had to be updated to account for that change, I'm adding test_1486 that exercises the scope customization

Fixes #1486


translator/writer/writer.h
```diff
@@ -89,6 +89,20 @@ class UsdArnoldWriter {
...
+    const std::string &GetMtlScope() const {return _mtlScope;}
+    void SetMtlScope(const std::string &scope) {
+        _mtlScope = scope;
+        if (!_mtlScope.empty()) { 
+            // First character needs to be a slash
+            if (_mtlScope[0] != '/')
+                _mtlScope = std::string("/") + _mtlScope;
+            // Last character should *not* be slash, otherwise we could have 
+            // double slashes in the nodes names, which can crash usd
+            if (_mtlScope.back() == '/')
```
cpichard, line 102: it the name has multiple slashes in the end (hopefully it shouldn't happen, but it can), this will remove only one, should we try to remove all the slashes ?
I was thinking that we could may be use SdfPath instead, in the future, as they are efficient memory wise, and that would avoid us to write checking code ?

cpichard (approved):

PR #1518, opened 2023-05-04
First draft of arnold procedural using hydra code path

### LINKED ISSUE #25: Benchmark a Hydra/UsdImaging procedural against the procedural
Using Hydra and UsdImaging for a procedural is an alternative approach to directly accessing and traversing a Usd Stage. This is a similar approach to what the HdPrman plugin does and it offers significant code reuse between the Render Delegate and the Procedural. Besides this, Hydra also takes care of parallel access to the stage automatically.

Creating a procedural prototype should be straightforward, but the most important thing is to make sure performance is similar to a handwritten procedural

[HdPrman example](https://github.com/PixarAnimationStudios/USD/tree/702873ecea7fed0d0a43f25ba4c93a901602ff98/third_party/renderman-22/plugin)

cpichard:
This is a draft PR for early review and testing on multiple system/configurations

- PROCEDURAL_USE_HYDRA env enables rendering with hydra in kick
- reorganize the code into 2 main folders: libs for static libs and plugins for the compiled plugins.
- ENABLE_HYDRA_IN_USD_PROCEDURAL to enable the compilation of the render_delegate code in the arnold procedural
- remove hydra_test executable
- remove cmd 

Fixes #25


render_delegate/light.cpp
```diff
@@ -254,7 +254,7 @@ auto spotLightSync = [](AtNode* light, AtNode** filter, const AtNodeEntry* nentr
     const auto barndoorrightedge = getBarndoor(_tokens->barndoorrightedge);
     const auto barndoortop = getBarndoor(_tokens->barndoortop);
     const auto barndoortopedge = getBarndoor(_tokens->barndoortopedge);
-    auto createBarndoor = [&]() { *filter = AiNode(renderDelegate->GetUniverse(), str::barndoor); };
+    auto createBarndoor = [&]() { *filter = AiNode(renderDelegate->GetUniverse(), str::barndoor); }; // TODO procParent
```
sebastienblor, line 257: we need to set the proc parent here

render_delegate/render_delegate.cpp
```diff
@@ -1493,29 +1513,33 @@ void HdArnoldRenderDelegate::TrackDependencies(const SdfPath& source, const Path
...
-    AtNode* node;
-    while (_renderTagUntrackQueue.try_pop(node)) {
-        _renderTagMap.erase(node);
-    }
-    if (renderTags != _renderTags) {
-        _renderTags = renderTags;
-        for (auto& elem : _renderTagMap) {
-            const auto disabled = std::find(_renderTags.begin(), _renderTags.end(), elem.second) == _renderTags.end();
-            AiNodeSetDisabled(elem.first, disabled);
+    if (!IsBatchContext()) {
```
sebastienblor, line 1526: this should depend on the purpose and not on the batch context

sebastienblor (dismissed):
let's go

PR #1526, opened 2023-05-09
Set proper defaults in the options AA samples and depth

### LINKED ISSUE #1525: Set proper defaults when using the usd scene format plugin
When default render settings are used, some defaults (AA_samples, GI_diffuse_depth, GI_specular_depth) are different from husk.
This happens because the defaults in arnold core are different from the ones used in the Arnold plugins

sebastienblor:
Before reading the render settings prim, we set "default" values in the options, in a way that is consistent with the render delegate.
This will help to get similar results between husk and kick.
Adding test_1525

Fixes #1525


cpichard (commented):
would be cool to also have a line in the changelog

sebastienblor:
> would be cool to also have a line in the changelog

added

cpichard (approved):

PR #1533, opened 2023-05-17
Declare filename attributes as assets in the schemas

### LINKED ISSUE #1532: Schemas are not declaring asset parameters for filenames
We have changed filename attributes to be assets instead of strings, but this change wasn't ported to the schemas.
This causes problems with mayaUSD, where these schemas are loaded, and then exporting nodes with filenames don't write them properly, with an error saying that the attribute should be set as strings.

sebastienblor:
The filename attributes have been modified to be set as assets instead of strings. But this change had to be ported to the schemas

Fixes #1532


cpichard (approved):
looks good, may be a line in the changelog ?

sebastienblor:
done

PR #1551, opened 2023-06-02
Support UsdPrimvarReader_float2 correctly under a procedural

### LINKED ISSUE #1550: UsdPrimvarReader_float2 returning "st" not working in the usd procedural
In #1522 we have added support for UsdPrimvarReader_float2 shaders returning "st", which is working when we render a usd file.
But this is not working when the usd file is loaded through a procedural

sebastienblor:
The previous solution to support UsdPrimvarReader_float2 shaders returning "st" was not ideal (in #1522 ). We were first creating a user_data_rgb shader, translating it, and after the translation we were finding out that we needed another shader type. We were then renaming this shader with a dummy name and creating a utility instead with the expected name.

However, this doesn't seem to work when the nodes are created under a procedural. It seems like creating a procedural child node, then immediately renaming it isn't supported correctly. Because of that, the render would still rely on the user_data_rgb node and the previous fix wouldn't work.

Here instead, I'm first checking the "varname" attribute to see if I need to create a user_data_rgb or a utility shader. The reason I hadn't done this in the first place was that the existing functions to read attributes are all designed to rely on an existing AtNode (which I don't have in this context). I added a function that gets a VtValue and eventually resolves the recursive NodeGraph connections. 

Adding test_1550 with such a use case

Fixes #1550


libs/common/parameters_utils.cpp
```diff
@@ -294,6 +294,38 @@ void _ReadAttributeConnection(
...
+
+    if (attr.HasAuthoredConnections()) {
+        // This attribute has connections, we want to resolve NodeGraph connections
+        SdfPathVector targets;
+        attr.GetConnections(&targets);
+        if (!targets.empty()) {
+            SdfPath &target = targets[0]; // just consider the first target
+            SdfPath primPath = target.GetPrimPath();
+            UsdPrim targetPrim = attr.GetPrim().GetStage()->GetPrimAtPath(primPath);
+            if (targetPrim && targetPrim.IsA<UsdShadeNodeGraph>()) {
```
cpichard, line 311: I think it should be possible to use UsdShadeUtils, like `GetValueProducingAttributes` :https://openusd.org/release/api/class_usd_shade_utils.html#ad1f99058557167a9282c2531f2332bd6
This recursively goes into nodegraphs and connections and find the attribute with the value.

cpichard: and potentially remove this function

cpichard (approved):
lgtm

PR #1552, opened 2023-06-06
Retain attributes ordering in the ndr registry

sebastienblor:
The ndr registry was returning attributes in the alphabetic order, instead of keeping the attributes order defined in the arnold node entries.
This PR stores this ordering as a metadata for each node type, and then declares attributes in this same order.

We also skip attribute "name" which exists in all arnold nodes, but that doesn't make sense in USD where the name is not an attribute.


cpichard (approved):
looks good to me, a line in the changelog is missing though

PR #1568, opened 2023-06-14
Fix metallic attribute in UsdPreviewSurface in the render delegate 

### LINKED ISSUE #1567: Metallic attribute in UsdPreviewSurface is not supported in hydra
When rendering a UsdPreviewSurface is rendered in the render delegate, the attribute "metallic" is ignored

sebastienblor:
In the render delegate, when translating UsdPreviewSurface, there was a typo in the translation of the "metallic" attribute. Instead of converting usd-metallic into arnold-metalness, we were trying to do the opposite : looking for a usd attribute "metalness", to convert it to an arnold attribute "metallic". Because of that, metalness was always left to its default value of 0.

Adding test_1567 with a fully metallic sphere.

Fixes #1567

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ### Bug fixes
+- [usd#1567](https://github.com/Autodesk/arnold-usd/issues/1567) - Fix metallic attribute in UsdPreviewSurface in the render delegate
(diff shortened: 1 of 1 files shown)


cpichard (changes requested):
looks good ! changelog is missing though

sebastienblor:
> looks good ! changelog is missing though

done

cpichard (approved):

PR #1594, opened 2023-07-20
Add architecture to the compiler commands

TomMinor:
- Added basic coloured compiler commands for clearer multi arch building
- SHOW_CMDS=False will enable the shortened commands
- COLOR_CMDS can be used to set if they're colourful or not

Windows build:
image

This is handy when dealing with multiple archs:
image

Fixes #

**Additional context**
I've used ansi codes explicitly to avoid any additional dependencies (i.e. colorama), but let me know if this is a bad idea

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -50,6 +50,8 @@ vars.AddVariables(
     EnumVariable('COMPILER', 'Set compiler to use', ALLOWED_COMPILERS[0], allowed_values=ALLOWED_COMPILERS),
     PathVariable('SHCXX', 'C++ compiler used for generating shared-library objects', None),
     EnumVariable('CXX_STANDARD', 'C++ standard for gcc/clang.', '11', allowed_values=('11', '14', '17', '20')),
+    BoolVariable('SHOW_CMDS', 'Display the actual command lines used for building', False),
+    BoolVariable('COLOR_CMDS' , 'Display colored output messages when building', False),
     PathVariable('ARNOLD_PATH', 'Arnold installation root', os.getenv('ARNOLD_PATH', None), PathVariable.PathIsDir),
     PathVariable('ARNOLD_API_INCLUDES', 'Where to find Arnold API includes', os.path.join('$ARNOLD_PATH', 'include'), PathVariable.PathIsDir),
     PathVariable('ARNOLD_API_LIB', 'Where to find Arnold API static libraries', arnold_default_api_lib, PathVariable.PathIsDir),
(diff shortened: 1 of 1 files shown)


SConstruct
```diff
@@ -378,6 +380,34 @@ elif env['COMPILER'] == 'msvc':
         env.Append(CCFLAGS=Split('/Od /Zi /MD'))
         env.Append(LINKFLAGS=Split('/DEBUG'))
 
+if not env['SHOW_CMDS']:
+    # Hide long compile lines from the user
+    arch = env['MACOS_ARCH'] if IS_DARWIN else 'x86_64'
```
cpichard, line 385: should that be `env['ARCH']` instead of x86_64 ?

SConstruct
```diff
@@ -378,6 +380,34 @@ elif env['COMPILER'] == 'msvc':
...
+    env['LEXCOMSTR']    = 'Generating $TARGET ...'
+    env['YACCCOMSTR']   = 'Generating $TARGET ...'
+    env['RCCOMSTR']     = 'Generating $TARGET ...'
+    if env['COLOR_CMDS']:
+        ansi_bold = '\033[1m'
+        ansi_bold_green     = '\033[32;1m' + ansi_bold
+        ansi_bold_red       = '\033[31;1m' + ansi_bold
+        ansi_bold_yellow    = '\033[33;1m' + ansi_bold
+
+        env['CCCOMSTR']     = ansi_bold_green + env['CCCOMSTR']
```
cpichard, line 401: I can't see how the color is reset ? I was expecting something like the following:
```
env['CCCOMSTR'] = ansi_bold_green + env['CCCOMSTR'] + ansi_normal
``` 
otherwise the color remains for the following text, may be this is expected ?

SConstruct
```diff
@@ -378,6 +380,34 @@ elif env['COMPILER'] == 'msvc':
         env.Append(CCFLAGS=Split('/Od /Zi /MD'))
         env.Append(LINKFLAGS=Split('/DEBUG'))
 
+if not env['SHOW_CMDS']:
```
cpichard, line 383: I am not sure about the condition here, should that be `if env['SHOW_CMDS']` ?

SConstruct
```diff
@@ -378,6 +380,34 @@ elif env['COMPILER'] == 'msvc':
...
+    env['SHCCCOMSTR']   = 'Compiling {} $SOURCE ...'.format(arch)
+    env['CXXCOMSTR']    = 'Compiling {} $SOURCE ...'.format(arch)
+    env['SHCXXCOMSTR']  = 'Compiling {} $SOURCE ...'.format(arch)
+    env['LINKCOMSTR']   = 'Linking {} $TARGET ...'.format(arch)
+    env['SHLINKCOMSTR'] = 'Linking {} $TARGET ...'.format(arch)
+    env['LEXCOMSTR']    = 'Generating $TARGET ...'
+    env['YACCCOMSTR']   = 'Generating $TARGET ...'
+    env['RCCOMSTR']     = 'Generating $TARGET ...'
+    if env['COLOR_CMDS']:
+        ansi_bold = '\033[1m'
```
cpichard, line 396: I think we have colorama in tools/utils/colorama, may be it can be used ?

SConstruct
TomMinor: Nah, arnold-usd doesn't have an ARCH variable, just a MACOS_ARCH variable for Darwin (although it is driven by the ARCH variable from HtoA)

SConstruct
TomMinor: Good catch! I'll fix that

SConstruct
TomMinor: I copied the logic from arnold-core, where it's meant to be something like 'if SHOW_CMDS is True, just print the original verbose compile commands, otherwise show the minimal commands'

cpichard (approved):

PR #1628, opened 2023-08-21
Deprecate old typed schemas format

sebastienblor:
This PR removes the deprecated attribute format for typed schemas. They require an arnold: namespace, and we're no longer supporting formats without it.
Several test scenes were outdated and are fixed here. 
Also, the writer was still authoring attributes without this namespace for ArnoldGinstance primitives

### DIFF
--- a/testsuite/test_0132/data/test.usda
+++ b/testsuite/test_0132/data/test.usda
@@ -15,10 +15,8 @@ def ArnoldOptions "options"
         elementSize = 1
         interpolation = "constant"
     )
-    string arnold:procedural_searchpath = "C:/Users/blaines/Documents/maya/projects/default/"
     bool arnold:texture_automip = 0
     bool arnold:texture_per_file_stats = 1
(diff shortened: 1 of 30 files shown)


testsuite/test_0132/data/test.usda
```diff
@@ -2,38 +2,38 @@
...
-    bool texture_automip = 0
-    bool texture_per_file_stats = 1
-    string texture_searchpath = "C:/Users/blaines/Documents/maya/projects/default/sourceimages"
-    int threads = 40
-    int xres = 160
-    int yres = 120
+    string arnold:procedural_searchpath = "C:/Users/blaines/Documents/maya/projects/default/"
+    bool arnold:texture_automip = 0
+    bool arnold:texture_per_file_stats = 1
+    string arnold:texture_searchpath = "C:/Users/blaines/Documents/maya/projects/default/sourceimages"
```
cpichard, line 21: could be the opportunity to remove local path ?

cpichard (approved):

PR #1643, opened 2023-09-05
Ensure nodes are created an looked up in the right procedural scope

### LINKED ISSUE #1641: Set parent proc for all nodes created by the hydra procedural
Currently, in the hydra version of the translator, a lot of nodes are created without a parent procedural. This can cause lots of problems on the arnold side and should be fixed

sebastienblor:
When a procedural creates child nodes, it's extremely important that all of them are created with the right parent procedural, and that they're returned through the `procedural_get_node` callback. Not doing it can result with hangs or crashes.
So, as it's done in the translator, we should not call AiNode explicitely, we should rely on the factorized function ` CreateArnoldNodes` that will handle this properly.

Something that procedurals don't handle properly, is when a child node is created and then renamed. So we need to ensure nodes are given the expected name right away on creation. This forces to change the naming of some fallback nodes, which had their own pointer stored in the name. 

Also, we should never destroy any node when there is a procedural parent. This will be done when the procedural is deleted, and deleting these child nodes manually will likely cause crashes.

When we use `AiNodeLookUpByName` we always need to set the proper procedural scope, otherwise the nodes won't be found. 
These changes fix a dozen of tests. I cleaned up a few comments on the way. Updated one of the tests so that the shading tree is under the material's hierarchy

Fixes #1641


libs/render_delegate/node_graph.cpp
```diff
@@ -760,7 +760,14 @@ AtNode* HdArnoldNodeGraph::FindNode(const SdfPath& id) const
...
     const auto* pp = path.GetText();
-    if (pp == nullptr || pp[0] == '\0') {
+    // if the material path is already included in the shader path
+    // (which is supposed to be the case with shading trees under their respective materials)
+    // then we don't need to duplicated the material path, and we can just use the shader 
+    // path as-is
+    const std::string &idStr = GetId().GetString();
+    const std::string &pathStr = path.GetString();
+    
+    if (pp == nullptr || pp[0] == '\0' || pathStr.rfind(idStr.c_str(), 0) == 0) {
```
cpichard, line 770: if using pathStr, pp becomes redundant, they should be the same data

libs/render_delegate/light.cpp
```diff
@@ -526,10 +531,10 @@ void HdArnoldGenericLight::Sync(HdSceneDelegate* sceneDelegate, HdRenderParam* r
                 param->Interrupt();
                 interrupted = true;
                 const AtString oldName{AiNodeGetName(_light)};
-                AiNodeDestroy(_light); // TODO should we also remove it from the delegate list ???  !!!!
+                if (!_delegate->GetProceduralParent())
```
cpichard, line 534: I think it would be cool to have a function `_delegate->DestroyNode(node)` ,  a bit like `_delegate->CreateNode()`. That would allow to avoid repeating the pattern `if (!_delegate->GetProceduralParent()) {AiDestroy();}` and allow to add more logic later on in one place if needed.

cpichard (approved):

PR #1645, opened 2023-09-05
Support nodes mask in the hydra procedural

### LINKED ISSUE #1644: Support node mask in the hydra procedural
When we read usd through the reader, we can filter out certain nodes by setting a mask. We need the same logic in the hydra proc, in particular to avoid reading render settings, etc... through a procedural node

sebastienblor:
When the render delegate is used through a procedural, we need to ensure some nodes are filtered out (options, etc..).
This PR adds the node mask that is already used in the USD reader, and uses it to eventually skip certain nodes to be created by the render delegate. This fixes a few more tests in "hydra" mode

Fixes #1644


cpichard (approved):
Should we add a line in the changelog for this one ?
It might be good to have a generic entry like "Hydra procedural improvements - improve tests acceptance"

sebastienblor:
indeed, this one was worth a line in the changelog. Added

PR #1646, opened 2023-09-05
Add threshold to some tests for hydra

sebastienblor:
This PR adds a relatively small hardfail threshold for some tests that show almost the same result between hydra and usd


cpichard (approved):
I am wondering what is different in the generated ass scenes such that the renders are not the same ? did you check ?

sebastienblor:
the tessellated geometries (sphere, cube, etc...) are slightly different than in hydra. It's not visible, but there seems to be a small difference in the numbers

PR #1662, opened 2023-09-11
Support motion blur in hydra procedural

### LINKED ISSUE #1586: Hydra test_0018 : animation not translated in hydra procedural
Test_0018 renders an animated sphere, but the hydra version of the procedural renders a static sphere

sebastienblor:
When the usd procedural is used in hydra mode, and the camera lives in the parent .ass file, we need to create a dummy camera primitive with the right shutter settings, so that hydra can know in which motion range to sample attributes. 
It fixes a few tests in hydra mode, like test_0018

Fixes #1586


sebastienblor:
This almost fixed tests 23 and 24, we just needed to add a small threshold

cpichard (dismissed):

PR #1665, opened 2023-09-11
Support overrides in hydra procedural

### LINKED ISSUE #1585: Hydra test_0017 : overrides string ignored in the hydra procedural
test 17 loads a usd procedural, and sets a serialized string in the "overrides" attribute (adding a usd snippet that changes the current variant). It seems to be ignored in the hydra version of the procedural

sebastienblor:
This PR adds support for overrides when a usd file is read, going through the render delegate. This fixes test_0017 in hydra mode. Tests 31 and 32 needed to set the cube smoothing attribute as we did in other tests.
Right after these last PRs are merged, I'm planning to factorize the code used in both usd and hydra readers, as there's more and more stuff in common. 

Fixes #1585

**Additional context**

### DIFF
--- a/libs/render_delegate/reader.cpp
+++ b/libs/render_delegate/reader.cpp
@@ -28,6 +28,9 @@ PXR_NAMESPACE_USING_DIRECTIVE
 #include "api_adapter.h"
 #include "reader.h"
 
+static int s_anonymousOverrideCounter = 0;
+static AtMutex s_globalReaderMutex;
+
 // check pxr/imaging/hd/testenv/testHdRenderIndex.cpp
 
 class HydraArnoldAPI : public ArnoldAPIAdapter {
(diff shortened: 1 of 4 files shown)


libs/render_delegate/reader.cpp
```diff
@@ -132,10 +135,55 @@ void HydraArnoldReader::Read(const std::string &filename, AtArray *overrides,
     if (arnoldRenderDelegate == 0)
         return;
 
+    UsdStageRefPtr stage = nullptr;
+    SdfLayerRefPtr rootLayer = SdfLayer::FindOrOpen(filename);
+
     HydraArnoldAPI context(arnoldRenderDelegate);
-    UsdStageRefPtr stage = UsdStage::Open(filename, UsdStage::LoadAll);
-    // TODO check that we were able to load the stage
+    if (overrides == nullptr || AiArrayGetNumElements(overrides) == 0) {
```
cpichard, line 142: this code is very similar to the code opening the stage in UsdArnoldReader. 
HydraArnoldReader and UsdArnoldReader inherits from ProceduralReader, so what about mutualising this code in a function sounding like "ProceduralReader::OpenStageWithOverrides()"  that both reader could use ?

cpichard (approved):

PR #1666, opened 2023-09-11
Check the value of PROCEDURAL_USE_HYDRA instead of if it's defined

sebastienblor:
So far we were using the hydra procedural as soon as the variable PROCEDURAL_USE_HYDRA is defined. But it can be useful to switch this variable back and forth so in this PR I'm checking its value. If it's set to "0" we use the regular usd procedural, but otherwise we create the hydra reader

### DIFF
--- a/plugins/procedural/main.cpp
+++ b/plugins/procedural/main.cpp
@@ -42,14 +42,18 @@
 #define XARNOLDUSDSTRINGIZE(x) ARNOLDUSDSTRINGIZE(x)
 #define ARNOLDUSDSTRINGIZE(x) #x
 
+#include <iostream>
 inline ProceduralReader *CreateProceduralReader(AtUniverse *universe)
 {
 #ifdef ENABLE_HYDRA_IN_USD_PROCEDURAL
     if (ArchHasEnv("PROCEDURAL_USE_HYDRA")) {
-        return new HydraArnoldReader(universe);
-    } else {
-        return new UsdArnoldReader();
+        const std::string useHydra = ArchGetEnv("PROCEDURAL_USE_HYDRA");
+        int envVal = std::stoi(useHydra);
+        if (envVal != 0)
+            return new HydraArnoldReader(universe);
     }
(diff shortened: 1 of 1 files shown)


plugins/procedural/main.cpp
```diff
@@ -46,10 +46,13 @@ inline ProceduralReader *CreateProceduralReader(AtUniverse *universe)
 {
 #ifdef ENABLE_HYDRA_IN_USD_PROCEDURAL
     if (ArchHasEnv("PROCEDURAL_USE_HYDRA")) {
-        return new HydraArnoldReader(universe);
-    } else {
-        return new UsdArnoldReader();
+        const std::string useHydra = ArchGetEnv("PROCEDURAL_USE_HYDRA");
+        int envVal = std::stoi(useHydra);
```
cpichard, line 50: why not testing the string directly ? the code might raise an exception if PROCEDURAL_USE_HYDRA is not convertible to int

sebastienblor: in my first test after setting the variable in command line, the string was "0 " and not "0" as I was expecting. So I thought this would be better than stripping empty characters. But we can do that too

cpichard (approved):

plugins/procedural/main.cpp
```diff
@@ -46,10 +46,17 @@ inline ProceduralReader *CreateProceduralReader(AtUniverse *universe)
 {
 #ifdef ENABLE_HYDRA_IN_USD_PROCEDURAL
     if (ArchHasEnv("PROCEDURAL_USE_HYDRA")) {
-        return new HydraArnoldReader(universe);
-    } else {
-        return new UsdArnoldReader();
+        std::string useHydra = ArchGetEnv("PROCEDURAL_USE_HYDRA");
+        std::string::size_type i = useHydra.find(" ");
```
cpichard, line 50: I was thinking more of testing just the first character like

if (!useHydra.empty && useHydra[0] == '0') {
  // use usd reader
}

PR #1668, opened 2023-09-11
Refactoring to add support for cacheID in the hydra delegate

### LINKED ISSUE #1664: Support hydra procedural with a cacheID
This will be needed to render e.g. mayaUSD proxy shapes in hydra mode

sebastienblor:
This PR factorizes some of the code that was duplicated between the hydra and the usd reader, and moves it to the ProceduralReader class. Now the class that needs to be overridden is ReadStage that will work for usd filenames or a stage cache ID.

Fixes #1664


cpichard (approved):
cool ! should we add a comment in the changelog ? I am thinking that it could help to mention refactorings in the changelog, even if we don't copy them in the release note,  just that, later on, if there is a bug, we have a clearer understanding of what happen when looking for bugs. to be discussed ...

sebastienblor:
Added changelog

PR #1684, opened 2023-09-19
Fix test 739 along with a failing use case in the delegate

### LINKED ISSUE #1641: Set parent proc for all nodes created by the hydra procedural

sebastienblor:
This PR fixes test 739 which was missing some files. I changed a bit to base it from existing test 183, but with the ArnoldProceduralCustom schema.

In the refactoring done for #1641 , we ensured to stop calling AiNodeLookUpByName explicitely in the render delegate code, as this doesn't work properly in the hydra procedural (the universe and parent proc need to be specificied).

We left however, one explicit call to this function, as there seemed to be an equivalent way to handle it which could avoid passing along a render delegate pointer in lots of functions. It turns out that this workaround doesn't work correctly, which was revealed by this test 739. The parameter "arnold:nodes" which is a string of node names, didn't manage to resolve properly because we were not able to lookup the node in the right scope.  
In this PR I'm conforming with the rest of the code, to call instead renderDelegate->LookupNode, but unfortunately this requires to pass a delegate pointer in lots of functions. It does fix the issue, and allows test 739 to render properly in hydra

I also found out that byte array attributes (like instancer's "instance_visibility" attribute) were skipped, it was also causing this test to fail.
Finally, the ArnoldProceduralCustom schema loops over all the arnold attributes for this node. The render was erroring out because some of these parameters (which were not even set in the scene) couldn't be translated. In the context of USD with lots of possible attributes defined in each prim, this seems a bit too much. I turned them to warnings

Fixes #1641


libs/render_delegate/utils.cpp
```diff
@@ -46,9 +46,11 @@
 #include "pxr/imaging/hd/extComputationUtils.h"
 
 #include "debug_codes.h"
+#include "render_delegate.h"
 
 #include <type_traits>
 
+#include <iostream>
```
cpichard, line 53: do we need iostream ?

sebastienblor: no. Removed

cpichard (approved):
lgtm

PR #1687, opened 2023-09-20
Ensure testsuite logs are appended for resaved tests

sebastienblor:
Some tests are running several processes, like those that resave a scene and then kick the resaved scene. In this case, the log file is being overwritten over and over. This change ensures the log is appended so that we can get the full info.


cpichard (commented):
I am wondering why we don't delete the log before the run ? we delete everything else but not the log. Would that work ? then the open could always be in append mode ?

sebastienblor:
good point. Changed in the last commit

cpichard (approved):

PR #1702, opened 2023-10-02
Build usdGenSchema for the host architecture

ansono:
This PR will force usdGenSchema to be built for the host architecture, since it's a tool run during the build process.

### DIFF
--- a/tools/usdgenschema/SConscript
+++ b/tools/usdgenschema/SConscript
@@ -15,13 +15,10 @@ if system.IS_DARWIN:
     # the CCFLAGS and LINKFLAGS variables
     for flags in ['CCFLAGS', 'LINKFLAGS']:
         new_flags = []
-        arch_found = False
-        for flag in local_env[flags]:
+        flag_iter = iter(local_env[flags])
+        for flag in flag_iter:
             if flag == '-arch':
-                arch_found = True
-                continue
-            if arch_found:
-                arch_found = False
+                next(flag_iter) # skip next one
                 continue
             new_flags.append(flag)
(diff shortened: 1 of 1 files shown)


tools/usdgenschema/SConscript
```diff
@@ -8,6 +8,23 @@ import os.path
...
 
+if system.IS_DARWIN:
+    # This tool always has to be built for the host architecture. Let's do it in
+    # a super naive way, by removing any "-arch X" flag set in the parent
+    # construction environment. This can be found as two consecutives items in
+    # the CCFLAGS and LINKFLAGS variables
+    for flags in ['CCFLAGS', 'LINKFLAGS']:
+        new_flags = []
+        arch_found = False
+        for flag in local_env[flags]:
```
cpichard, line 19: would iterators work here ?
```
flag_iter = iter(local_env[flags])
for flag in flag_iter:
    if flag == "-arch":
        next(flag_iter) # skip next one
        continue
    new_flags.append(flag)
```

ansono: Not sure if that would make any difference? And the code is less readable.

sebastienblor (approved):

tools/usdgenschema/SConscript
cpichard: I am curious to know what makes you say it's less readable ? It would make no difference in terms of performance, but the code is more concise and readable to me. 
`arch_found` is used to skip the next element, why not calling it `skip_next` instead ?

cpichard: oh, just seeing your message now :)

PR #1706, opened 2023-10-04
Add test for schemas

sebastienblor:
This PR adds a test that will try to open the ascii files generated for schemas / ndr / usdImagingArnold, and will search for specific schemas names to be present.

I haven't been able to make this a generic test that will verify everything, so I picked a few harcoded schemas names, from different types. Hopefully, if something goes wrong in the schemas generation, this should likely fail


testsuite/test_0245/data/test.py
```diff
@@ -0,0 +1,76 @@
+import os
+import sys
+
+arnold_plugin = os.environ['ARNOLD_PLUGIN_PATH']
+
+def find_in_file(expectedTypes, filename):
+
+    file = open(filename, 'r')
```
cpichard, line 8: it's generally recommended to use the  `with open("filename", 'r') as f:` construct so the file is closed when the code goes out of the scope, it avoids having to manually close the file.

cpichard (dismissed):
awesome ! that will definitely help checking if the schema were correctly generated

testsuite/test_0245/data/test.py
sebastienblor: Thanks ! fixed

PR #1710, opened 2023-10-05
Properly check if attributes are authored / fixes test 216

### LINKED ISSUE #1709: Procedural failures if schemas are present
Some failures with test 216 revealed that some code in the procedural is not behaving properly if the arnold schemas are released along with the procedural.
We are using the function `UsdPrim::HasAttribute` to determine if an attribute is authored or not. However, once schemas are released, this function might return true even if the attribute isn't explicitely authored.

sebastienblor:
I checked all occurrences of `UsdPrim::HasAttribute` in the code base. We were miss-using it, when schemas are present, this function can return true even if the attribute isn't authored.

Since there doesn't seem to be a function `UsdPrim::HasAuthoredAttribute()` I added a small utility function in `parameter_utils.cpp` to get that information. 
Did the same with `arnold:driver` in the render settings code

I'm therefore updating test_0216 with the correct image (where the sphere is subdivided as expected)

Fixes #1709


libs/common/rendersettings_utils.cpp
```diff
@@ -475,7 +476,7 @@ void ReadRenderSettings(const UsdPrim &renderSettingsPrim, ArnoldAPIAdapter &con
             continue;
 
         AtNode *driver = nullptr;
-        if (productPrim.HasAttribute(TfToken("arnold:driver"))) {
+        if (HasAuthoredAttribute(productPrim, _tokens->aovDriver)) {
```
cpichard, line 479: but we don't add arnold:driver to the RenderProduct schema ? do we ?

libs/translator/reader/read_geometry.cpp
```diff
@@ -41,6 +41,7 @@
 #include <cstring>
 #include <string>
 #include <vector>
+#include <iostream>
```
cpichard, line 44: is iostream needed ?

libs/translator/reader/read_geometry.cpp
```diff
@@ -400,7 +401,7 @@ void UsdArnoldReadMesh::Read(const UsdPrim &prim, UsdArnoldReaderContext &contex
     // we get smoothed normals by default.
     // Also, we only read the builting subdivisionScheme if the arnold
     // attribute wasn't explcitely set above, through primvars:arnold (see #679)
-    if ((!prim.HasAttribute(str::t_primvars_arnold_subdiv_type)) &&
+    if ((!HasAuthoredAttribute(prim, str::t_primvars_arnold_subdiv_type)) &&
```
cpichard, line 404: is it where it wasn't working ? on meshes ?

libs/common/parameters_utils.cpp
```diff
@@ -723,3 +723,15 @@ void ReadPrimvars(
         ReadAttribute(prim, inputAttr, node, name.GetText(), attrTime, context, primvarType, arrayType);
     }
 }
+
+bool HasAuthoredAttribute(const UsdPrim &prim, const TfToken &attrName)
```
cpichard, line 727: looks good ! I would test the prim as well, just to be paranoid
```
if (!prim)
    return false;
```

cpichard (approved):
lgtm with a haiku in the changelog

libs/translator/reader/read_geometry.cpp
sebastienblor: yes this is what was causing test_0216 to produce the wrong results. This attribute "arnold:subdiv_type" isn't authored and we were supposed to enter the "if" condition and thus to set the proper subdivision attribute. But HasAttribute was always returning true since we introduced the schemas

libs/common/rendersettings_utils.cpp
sebastienblor: not at the moment. But since the test we want to do here is actually to know if the attribute is authored, I think it makes more sense to use the new explicit method. At least if we add it to the schemas in the future we won't have the problem

PR #1712, opened 2023-10-05
Discard primvars relative to arnold builtin attributes

### LINKED ISSUE #1711: Spurious user data added to arnold nodes
#1608  introduced a regression, as it's now causing all arnold builtin attributes to be also authored as user data.

sebastienblor:
In ReadPrimvars, we were checking if the primvar names was starting with "arnold" but we needed instead to check if it started with "primvars:arnold".
After the proposed changes, the useless user data disappears from the output arnold scenes

Fixes #1711


cpichard (dismissed):
Looks good to me, but it will probably change the behaviour with old scene ?  I think it needs a line in the changelog in case we need to find when it changed

sebastienblor:
Added a line in the changelog. It won't break old scenes, because this issue was introduced by #1608 in the 7.2.3.0 release. In previous versions, these attributes were already discarded.

PR #1714, opened 2023-10-05
Fix coding errors due to uv_remap attribute

### LINKED ISSUE #1713: Coding error "attempt to get string for GfVec4f"
In some scenes we're seeing error messages saying 
`Coding Error (secondary thread): in _FailGet at line 567 of S:\gocd\pipelines\kook\kook\usd\build\usd-22.11_windows-x86_64_static_vc-14.2_cxx14\pxr\base\vt\value.cpp -- Attempted to get value of type 'string' from VtValue holding 'GfVec4f'
`

sebastienblor:
We were getting errors because we tried to read uv_remap as a string attribute.  Before that we check for the value type, and if it's not a string we check if the attribute is connected.
This removes the coding error from a lot of test scenes

Fixes #1713


libs/translator/reader/utils.cpp
```diff
@@ -621,7 +621,17 @@ bool ReadNodeGraphAttr(const UsdPrim &prim, AtNode *node, const UsdAttribute &at
     VtValue value;
     if (attr && attr.Get(&value, time.frame)) {
         // RenderSettings have a string attribute, referencing a prim in the stage
-        std::string valStr = VtValueGetString(value, &attr);
+        std::string valStr;
+        // First check if the attribute is actually holding a "string" value
+        if (value.IsHolding<std::string>() || value.IsHolding<TfToken>() ||  value.IsHolding<SdfPath>())
+            valStr = VtValueGetString(value, &attr);
+        else if (attr.HasAuthoredConnections()) {
```
cpichard, line 628: should we test first for the connection ? and then the value ? as people tend to first put a value and then connection, and both are stored on the attribute, what do you think ?

sebastienblor: in theory these node graph attributes are just set as a string attributes. I could have not added the connection at all, I just added it as a fallback solution, in case we find an attribute of an unexpected type

cpichard (approved):
lgtm with a line in the changelog

cpichard (approved):

PR #1724, opened 2023-10-11
Fix testsuite path

ansono:


sebastienblor (approved):

ansono:
This is not working :(

ansono:
This works already

PR #1736, opened 2023-10-26
Add support for -j argument in husk #1077

### LINKED ISSUE #1077: Support --threads option in husk in the render delegate
When rendering via husk we'd like to pass the number of render threads through to the render delegate, ideally via hydra or potentially using an environment variable.

sebastienblor:
When we parse the command line arguments in the render delegate (which at the moment is specific to husk), we now look for an argument -j or --threads that is used to specify the amount of threads.

Unortunately this is hard to test, since the options attribute "threads" is an exception and is never saved to ass files

Fixes #1077

### DIFF
--- a/libs/render_delegate/render_delegate.cpp
+++ b/libs/render_delegate/render_delegate.cpp
@@ -651,7 +651,7 @@ void HdArnoldRenderDelegate::_SetRenderSetting(const TfToken& _key, const VtValu
             for (unsigned int i = 0; i < commandLine.size(); ++i) {
                 // husk argument for output image
                 if (commandLine[i] == "-o" && i < commandLine.size() - 2) {
-                    _outputOverride = commandLine[i+1];
+                    _outputOverride = commandLine[++i];
                     break;
                 }
                 // husk argument for thread count (#1077)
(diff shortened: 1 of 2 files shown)


libs/render_delegate/render_delegate.cpp
```diff
@@ -649,10 +649,18 @@ void HdArnoldRenderDelegate::_SetRenderSetting(const TfToken& _key, const VtValu
...
                     _outputOverride = commandLine[i+1];
                     break;
                 }
+                // husk argument for thread count (#1077)
+                if ((commandLine[i] == "-j" || commandLine[i] == "--threads") 
+                        && i < commandLine.size() - 2) {
+                    // if for some reason the argument value is not a number, atoi should return 0
+                    // which is also the default arnold value. 
+                    AiNodeSetInt(_options, str::threads, std::atoi(commandLine[i+1].c_str()));
+                }
```
cpichard, line 663: should we also increment `i` ?

cpichard (approved):

libs/render_delegate/render_delegate.cpp
sebastienblor: good point. Done, for both this argument as well as the -o argument above it

cpichard (approved):

cpichard (approved):

PR #1737, opened 2023-10-26
Fix skinning interpolation outside keyframe boundaries

### LINKED ISSUE #1735: Arnold/USD has wrong character position with Motion blur on
**Describe the bug**
Arnold kick display skinned character in bind pose instead of the first animation frame before the animation range
Arnold kick create a motion blur artifact at the last frame(s) of the animation range (blur with bind pose)

**To Reproduce**
Repro steps:
unzip and load in a USD stage HumanFemale.walk.usd [HumanFemale.zip](https://github.com/secure/attachment/4057213/4057213_HumanFemale.zip)
with motion blur on, render before and at the end of the animation range (see above) (motion blur on by default in Max, turn on in render setting for Maya) (extend time range to 200 in 3ds Max, anim start after 100)
Expected result: the viewport and rendering character position should match without motion blur artifact issue at any frames, when motion blur is on or off.

cpichard:
- Simplify the interpolation code of usdskel primitive in the considered time interval by removing the masking system. This allows to remove the function ShouldProcessAtTime.
- Make sure all the skinning geometry passed to arnold is correctly interpolated, before if the key wasn't found the code was falling back to the previous sample.
- Update test_1443 for future work in hydra.
- Move skin.usd from test_1331 in common/usd to reuse it.

**Notes**
This new method will change the way the motion blur of animated skinned object is rendered compared to the previous version.

Fixes #1735


sebastienblor (commented):
it's not compiling on windows in the CI.

sebastienblor (approved):

PR #1748, opened 2023-11-08
Support implicit conversions between Float3 and Float4 ARNOLD-14028

sebastienblor:
This PR adds implicit conversions between RGB and RGBA values, so that using one type instead of the expected one still manages to read the data. Done in both the procedural and the render delegate.
Added test_14028 to exercise this.

I'm setting this PR as a draft as it's not clear yet whether we want to accept these use cases, or if we want to keep being restrictive about these data types.


libs/render_delegate/utils.cpp
```diff
@@ -144,6 +144,21 @@ auto nodeSetVec2FromVec2d = [](AtNode* node, const AtString paramName, const GfV
...
 
+auto nodeSetRGBFromVec4 = [](AtNode* node, const AtString paramName, const GfVec4f& v) {
+    AiNodeSetRGB(node, paramName, v[0], v[1], v[2]);
+};
+auto nodeSetRGBFromVec4d = [](AtNode* node, const AtString paramName, const GfVec4d& v) {
+    AiNodeSetRGB(node, paramName, static_cast<float>(v[0]), static_cast<float>(v[1]), static_cast<float>(v[2]));
+};
+
+auto nodeSetRGBAFromVec3 = [](AtNode* node, const AtString paramName, const GfVec3f& v) {
+    AiNodeSetRGBA(node, paramName, v[0], v[1], v[2], 1.f);
```
cpichard, line 155: are we sure we always want 1.f ?

cpichard (dismissed):

libs/render_delegate/utils.cpp
cpichard: One test on linux is failing

sebastienblor: this is what the implicit conversion in Arnold does when you convert color3 to color4, alpha is set to 1

PR #1752, opened 2023-11-13
Fix sidedness inconsistencies between hydra and USD

### LINKED ISSUE #1746: Double check mesh sidedness
**Describe the bug**
There has been several reports of errors with sidedness, as by default, arnold has double sided meshes whereas usd has single sided meshes by default. We want to make sure this is working correctly and if so add some documentation explaining the difference.

sebastienblor:
This PR fixes the inconsistencies found regarding sidedness between the procedural and the delegate. 
All this comes from the issue that usd considers prims as being single sided by default, whereas arnold defaults to double sided (sidedness = 255)

So, for UsdGeomGprim primitives, we need to follow the USD default. But for Arnold typed schemas, which are a 1-1 mapping of Arnold nodes (and which don't inherit from UsdGeomGprim), then we need to follow arnold defaults.

Changes in this PR : 
- when the delegate reads an arnold native gprim, it doesn't look at the "doubleSided" parameter anymore, as it's not supposed to exist in this prim. It defaults to AI_RAY_ALL to be consistent with Arnold.
- When doubleSided is disabled, the procedural uses 0 as a sidedness value, whereas the delegate uses AI_RAY_SUBSURFACE. This PR sets it to 0 to be consistent with other arnold plugins, but it might want to set them to AI_RAY_SUBSURFACE anywhere instead. 
- When the USD translators reads cube / cone / sphere / capsule primitives, we were not reading the doubleSided attribute, which caused differences with hydra. This PR reads the value for all these nodes.
- The result for test_1718 has now changed. There is shadowing on only one of the 2 cubes, because one is a UsdGeomCube prim and the other is an ArnoldBox prim. So by default, the 2 are now rendering differently, which is expected.
- test_1718 can now be ran in hydra mode
- I noticed other tests (216, 1726, 1735) can now be added to the hydra list as well, as they seem to be succeeding. But we first need to check in the CI if it passes in all OSes

Fixes #1746


sebastienblor:
In order to avoid breaking existing Solaris scenes, we choose to keep the render delegate behaviour, where single sided meshes are still visible for SSS rays (otherwise they don't render correctly). The procedural now behaves identically

libs/translator/reader/read_geometry.cpp
```diff
@@ -668,6 +672,13 @@ void UsdArnoldReadCube::Read(const UsdPrim &prim, UsdArnoldReaderContext &contex
     for (GfVec3f& pt : points)
         pt = scale.Transform(pt);
 
+    VtValue sidednessValue;
```
cpichard, line 675: it looks like all the added code for setting the sidedness is the same for all the different type of prim, we could have a function `void SetSidedNess(UsdGeomGprim &)` ?

testsuite/test_1331/data/test.usda
```diff
@@ -122,6 +122,7 @@ def Xform "debug_mb_v3"
         float arnold:frame = 18
         float arnold:motion_end = 1
         float arnold:motion_start = -1
+
```
cpichard, line 125: newline is not necessary

cpichard (approved):

libs/translator/reader/read_geometry.cpp
sebastienblor: absolutely. Done

testsuite/test_1331/data/test.usda
sebastienblor: fixed

PR #1756, opened 2023-11-15
Registry should declare filenames as assets in GetTypeAsSdfType

### LINKED ISSUE #1755: Registry should declare filenames as assets in GetTypeAsSdfType
In the registry, filename attributes are declared as strings with the IsAssetIdentifier metadata. We need to ensure that [GetTypeAsSdfType](https://github.com/Autodesk/arnold-usd/blob/master/plugins/ndr/parser.cpp#L94) returns `SdfValueTypeNames->Asset`

sebastienblor:
Asset attributes of arnold shaders (e.g. image.filename) need to be declared as strings, as it was discussed [here ](https://github.com/Autodesk/arnold-usd/pull/1418#issuecomment-1420909766). However, the function `GetTypeAsSdfType` still needs to declare the attributes sdf type as being an asset. 

But by just changing this, I had other usd errors because the default value was a string, whereas the sdf type was an asset. Making the default value an empty SdfAssetPath worked, but then in Lookdevx the default filename showed as `@@` in the UI, which is confusing and doesn't match what other texture nodes look like. Since these asset attributes only exist for 2 attributes and in practice those filenames always default to empty values, the best solution seemed to declare an empty VtValue for default value.  In this case, everything works as expected

Fixes #1755

### DIFF
--- a/plugins/ndr/parser.cpp
+++ b/plugins/ndr/parser.cpp
@@ -93,6 +93,11 @@ public:
 #endif
     GetTypeAsSdfType() const override
     {
+        // Asset attributes are supposed to be declared as strings, but this function
+        // should still return an asset type name #1755
+        if (_typeName == SdfValueTypeNames->String && IsAssetIdentifier())
+            return {SdfValueTypeNames->Asset, SdfValueTypeNames->Asset.GetAsToken()};
+        
         return {_typeName, _typeName.GetAsToken()};
     }
 
@@ -123,8 +128,9 @@ void _ReadShaderAttribute(const UsdAttribute &attr, NdrPropertyUniquePtrVec &pro
         return;
     }
 
+    bool isAsset = false;
     VtValue v;
-    attr.Get(&v);
+    
     // The utility function takes care of the conversion and figuring out
     // parameter types, so we just have to blindly pass all required
     // parameters.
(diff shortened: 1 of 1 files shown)


plugins/ndr/parser.cpp
```diff
@@ -176,6 +186,13 @@ void _ReadShaderAttribute(const UsdAttribute &attr, NdrPropertyUniquePtrVec &pro
         hints.insert({TfToken(it.first), TfStringify(it.second)});
     }
 
+    // Asset attributes have to be treated differently as they need to be considered 
+    // as strings in some parts of USD, but as assets in others. Since in practice, 
+    // these attributes always default to empty strings, it's better not to set the
+    // VtValue at all, so that we don't get errors about invalid types.
+    if (!isAsset)
+        attr.Get(&v);
```
cpichard, line 194: I am wondering if the VtValue shouldn't be initialized before passed as defaultValue, because at this stage it doesn't have a defined type

sebastienblor: that's the problem. These attributes are designed in a strange way in USD, being strings in some way but assets in others. If I set it as an empty asset attribute, we have an undesirable behaviour in the lookdevx UI (there's a "@@" value by default) which doesn't happen with other shader libraries like materialx. If I set it as an empty string, I get error messages about a type mismatch. The only way to get the expected behaviour is to not set the value at all, and then we get the same behaviour as all the other shader filenames

cpichard (approved):

PR #1757, opened 2023-11-15
Declare closure attributes as terminals in the registry ARNOLD-13658

sebastienblor:
To be consistent with other shader libraries, we're declaring closure parameters (inputs and outputs) as "terminal" attributes for the attribute type, but the attribute Sdf Value needs to be "token".

### DIFF
--- a/plugins/ndr/parser.cpp
+++ b/plugins/ndr/parser.cpp
@@ -79,9 +79,9 @@ namespace {
 class ArnoldShaderProperty : public SdrShaderProperty {
 public:
     ArnoldShaderProperty(
-        const TfToken& name, const SdfValueTypeName& typeName, const VtValue& defaultValue, bool isOutput,
+        const TfToken& name, const SdfValueTypeName& typeName, const TfToken& typeToken, const VtValue& defaultValue, bool isOutput,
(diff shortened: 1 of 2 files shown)


plugins/ndr/parser.cpp
```diff
@@ -175,10 +177,16 @@ void _ReadShaderAttribute(const UsdAttribute &attr, NdrPropertyUniquePtrVec &pro
 
         hints.insert({TfToken(it.first), TfStringify(it.second)});
     }
+    // We're explicitely using token types for closures, to be consistent with other shader libraries.
+    // But the declared type must be "Terminal"
+    if (typeName == SdfValueTypeNames->Token) {
```
cpichard, line 182: what are the other shader libraries ? I am wondering if we shouldn't use "opaque" instead as token is for "interned string with fast comparison and hashing" which we might want to use for shaders ... what do you think ?
https://openusd.org/release/api/_usd__page__datatypes.html

cpichard: just seeing https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usdMtlx/utils.cpp#L539 
so I understand why now

cpichard (approved):

PR #1758, opened 2023-11-15
Return a zero value when an attribute type is not recognized

sebastienblor:
Crashes where reported in ARNOLD-14028, which happen when we try to get the value of an attribute which type isn't supported.
At the moment, after testing specific attribute types in VtValueGet* functions, we do a blind value.Get with the type that we want to convert it to.
But apparently, this can sometimes crash in USD when the types don't match. Now, we're instead returning a 0-value when the value doesn't hold any of the types that are supported

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - [usd#1735](https://github.com/Autodesk/arnold-usd/issues/1735) - Fix usdskel geometry and motion blur interpolation outside the keyframe boundaries.
 - [usd#1524](https://github.com/Autodesk/arnold-usd/issues/1524) - Fix material binding on instances under a SkelRoot
 - [usd#1718](https://github.com/Autodesk/arnold-usd/issues/1718) - Support primvars:arnold attributes in Arnold typed schemas
+- [usd#1758](https://github.com/Autodesk/arnold-usd/issues/1758) - Return a zero value when an attribute type is not recognized
 
 ## [7.2.4.1] - 2023-10-18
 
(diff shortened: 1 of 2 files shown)


libs/translator/reader/utils.h
```diff
@@ -425,7 +425,7 @@ static inline bool VtValueGetBool(const VtValue& value)
         VtArray<long> array = value.UncheckedGet<VtArray<long>>();
         return array.empty() ? false : (array[0] != 0);   
     }
-    return value.Get<bool>();
+    return false;
```
cpichard, line 428: something that would be great is to pass the default value as an argument:
```
VtValueGetBool(const VtValue &value, bool default=false) 
{ 
   ...
    return default;
}
```
This has the benefit of exposing the fact that the function returns a default if the value is not found ... and later on be able to change the default, as sometimes we might want true and other times false. Same for the other similar functions. 
What do you think ?

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - [usd#1077](https://github.com/Autodesk/arnold-usd/issues/1077) - Support --threads / -j argument in husk to control the amount of render threads
 - [usd#658](https://github.com/Autodesk/arnold-usd/issues/658) - Support pixel aspect ratio in Hydra
 - [usd#1746](https://github.com/Autodesk/arnold-usd/issues/1746) - Made the behaviour for doubleSided gprims consistent between USD and Hydra
+- [usd#1758](https://github.com/Autodesk/arnold-usd/issues/1758) - Return a zero value when an attribute type is not recognized
```
cpichard, line 12: now it is returning a default value not necessarily zero, so may be this could be updated ?

cpichard (dismissed):

PR #1774, opened 2023-12-01
Use VtValueGet functions from the render delegate

### LINKED ISSUE #1767: Factorize VtValueGet functions
We would want to move the VtValueGet*  functions to libs/common, so that it can be shared between the procedural and the render delegate

sebastienblor:
This PR removes the delegate's specific conversion functions, and uses instead the VtValueGet* functions.
It's a first step to factorize even more both code paths (e.g. the whole HdArnoldSetParameter function)

Fixes #1767

### DIFF
--- a/libs/common/parameters_utils.h
+++ b/libs/common/parameters_utils.h
@@ -148,238 +148,112 @@ void _ReadAttributeConnection(
 bool HasAuthoredAttribute(const UsdPrim &prim, const TfToken &attrName);bool HasAuthoredAttribute(const UsdPrim &prim, const TfToken &attrName);
 
 
-static inline bool VtValueGetBool(const VtValue& value, bool defaultValue = false)
+template <typename CastTo, typename CastFrom>
+inline bool _VtValueGet(const VtValue& value, CastTo& data)
+{    
+    using CastFromType = typename std::remove_cv<typename std::remove_reference<CastFrom>::type>::type;
+    if (value.IsHolding<CastFromType>()) {
+        data = static_cast<CastTo>(value.UncheckedGet<CastFromType>());
+        return true;
+    } else if (value.IsHolding<VtArray<CastFromType>>()) {
+        const auto& arr = value.UncheckedGet<VtArray<CastFromType>>();
+        if (!arr.empty()) {
+            data = static_cast<CastTo>(arr[0]);
+            return true;
+        }
+    }
+    return false;
+}
+
+template <typename CastTo>
+inline bool _VtValueGetRecursive(const VtValue& value, CastTo& data)
 {
-    if (value.IsHolding<bool>())
-        return value.UncheckedGet<bool>();
-    if (value.IsHolding<int>())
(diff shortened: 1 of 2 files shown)


libs/common/parameters_utils.h
```diff
@@ -164,6 +166,10 @@ static inline bool VtValueGetBool(const VtValue& value, bool defaultValue = fals
         VtArray<int> array = value.UncheckedGet<VtArray<int>>();
         return array.empty() ? false : (array[0] != 0);
     }
+    if (value.IsHolding<VtArray<unsigned int>>()) {
+        VtArray<unsigned int> array = value.UncheckedGet<VtArray<unsigned int>>();
+        return array.empty() ? false : (array[0] != 0);
```
cpichard, line 171: should it return the defaultValue instead of false ? the other VtValueGet functions might need to be updated to return the defaultValue

libs/common/parameters_utils.h
```diff
@@ -164,6 +166,10 @@ static inline bool VtValueGetBool(const VtValue& value, bool defaultValue = fals
         VtArray<int> array = value.UncheckedGet<VtArray<int>>();
         return array.empty() ? false : (array[0] != 0);
     }
+    if (value.IsHolding<VtArray<unsigned int>>()) {
+        VtArray<unsigned int> array = value.UncheckedGet<VtArray<unsigned int>>();
```
cpichard, line 170: I think we can use the const reference to the object instead of a copy, even if I am guessing the underlying array buffer won't be copied anyway.

libs/common/parameters_utils.h
```diff
@@ -154,6 +154,8 @@ static inline bool VtValueGetBool(const VtValue& value, bool defaultValue = fals
         return value.UncheckedGet<bool>();
     if (value.IsHolding<int>())
         return value.UncheckedGet<int>() != 0;
+    if (value.IsHolding<unsigned int>())
```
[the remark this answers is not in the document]
sebastienblor: yes, the only problem I saw is that the docs seem to say that this is not thread-safe if the same VtValue is used in parallel. I'm not sure if this could happen on our side

libs/common/parameters_utils.h
```diff
@@ -148,231 +148,112 @@ void _ReadAttributeConnection(
...
+    return _VtValueGet<CastTo, CastTo>(value, data) || 
+           _VtValueGetRecursive<CastTo, CastFrom, CastRemaining...>(value, data);
+}
 
-    if (value.IsHolding<double>())
-        return static_cast<float>(value.UncheckedGet<double>());
+static inline bool VtValueGetBool(const VtValue& value, bool defaultValue = false)
+{   
+    if (!value.IsEmpty())
+        VtValueGet<bool, int, unsigned int, long>(value, defaultValue);
```
cpichard, line 190: do we want to also convert unsigned char, char, unsigned long as well ?

cpichard (approved):

PR #1780, opened 2023-12-08
Expose API function writing an arnold universe to an existing usd stage

### LINKED ISSUE #1779: Expose API to write an arnold scene to a given usd stage
MtoA and MAXtoA are compiling a small function that allows to get a UsdStage from a cache id, and then write arnold data on it. 
This allows to add arnold nodes and attributes on top of a UsdStage exported by MayaUSD / MaxUSD.
We want to move this functions in arnold-usd, so that all the usd-related code can be compiled by users having a custom  USD.

sebastienblor:
This PR moves a function that was previously in MtoA and MAXtoA. These plugins look for an explicit symbol in a dso that is compiled against the USD libs from MayaUSD / MaxUSD. 
The function just takes an integer as an argument, looks for the usdStage in a usdStageCache and writes to this specific usdStage (as opposed to the scene format API that writes to a usd file on disk

Fixes #1779

### DIFF
--- a/plugins/procedural/main.cpp
+++ b/plugins/procedural/main.cpp
@@ -33,9 +33,15 @@
 #include <pxr/base/tf/pathUtils.h>
 #include <pxr/base/arch/env.h>
 #include "procedural_reader.h"
+#include <pxr/usd/usd/stageCache.h>
+#include <pxr/usd/usdUtils/stageCache.h>
+#include "writer.h"
 
 #if defined(_DARWIN) || defined(_LINUX)
 #include <dlfcn.h>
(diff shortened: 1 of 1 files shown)


cpichard (approved):
Would there be a way to test this function ? just to make sure the exports works correctly ?

sebastienblor:
> Would there be a way to test this function ? just to make sure the exports works correctly ?

The MtoA testsuite is already testing this. On the arnold-usd side, I'm not sure how we can test it as it only makes sense when it's linked against a dynamic usd (since it looks for a usdStage in a usdStageCache)

PR #1782, opened 2023-12-11
Fix RectLight texture uvs

### LINKED ISSUE #1772: Rect light have inverted map uvs in husk and kick
**Describe the bug**
When using an area light, a RectLight with a texture, husk and kick will treat the uvs differently, they look inverted.
Attached a scene to show the issue, just render it with kick and husk.
[rectlight_invertedUVs.zip](https://github.com/Autodesk/arnold-usd/files/13498712/rectlight_invertedUVs.zip)

cpichard:
- We use the same RectLight  vertices in the render delegate and the procedural. We flip the X axis as it should be inverted as per the USD doc : min coordinates should be at (+X, +Y) and max coordinates at (-X, -Y). The Y axis is already inverted.

Fixes #1772


libs/render_delegate/light.cpp
```diff
@@ -709,6 +709,9 @@ void HdArnoldGenericLight::SetupTexture(const VtValue& value)
     } else { // Connect to color if filename doesn't exists.
         AiNodeLink(_texture, str::color, _light);
     }
+    if (AiNodeEntryGetNameAtString(nentry) == str::quad_light) {
+        AiNodeSetBool(_texture, AtString("sflip"), true);
```
sebastienblor, line 713: let's add `sflip` to constant_strings.h, so that this call is cheaper

PR #1790, opened 2024-01-03
Factorize shader translation between usd and hydra

### LINKED ISSUE #1783: Share shader export code between USD and Hydra
We'd want the whole logic of shader conversion to be shared between the procedural and the render delegate.

sebastienblor:
This PR moves the shader translation logic to a new file materials_utils.cpp in the common folder, and makes it so that both the usd reader and the render delegate use it. All the logic of what to do for each shader is in this file, whether it's for arnold specific shaders, UsdPreviewSurface shaders, materialX, etc...

I made it so that InputAttribute would be a simple class that nobody derives from, and that just has a VtValue and a connection. This allows to allocate a bunch of InputAttributes at once in a list / map, without needing them to be from derived classes with different sizes. The attribute values are now flattened before ReadAttribute is even called. I added a function CreateInputAttribute that builds an inputAttribute from a UsdAttribute, with an eventual customization for primvars (which is now moved to the translator utils)

In the hydra node_graph class, all the previous logic was removed and the code is now relying on the common materialReader.
When a material network is being read, we check the list of "terminals"  (displacement, surface, etc...) and translate each of them.
We get the list of nodes to translate, and for each of them we build a map of "input attributes" (with a VtValue and a connection). This is a bit more tricky than in the procedural as the list of connections in hydra (aka relationships) is stored in a big list for each shading network.
Then for each shader to translate we call a common function ReadShader with list of attributes to translate.
As a consequence of this shared code, the node naming (with `primvars:arnold:name`) is now supported in the delegate as it is in the procedural

For the USD reader, in a first step I wanted to couple this change with a refactoring of how materials were being read, and make it so that all shading networks under a NodeGraph/Material would be read from a single thread, and on demand (i.e. unconnected shaders would be ignored).
I thought it would be needed, but this ended up being orthogonal to the refactoring. Since we might still want to do this change I kept it here inside of a ifdef, so that we can eventually keep working on it (there are still some issues to fix).
For that change, I also needed all the primReaders to return a AtNode from their Read function, that returns the "root" node they created. 

I added a few more tests for use cases I had to debug, and some existing tests are now magically working in hydra mode so they were added to the hydra group.

Fixes #1783


CHANGELOG.md
```diff
@@ -16,6 +16,9 @@
 - [usd#1784](https://github.com/Autodesk/arnold-usd/issues/1784) - The aov layer name is now correctly taken into account when rendering exrs with husk and using the arnold productType.
 
 
+### Build
```
cpichard, line 19: shouldn't it be in the Features instead ?

libs/common/parameters_utils.h
```diff
@@ -22,139 +22,31 @@
...
-    bool Get(VtValue *value, double time) override
-    {
-        if (!_attr || !value)
-            return false;
-        
-        return _attr.Get(value, time);
-    }
-    bool IsAnimated() const override {return _attr.ValueMightBeTimeVarying();}
-    void ValidatePrimPath(std::string &path) override;
+    ~InputAttribute() {delete timeValues;}
```
cpichard, line 27: we should check if timeValues is not nullptr before deleting it, specially as the default constructor doesn't allocate it

libs/render_delegate/node_graph.cpp
```diff
@@ -43,454 +43,153 @@
...
-        if (remapIt == _NodeRemapFuncs().end()) {
-            continue;
-        }
-
-        HydraMaterialNetworkEditContext editCtx(network, material);
-        remapIt->second(&editCtx);
+    // Ensure all AtNodes created for this node graph are properly deleted
+    for (const auto& node : _nodes) {
+        if (node.second)
+            AiNodeDestroy(node.second);
```
cpichard, line 122: it would be great to make sure _nodes doesn't contain invalid data, would it make sense to clear it ? or assign nullptr to each node ? (node.second = nullptr)

sebastienblor: but this is in the class destructor, so the whole vector will be gone after that

libs/common/parameters_utils.h
sebastienblor: `delete nullptr`  is supposed to be safe, isn't it ?

CHANGELOG.md
sebastienblor: changed

libs/common/parameters_utils.h
cpichard: it is, and timeValues is correctly initialised, my bad.

libs/render_delegate/node_graph.cpp
cpichard: didn't see it was in the destructor, sloppy review ...

cpichard:
Tested in interactif mode in htoa on different scenes, I didn't encounter any issues.

cpichard (approved):

cpichard (approved):

cpichard (approved):

PR #1802, opened 2024-01-11
Replace ComputePointsAtTime with a manual extrapolation of positions

### LINKED ISSUE #1801: Fix velocities computation with USD 23.11
In order to compute motion blurred positions based on velocities and accelerations, we call `UsdGeomPointBased::ComputePointsAtTime`. But on windows with USD 23.11, this seems to be giving us wrong results, and therefore several tests are failing.

sebastienblor:
Instead of calling `UsdGeomPointBased::ComputePointsAtTime`, we are now extrapolating the positions manually based on velocities, and optionally accelerations. This is fixing the testsuite with USD 23.11

Fixes #1801


cpichard (approved):
should we add a line in the changelog ?

sebastienblor:
> should we add a line in the changelog ?

right, done

cpichard (approved):

sebastienblor:
The problem actually came from an internal USD build configuration. We can close this PR

PR #1813, opened 2024-01-16
Improve material network creation times by caching the nodes returned by Arnold

### LINKED ISSUE #1812: Material networks creation in 7.2.4 is slower than before
**Describe the bug**
Since version 7.2.4 the creation of material networks is slower than it was in 7.2.3, the time can be multiplied by an order of magnitude. This happens only in the hydra render delegate.

**To Reproduce**
1. Install 7.2.3.x and 7.2.4.x 
2. Use a profiler and render in a hydra enabled application both version

cpichard:
- The functions AiMaterialXGetNodeEntryFromDefinition and AiMaterialXGetOslShaderCode are costly and we called them for every shader prim. The easiest solution is to cache the returned node entry associated to a particular shader type. We do the same for the osl code.

Fixes #1812


sebastienblor (commented):
I'm afraid this is not thread-safe. This function can be called simultaneously by different threads to translate different shading trees

cpichard:
even worse, it seems that some cached memory becomes invalid when the render delegate is destroyed and recreated. We can't reliably use static variables.

sebastienblor (approved):

PR #1816, opened 2024-01-17
Restore previous testsuite behaviour with abort on license fail

sebastienblor:
abort_on_license_fail now defaults to true. This PR sets it back to the previous behaviour during the testsuite.


cpichard:
do we need this change if we use the environment variable ARNOLD_FORCE_ABORT_ON_LICENSE_FAIL ?

PR #1819, opened 2024-01-18
Cherry pick + refactor fix for #1812 from branch fix-7.2.5

### LINKED ISSUE #1812: Material networks creation in 7.2.4 is slower than before

cpichard:
- Add the fix related to issue #1812 coming from the branch fix-7.2.5. This is the fix to optimise/limit the calls to the AiMaterialx functions

Fixes #1812

### DIFF
--- a/libs/common/api_adapter.h
+++ b/libs/common/api_adapter.h
@@ -8,6 +8,8 @@
 #include <pxr/usd/usdGeom/primvar.h>
 #include "constant_strings.h"
 
+#include <ai.h>
+
 PXR_NAMESPACE_OPEN_SCOPE
 
 // This is a base class used to call Arnold API functions within a particular context.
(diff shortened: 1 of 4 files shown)


sebastienblor (approved):

libs/common/api_adapter.h
```diff
@@ -136,8 +138,43 @@ class ArnoldAPIAdapter {
...
+        return shaderNodeEntryIt->second;
+    };
+
+    AtString GetCachedOslCode(const std::string &oslCodeKey, const char *nodeDefinition, AtParamValueMap *params) {
+        std::lock_guard<AtMutex> lock(_oslCodeCacheMutex);
+        const auto oslCodeIt = _oslCodeCache.find(oslCodeKey);
+        if (oslCodeIt == _oslCodeCache.end()) {
+    #if ARNOLD_VERSION_NUM > 70104
+            _oslCodeCache[oslCodeKey] = AiMaterialxGetOslShaderCode(nodeDefinition, "shader", params);
+    #elif ARNOLD_VERSION_NUM >= 70104
```
BigRoy, line 161: Is this the correct version number being compared against? It looks odd that it's comparing to the same value - basically this else statement says after the `if` says `ARNOLD_VERSION_NUM == 70104` right?

It also means that if the version would be `< 70104` this `if` statement body would effectively be empty?

sebastienblor: Yes, before Arnold 7.1.4 this API didn't exist and there was no way to get any osl code, so it won't render this shader.
In 7.1.4 the API was first introduced, and in following versions an argument was added. So that's why we have the current ifdefs. Indeed it would have behaved the same way if it had been written `elif ARNOLD_VERSION_NUM == 70104`

BigRoy: Thanks, makes sense. 👍

Issue #1835, opened 2024-01-29
Update "help" metadata in Sdr


sebastienblor:
We're currently using the arnold "desc" metadata to provide USD registry a "Help" value.
However, this is changing in the arnold side, as the metadata is now called "help", so we need to update the ndr parser code

PR #1836, opened 2024-01-29
Rename "desc" Arnold metadata as "help" #1835

sebastienblor:
The previous arnold metadata "desc" is being renamed as "help". This PR updates the ndr registry code, to look for the correct metadata name


cpichard (dismissed):
Should we add the changelog now or in a separate dedicated PR?

cpichard (approved):

PR #1855, opened 2024-02-13
Fix slowness in the render delegate due to scene update when the camera moves

### LINKED ISSUE #1854: Moving the camera triggers an update of the scene in the render delegate.
**Describe the bug**
When we move the camera in the render delegate, it looks like Arnold is updating the scene while it should be only re-rendering. On big scene this could be very slow.
This happens since this PR: https://github.com/Autodesk/arnold-usd/pull/1692

cpichard:
- We are testing if the camera is a perspective or ortho camera in the render delegate camera sync. Unfortunately the test was incorrect as the type returned by the function AiNodeEntryGetTypeName is the generic type "camera" and not the specific type of the camera persp_camera or ortho_camera. A new arnold camera was created everytime there was a change in parameter, thus forcing an update of the scene.

Fixes #1854

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,12 @@
 
 ## [fix-7.2.5]
 
+### Bug fixes
+- [usd#1854](https://github.com/Autodesk/arnold-usd/issues/1854) - Fix unnecessary updating of the scene in the render delegate when the camera is moving. This improves the interactivity in Solaris.
+
 ## [7.2.5.2]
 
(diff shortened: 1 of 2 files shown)


BigRoy:
Too bad this one didn't make it into the new Arnold-USD release. Hoping to see this soon in HtoA! 🙉 🙈 🙊

sebastienblor (dismissed):

sebastienblor (approved):

PR #1862, opened 2024-02-16
Enable MSVC_USE_SCRIPT Scons option to override default MSVC_VERSION detection

ansono:
This PR enables a built-in option in SCons ([`MSVC_USE_SCRIPT`](https://scons.org/doc/production/HTML/scons-man.html#cv-MSVC_USE_SCRIPT)) in order to override the default detection from [`MSVC_VERSION`](https://scons.org/doc/production/HTML/scons-man.html#cv-MSVC_VERSION)

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -139,6 +139,10 @@ vars.AddVariables(
 
 if IS_WINDOWS:
     vars.Add(('MSVC_VERSION', 'Version of MS Visual C++ Compiler to use', '14.2'))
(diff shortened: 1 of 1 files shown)


ansono:
> do we need something equivalent for cmake builds ?

I'm not sure yet if cmake can do this fine grained selection through [`MSVC_TOOLSET_VERSION`](https://cmake.org/cmake/help/latest/variable/MSVC_TOOLSET_VERSION.html). According to that doc page, it suffer from the same "issue" as SCons ... only 14.3 can be specified.

PR #1863, opened 2024-02-16
Fix crash in the procedural cleanup

sebastienblor:
It appears that the user_ptr can be null in `procedural_cleanup`, meaning that we didn't go through `procedural_init`. This is causing crashes in the plugins integration. 
We just need to test against null pointers, as it's already done in other procedural functions


cpichard:
looks good ! just need a line in the changelog

PR #1866, opened 2024-02-19
Add usdImaging adapter for ArnoldOptions 

### LINKED ISSUE #1865: Add usdImaging Adapter for ArnoldOptions
For usd scenes that were authored with ArnoldOptions primitives, we want to ensure they can be loaded by the Hydra render delegate

sebastienblor:
Adding a usdImaging adapter for the ArnoldOptions primitive (which is currently being authored by the Arnold USD writer), as well as a HdArnoldOption translator to read it properly.
This allowed to add 2 tests to the hydra group, that are now exercising it.

Fixes #1865

**Additional context**


sebastienblor:
Note that with the changes done in #1867 , the current change is not strictly needed as new scenes would have RenderSettings primitives instead of ArnoldOptions. The goal here is to eventually be able to render with Hydra existing scenes that had ArnoldOptions primitives.
If we choose not to implement this PR, we'll need to update the existing tests so that they all use RenderSettings

cpichard (dismissed):

PR #1867, opened 2024-02-21
Write Arnold options as UsdRenderSettings primitives

### LINKED ISSUE #1852: Write arnold options as UsdRenderSettings prim
We are currently authoring arnold options as a ArnoldOptions primitive in USD. We want instead to write it as UsdRenderSettings (along with render products and render vars), so that it relies on USD builtin primitive types

sebastienblor:
In this PR we change the writer to author arnold options as UsdRenderSettings primitives instead of ArnoldOptions.
Similarly, drivers get authored as UsdRenderProducts and AOVs as UsdRenderVars. Each renderVar has the information about its filter, therefore arnold filter nodes are no longer authored to USD.

- The RenderSettings prim is saved as /Render/settings
- RenderProducts are authored under /Render/Products/ with the name of the arnold driver node
- RenderVars are authored under /Render/Vars/ with the AOV name. If the same AOVs shows up several times in the outputs list, we append an index suffix for the primitive name.
-  /Render , /Render/Products, and /Render/Vars are authored as Scopes. Primitives inside of it do not obey the export scope (which is meant to author primitives under a given hierarchy, this isn't supposed to affect /Render).

While implementing that and looking at the testsuite, I found several issues to be fixed :
- We were missing camera support in AOVs (that renders a different camera for a given AOV)
- When rendering a usd file in hydra mode, we were sometimes rendering the wrong camera and the wrong shutter. We were asking for AiUniverseGetCamera which was empty since the camera hadn't been translated to Arnold yet. In the Hydra reader (when we kick usd files in hydra mode), ReadRenderSettings now passes the render camera SdfPath so that we can properly tell hydra about it.  
- Arnold attributes in the RenderSettings can either have arnold as a namespace or arnold:global (as this is what Solaris authors). The usd reader was supporting both, but hydra was only supporting arnold:global. I'm now adding support for both
- When we write arnold attributes, we want to first check if they're in the list of already exported attributes before handling the "name" attribute. This way we can eventually prevent from authoring the name attribute 
- We had crashes in some scenes having indexed user data. We were having crashes when trying the reset the "idxs" user data, which is not something we're supposed to do in Arnold. I'm now checking if we're in this use case

With this change, about 40 new tests are now passing in hydra mode. Some of them just needed a small threshold so I added it here. Some tests needed to be adjusted for the new driver names.

Fixes #1852


cpichard (approved):
Awesome !

PR #1879, opened 2024-03-05
Make arnold relative path optional in image nodes

### LINKED ISSUE #1878: Make arnold relative path optional in image nodes
Depending on how a usd file was authored, we can want the image filename relative paths to be considered the "Arnold way" (i.e. relative to the arnold texture_search_path), or the "USD way" (relative to the layer). 

This is important when relative paths have UDIMs, as we're currently trying to solve this only for UvTexture shaders, but not for arnold image shaders.

The Arnold relative paths are needed when an arnold scene is converted to usd, in order to ensure a roundtrip between ass & usd is possible.
However, when the usd file was authored from another source (i.e. Lookdevx), it's more natural to consider relative paths as USD would.

A solution for that could be that the writer adds a special attribute on the UsdShade primitive (e.g. arnold_relative_path), and that the reader looks for it to know what to do. This way, all usd files authored from an arnold scene will keep behaving as arnold expects, but files created elsewhere would now behave as UvTexture does, and would support UDIM tokens on relative paths.

### LINKED ISSUE #1881: Support UDIM and relative paths on mtlx image shaders
In order to support udim and relative paths on UvTexture, some hacks needed to be done since USD doesn't always manage to resolve it properly. In #1878 this was (optionally)  propagated to arnold:image shaders. But as discussed in the ticket comments, we should also do it for mtlx image shaders

sebastienblor:
When usd data is authored from an image shader, we add a boolean custom metadata "arnold_relative_path" to the filename attribute.

When reading the usd data, this will tell arnold to use the filename as-is, without any additional hack for relative paths & udims. But without this flag, the behaviour will be the same as for UvTexture shaders.

I'm updating test_1546 that now needs this metadata to work properly.
Also adding test_1878 which is identical to test_0225 but with an arnold:image shader instead of UvTexture.

EDIT: I'm extending this PR to handle all asset attributes instead of having a white list of shaders & attributes to support. Whatever the attribute is, we will look for the eventual metadata to skip the path remapping logic

Fixes #1878
Fixes #1881

**Additional context**


libs/translator/reader/read_shader.cpp
```diff
@@ -232,7 +232,8 @@ AtNode* UsdArnoldReadShader::Read(const UsdPrim &prim, UsdArnoldReaderContext &c
     bool isArnoldShader = (TfStringStartsWith(id.GetString(), str::t_arnold_prefix.GetString()));
     const AtNodeEntry* nentry = isArnoldShader ? AiNodeEntryLookUp(id.GetString().substr(7).c_str()) : nullptr;
 
-    bool isUvTexture = (id == str::t_UsdUVTexture);    
+    bool isTextureShader = (id == str::t_UsdUVTexture || id == str::t_arnold_image);
```
BigRoy, line 235: Would we need to do any special consideration for MaterialX images/textures?

libs/translator/reader/read_shader.cpp
```diff
@@ -297,10 +298,23 @@ AtNode* UsdArnoldReadShader::Read(const UsdPrim &prim, UsdArnoldReaderContext &c
                     arrayType = (defaultValue) ? AiArrayGetType(defaultValue->ARRAY()) : AI_TYPE_NONE;
                 }
             }
-            if (isUvTexture && attrName == str::t_file) 
+            // We need a special treatment for texture shaders with relative paths, 
+            // as USD won't resolve them properly with relative paths + UDIMs (see #1129 and #1097)
+            // For arnold image shaders, we want to do the same thing *unless* a customData metadata
+            // arnold_relative_path is defined and set to true  (see #1546 and #1878)
```
BigRoy, line 304:
```suggestion
            // arnold_relative_path is defined and set to true (see #1546 and #1878)
```
Cosmetics

libs/translator/reader/read_shader.cpp
sebastienblor: good point. I created #1881 for that

sebastienblor:
Since we don't need to have a special default logic for the arnold image shaders anymore, we can now make the behaviour for udim & relative paths more generic. This will also address Roy's point about materialx shaders for #1881, and will work for the different mtlx texture shaders (ND_image*, ND_tiledimage* , ND_triplanarprojection*)

So now, every time we read an asset parameter, we first ask USD to resolve it. 
If it fails to do so, that the path is relative, and that the metadata "arnold_relative_path" is NOT enabled, then we go through the logic of remapping the relative path ourselves.

Since arnold image shaders are now authored with this metadata, when we read them back, we will have the expected roundtrip. Note that we could also use the metadata on other shader types (UvTexture, mtlx textures, etc...) but in practice it wouldn't make much sense.

I'm adding test_1881 that has the same scene as test_0225 & test_1878, but with a Mtlx image shader.

cpichard (approved):

BigRoy:
> So now, every time we read an asset parameter, we first ask USD to resolve it.
> If it fails to do so

Is this entirely true? Wouldn't a relative path, when USD tries to resolve it have USD anchor it to the USD layer the opinion is in. Technically there could be a relative file there with the same name, but the `arnold_relative_path` metadata still being authored to a different search path. In this scenario would the path still be picked from the USD resolving or the one from the `arnold_relative_path`.

If this is a scenario where we only want to use the `arnold_relative_path` if file is not found relatively by USD then we might want to clarify maybe that it's a "fallback" only and not an override?

PR #1882, opened 2024-03-08
Move the ReadMatrix code to workaround existing arnold issues

### LINKED ISSUE #168: Implement procedural_update function
Procedurals can define a function `procedural_update` to control what happens during IPR updates. We should implement it in the usd procedural

sebastienblor:
A complementary change to what was done in #1847  for #168 . I'm removing the previous AiNodeReset that shouldn't be needed  during interactive updates (and that can cause slowdowns). Also, I'm moving all ReadMatrix calls in read_geometry to happen before other parameters are set (nsides, vlist, etc...). This doesn't change the data that is created, but the order of operations does workaround an existing issue in arnold (ARNOLD-14829)

Fixes #168

### DIFF
--- a/libs/translator/reader/reader.cpp
+++ b/libs/translator/reader/reader.cpp
@@ -1101,11 +1101,11 @@ void UsdArnoldReaderThreadContext::SetDispatcher(WorkDispatcher *dispatcher)
 AtNode *UsdArnoldReaderThreadContext::CreateArnoldNode(const char *type, const char *name)
 {   
     // If we're doing an interactive update, we first want to check if the AtNode
-    // already exists. If so, we need to reset it so that it can be properly read again
+    // already exists. If so, we return it
     if (_reader->IsUpdating()) {
         AtNode *node = _reader->LookupNode(name);
         if (node) {
-            AiNodeReset(node);
+            // Note: should we reset the node ?
             return node;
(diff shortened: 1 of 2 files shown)


libs/translator/reader/reader.cpp
```diff
@@ -1101,11 +1101,11 @@ void UsdArnoldReaderThreadContext::SetDispatcher(WorkDispatcher *dispatcher)
 AtNode *UsdArnoldReaderThreadContext::CreateArnoldNode(const char *type, const char *name)
 {   
     // If we're doing an interactive update, we first want to check if the AtNode
-    // already exists. If so, we need to reset it so that it can be properly read again
+    // already exists. If so, we return it
     if (_reader->IsUpdating()) {
         AtNode *node = _reader->LookupNode(name);
         if (node) {
-            AiNodeReset(node);
+            // Note: should we reset the node ?
```
BigRoy, line 1108: Ha - who or what would answer this? Testing? Future bugs?

sebastienblor: this is an internal node we're leaving for ourselves. The whole PR can only make sense internally as it's based on unreleased arnold & mtoa code

cpichard (approved):

PR #1887, opened 2024-03-18
Fix PointInstancer velocities support

### LINKED ISSUE #1868: Procedural does not  render motion blur for primvar.velocities
**Describe the bug**
Render delegate renders motion blur, but procedural (sick, kick) do not.

If I convert the usda file to ass, I can see that the motion keys are exactly the same for each point.
In contrast, the hddebug ass file from Solaris has good motion keys.

**To Reproduce**
1. Extract the attached zip, load the hip file, and render with Direct USD Render.
or
kick the attached usd file

[particles_mb.zip](https://github.com/Autodesk/arnold-usd/files/14363100/particles_mb.zip)
[usda.zip](https://github.com/Autodesk/arnold-usd/files/14363115/usda.zip)

**Expected behavior**

**Screenshots**
primvar_velocities

**Used Software Versions**
 - Houdini, HtoA

sebastienblor:
This PR fixes velocity blur in the Point instancer for the procedural. We were calling a dedicated API (ComputeInstanceTransformsAtTimes) that was supposed to handle this, but it appears that this USD API has too many restrictions, and most of the time it was skipping motion blur. 

In this PR instead, we're now calling another version of this API for each time sample, with the expected arrays for each attribute. 

This way we can consider the desired time for each attribute, e.g. we want to get the velocities and positions array at the "current frame", so that we can extrapolate for the different keys. Otherwise we get positions at each time sample.

The previous flags IncludeProtoXform and ExcludeProtoXform are now replaced by either specifying a SdfPathVector for the protos or just passing an empty vector.

Adding test_1868, that also passes in hydra mode (it used to work in hydra, but not in usd)

Note : the description in #1868 mentions attributes primvars:velocities, which this PR does *not* address. You need to set the attribute as "velocities" as this is what USD expects. In general, hydra handles attributes the same way whether the primvars namespace is there or not, but that's not the case when we read the UsdGeomMesh API to get attributes

Fixes #1868


sebastienblor:
In the last commit, I had to do a change in the render delegate side, to match the new procedural behaviour :
Since orientations & scaling are independant of velocities & positions, I'm always sampling them at the shutter times so that eventual motion blur caused by scale & rotation can be applied too. Only positions & velocities should be sampled at the "current" frame.

cpichard (approved):

PR #1909, opened 2024-05-07
Read deform_keys independantly of the primvar interpolation

### LINKED ISSUE #1908: Inconsistency with deform_keys between usd and hydra
When the USD translator looks for the amount of deform_keys, it checks the primvar interpolation and only reads it if it's the expected one.
This doesn't seem to be the case in Hydra, where the primvar is read whatever the interpolation is.
This is causing some scenes to have several keys in hydra, but only 2 in the procedural.

sebastienblor:
When we read the amount of time keys (primvars:arnold:deform_keys), we no longer check the primvar interpolation.
This is now returning the same result as in the render delegate

Fixes #1908

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Bug fixes
 - [usd#1900](https://github.com/Autodesk/arnold-usd/issues/1900) - Fix transform hierarchies for Arnold non-xformable primitives
-- [usd#1908](https://github.com/Autodesk/arnold-usd/issues/1908) - Read deform_keys independantly of the primvar interpolation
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Bug fixes
 - [usd#1900](https://github.com/Autodesk/arnold-usd/issues/1900) - Fix transform hierarchies for Arnold non-xformable primitives
+- [usd#1908](https://github.com/Autodesk/arnold-usd/issues/1908) - Read deform_keys independantly of the primvar interpolation
```
cpichard, line 11: independantly -> independently

cpichard (dismissed):

cpichard (approved):

PR #1915, opened 2024-05-28
Ensure polymesh topology attributes are not written to USD when not set

### LINKED ISSUE #1914: Ensure polymesh topology attributes are not written to USD when not set
Our usd writer is meant to skip the authoring of topology arrays when they're not set, by checking if the arrays are null.
This doesn't seem to be enough as Arnold can turn them into zero-element arrays. This is preventing from getting the expected behaviour with MaxUSD

sebastienblor:
We were taking into account the use case where the vertex attributes had a nullptr array pointer, but we must also consider the use case where they have a zero-element array, as Arnold seems to be setting this kind of arrays by default. This is needed to handle maxUSD exports in a better way

Fixes #1914

### DIFF
--- a/libs/translator/writer/write_geometry.cpp
+++ b/libs/translator/writer/write_geometry.cpp
@@ -48,7 +48,7 @@ void UsdArnoldWriteMesh::Write(const AtNode *node, UsdArnoldWriter &writer)
     writer.SetAttribute(mesh.GetOrientationAttr(), UsdGeomTokens->rightHanded);    
     AtArray *vidxs = AiNodeGetArray(node, AtString("vidxs"));
     VtArray<int> vtArrIdxs;
-    bool exportVertices = true;
+    bool exportVertices = false;
 
     if (vidxs) {
         unsigned int nelems = AiArrayGetNumElements(vidxs);


MasterZap (approved):
Looks good to me!

MasterZap (approved):

libs/translator/writer/write_geometry.cpp
```diff
@@ -48,13 +48,19 @@ void UsdArnoldWriteMesh::Write(const AtNode *node, UsdArnoldWriter &writer)
     writer.SetAttribute(mesh.GetOrientationAttr(), UsdGeomTokens->rightHanded);    
     AtArray *vidxs = AiNodeGetArray(node, AtString("vidxs"));
     VtArray<int> vtArrIdxs;
+    bool exportVertices = true;
```
cpichard, line 51: Just to be sure, shouldn't exportVertices be initialised to false ? otherwise if vidxs is nulltr then the vertexCount will be exported, may be that is what we want...

MasterZap: I think you are right

sebastienblor:
For the record, I did something wrong in the creation of the fix-7.3.2 branch, so the commit d478600 is no longer showing up in the list of modified files here. It was properly merged though

PR #1921, opened 2024-06-03
Fix IPR update of linear curves in the render delegate

### LINKED ISSUE #1861: BasisCurves disappearing on update
**Describe the bug**

It seems that Arnold incorrectly 'updates' BasisCurves. This is mostly noticable in Houdini Solaris, where just moving the camera or clicking through frames makes the curves disappear. There, a "Restart Render" is required for the hair curves to show up.

**To Reproduce**

#### Reproduce in Houdini 20.0.590 with HtoA 6.2.5.2

Here's a simple usda file with some curves generated by "Draw Curve" with SOP Create in LOPs in Houdini.

[curves.zip](https://github.com/Autodesk/arnold-usd/files/14303407/curves.zip)

1. Sublayer this file in Houdini
2. Move camera - curves disappear
(Note that when using the free camera, and not camera tied to view that the bug occurs with Display Options > Camera > Clipping > Homing Auto Adjusts > Neither Clip Plane set)
image

This error is shown in the render logs.
```
00:22:39  1727MB ERROR   |  [curves] /sopcreate1/curve_0: invalid number of points (found 168, expected 484)
```

#### Reproducable in USDView

This can be harder to reproduce in USDView since there it doesn't occur on camera movements - but it does when you SWITCH cameras, e.g. switch from free cam to the camera in the USD file or vice-versa.

Since this is on camera updates it could be that #1854 resolves this already - however even if _that_ were the bug there still appears to be a bug on "syncing updates" from the scene for Arnold USD to curves/hairs where Arnold thinks its getting the wrong point counts.

With some other drawn curves, e.g. the error is:
```
00:00:11  1754MB ERROR   |  [curves] /sopcreate1/curve_0: invalid number of points (found 272, expected 784)
```

What's interesting is the ratio of points that it expects to what it found:

```
784 / 272 = 2,882352941176471
484 / 168 = 2,880952380952381
```
seems the ratio it's wrong is quite similar.

When trying it with a very simple single curve:
```
00:00:02  1832MB WARNING |  [curves] /sopcreate1/curve_0: invalid number of points (found 7, expected 0 or 19)
```

Here maybe it's interesting that it's expecting either 0 or 19.

Anyway, `9 / 17 = 2,714285714285714`  a bit further away, but still around 2.8.

**Expected behavior**

Curves/hairs should render fine, and not disappear.

**Screenshots**

**Used Software Versions**
 - HtoA: 6.2.5.2
 - OS: Windows 10
 - Houdini: 20.0.590

**Additional context**

I've reported this issue on Arnold Support as well - with less information. The ticket there is: HTOA-2504
There I mostly reported it happening on "frame changes" over time.

sebastienblor:
When a curves with linear basis are rendered, Arnold modifies them under the hood at node initialization, to convert them to bezier curves.
Because of that, when only positions (or certain primvars) are modified during Hydra updates, we get a mismatch in the amount of points. 
The fix is to ensure that as soon as any topology-related parameter is modified on linear curves, all of them are updated

Fixes #1861


BigRoy:
> When a curves with linear basis are rendered, Arnold modifies them under the hood at node initialization, to convert them to bezier curves.
> Because of that, when only positions (or certain primvars) are modified during Hydra updates, we get a mismatch in the amount of points.
> The fix is to ensure that as soon as any topology-related parameter is modified on linear curves, all of them are updated

Ah, that makes sense and looking forward to seeing this fixed in production. It's been an annoyance to say the least. :)

cpichard (approved):

PR #1931, opened 2024-06-11
Ensure subdiv_iterations is not set uselessly during procedural updates

### LINKED ISSUE #1929: Avoid useless attribute changes during procedural updates
During procedural updates, some attributes are always set back and forth, which is not efficient in some use cases.
For example, subdiv_iterations  is always reset to 0 before being set back to the value of primvars:arnold:subdiv_iterations, which forces the modified geometries to reload over and over on the arnold internal side

sebastienblor:
During procedural updates of usd geometries, we were always setting smoothing and subdiv_iterations back to the desired default value, before the attributes were eventually set explicitely during ReadArnoldParameters.

This PR only resets the attribute values if we find that they're not explicitely set as primvars:arnold:subdiv_iterations and primvars:arnold:smoothing, in which case the attributes would be set during ReadArnoldParameters

Issues fixed in this pull request
Fixes #1929

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,9 @@
 ## Pending feature release
 
 ## Pending bugfix release
-- [usd#1923](https://github.com/Autodesk/arnold-usd/issues/1923) - Fix instance primvar indices with multiple prototypes
+- [usd#1923](https://github.com/Autodesk/arnold-usd/issues/1923) - Fix instance primvar indices 
+with multiple prototypes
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,7 +4,9 @@
 ## Pending feature release
 
 ## Pending bugfix release
-- [usd#1923](https://github.com/Autodesk/arnold-usd/issues/1923) - Fix instance primvar indices with multiple prototypes
+- [usd#1923](https://github.com/Autodesk/arnold-usd/issues/1923) - Fix instance primvar indices 
+with multiple prototypes
```
cpichard, line 8: there is a line break, is it intended ?

PR #1938, opened 2024-06-13
restart the render when we receive a new houdini:frame render param

### LINKED ISSUE #1919: Render delegate doesn't render frame ranges when running with husk as host
**Describe the bug**
When we run a render through husk with a frame range, the arnold render delegate will render only the first frame and copy it.

**To Reproduce**
1. husk --complexity low -R Arnold -r 320 240 -o render.%f.tif -f 1 -n 10 chess_animated.usd
2. replace chess_animated.usd by an animated scene

**Expected behavior**

cpichard:
Fixes #1919

### DIFF
--- a/libs/render_delegate/render_delegate.cpp
+++ b/libs/render_delegate/render_delegate.cpp
@@ -636,6 +636,12 @@ void HdArnoldRenderDelegate::_SetRenderSetting(const TfToken& _key, const VtValu
         return colorManager;
     };
 
+    // When husk/houdini changes frame, they set the new frame number via the render settings.
+    if (_key == str::t_houdiniFrame) {
+        // We want to restart a new render in that case.
+        _renderParam->Restart();
+    }
+
     // Special setting that describes custom output, like deep AOVs or other arnold drivers #1422.
     if (_key == _tokens->delegateRenderProducts) {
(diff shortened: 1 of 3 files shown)


libs/render_delegate/render_delegate.cpp
```diff
@@ -636,6 +636,12 @@ void HdArnoldRenderDelegate::_SetRenderSetting(const TfToken& _key, const VtValu
         return colorManager;
     };
 
+    // When husk/houdini changes frame, they set the new frame number via the render settings.
+    if (_key == str::t_houdiniFrame) {
```
sebastienblor, line 640: I guess we'll receive that render setting even before the first render. Can it be a problem to call Restart() if the render didn't actually start yet ?

cpichard: Good question, we receive some render settings before the first render, the `Restart()` function sets `_needsRestart` which is only used when the render is finished or paused (but it was set before so unlikely to return something else than false). `Restart()` also sets `_paused` to false, which will be used if the render was paused and will call a AiRenderResume. So I am thinking it should be ok, what do you think ?

sebastienblor (approved):

PR #1947, opened 2024-06-25
Use batch session for single-frame husk renders

### LINKED ISSUE #1918: Make render delegate works in batch mode when running under husk
At the moment when we run husk the render delegate uses the interactive session mode as it's not clear if husk will update the scene after having it passed to arnold. This result in an increased memory consumption as the scene remains in memory.
We want to double check that husk can be used in interactive mode and update arnold-usd such that husk uses the batch mode which deletes the scene in memory when arnold has its internal data ready to render.

sebastienblor:
For batch renders of the render delegate we now use a batch render session instead of an interactive one, in order to optimize memory consumption.
We do an exception when the husk command line is executed with the -n argument that triggers the rendering of multiple frames in the same render session, which forces us to have an interactive session

Fixes #1918

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 ### Feature
 - [usd#1814](https://github.com/Autodesk/arnold-usd/issues/1814) - Support skinning on USD curves and points
+- [usd#1918](https://github.com/Autodesk/arnold-usd/issues/1918) - Use batch render sessions for husk renders
 
 ### Bug fixes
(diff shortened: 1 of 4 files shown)


cpichard (commented):
LGTM, tested with husk on mac, it is working. It's just missing the changelog.
(I wish I had seen that the render settings were passed in the plugin)

sebastienblor:
thanks, added the changelog

cpichard (approved):

PR #1952, opened 2024-06-26
Don't write camera aperture parameters if they're already set

sebastienblor:
In Maya USD exports, arnold is overwriting the aperture parameters to wrong values (internal ticket MTOA-1951).
Reason is that the arnold usd writer cannot set a correct aperture value since arnold cameras don't have this attribute, nor the focal length which would be needed to calculate it based on the fov.

Note that we had decided to author those values, even with an approximate value, because otherwise the exported cameras cannot be used in other usd tools.


libs/translator/writer/write_camera.cpp
```diff
@@ -58,21 +58,28 @@ void UsdArnoldWriteCamera::Write(const AtNode *node, UsdArnoldWriter &writer)
...
-            verticalAperture *= (float)AiNodeGetInt(options, AtString("yres")) / (float)AiNodeGetInt(options, AtString("xres"));
+        // Arnold only knows about FOV and not about horizontal & vertical aperture.
+        // We want to dump some aperture values in the usd file, as this is used by other
+        // usd tools, but we can't compute an exact value (we'd need a focal length)
+        // we only author these values if they weren't previously set. MTOA-1951
+        if (!cam.GetHorizontalApertureAttr().HasAuthoredValue() && 
+                !cam.GetVerticalApertureAttr().HasAuthoredValue()) {
+            
+            float fov = AiNodeGetFlt(node, AtString("fov"));
+            float horizontalAperature = tan(fov * AI_DTOR * 0.5f);
```
cpichard, line 69: could be a good time to fix the typo horizontalAperature -> horizontalAperture

cpichard (approved):

PR #1953, opened 2024-06-26
Fix potential "invalid Cache ID" error

### LINKED ISSUE #1902: Tests failing sporadically with invalid Cache ID
**Describe the bug**
Some of our tests are failing with the following error:

```
00:00:00   146MB WARNING |  [usd] Cache ID not valid 9223001
00:00:00   146MB ERROR   |  [usd] Failed to open file ()
```

We suspect that it could be a race condition in the procedural when the translator needs to create a nested child procedural.
It has been ongoing since a long time with all versions of arnold.

cpichard:
- promote _cacheId type to long int even though it should not be worrying for now. We could also store the stageId as a string instead. TBD
- When initialising the cache id value we Insert the stage and set the counter to one. However the Insert function can return the id of a stage which has already been inserted and for which the counter is already set, resetting to one will end up in the destruction of the stage when it is still needed by other procedurals and will error with "invalid Cache ID".  The counter is now incremented instead of reset to zero to have the correct number of stages, hopefully that should fix the issue.

Fixes #1902


libs/translator/reader/reader.cpp
```diff
@@ -1039,8 +1039,8 @@ void UsdArnoldReader::InitCacheId()
     UsdStageCache::Id id = stageCache.Insert(_stage);
     // now our reader will have a cacheID
     _cacheId = id.ToLongInt();
-    // there's one ref count for this cache ID, for the current proc
-    s_cacheRefCount[_cacheId] = 1; 
+    // stageCache.Insert can return an existing stage, so we increase the ref count for that stage in case it exists
+    s_cacheRefCount[_cacheId]++;
```
sebastienblor, line 1043: are we sure the first time we add an entry in the map by doing s_cacheRefCount[_cacheId] , the value will be initialized to 0 ? I would think this code would trigger an uninitialized variable.

Wouldn't it be safer to write it like :
```
if (s_cacheRefCount.find(_cacheId) == s_cacheRefCount.end())
    s_cacheRefCount[_cacheId] = 1;
else
    s_cacheRefCount[_cacheId]++;


```

ThiagoIze: yes, `s_cacheRefCount[_cacheId]` will default value initialize, which is 0.

sebastienblor: ok cool

sebastienblor (approved):

Issue #802, opened 2021-06-01
B-spline curves not using radius in procedural


compso:
**Describe the bug**
When rendering the attached .usd with the USD procedural it produces an error due to an incorrect number of points on either the curve or the radius parameter

Same data renders fine in hydra render delegate

**To Reproduce**
1. load the attached .usd and render using kick
2. `kick -i hair_test.usd -default_nodes all -r 1920 1080`
3. you should get a warning like:
  ```00:00:00   238MB WARNING | /Root/RootOffset/g_geometry/groom/hair_test/hair_testShape.id: use type UINT, not ARRAY
00:00:00   238MB WARNING | /Root/RootOffset/g_geometry/groom/hair_test/hair_testShape: could not set ARRAY parameter "id"
...
00:00:00   258MB WARNING |      [curves] /Root/RootOffset/g_geometry/groom/hair_test/hair_testShape: invalid radius specification: found 3135, expected 1, 285 or 2565 (constant, uniform or varying); just using the first
```

**Expected behavior**
Curves render with correct radius

**Screenshots**
Hydra
![image (6)](https://user-images.githubusercontent.com/332415/120323743-b975d400-c2dd-11eb-8926-aef7c2e5a3b6.png)
Arnold Procedural
![image (7)](https://user-images.githubusercontent.com/332415/120323748-baa70100-c2dd-11eb-8b9f-2526e23a6de8.png)

**Used Software Versions**
 - Arnold: 6.2.1.0

[hair_test.zip](https://github.com/Autodesk/arnold-usd/files/6576252/hair_test.zip)

PR #1962, opened 2024-07-09
Random curves width in hydra when primvars:radius is present

### LINKED ISSUE #1961: Hydra test_0179 random curves width when primvars:radius is present
The work done for #802  was not ported to the render delegate. Therefore the hydra translators receive 2 different primvars : "width" and "primvars:radius", the later being expected to be ignored. Since it's currently not ignored, it could override the correct "width" values depending on the order in which attributes are found.
This gives random failures in test_0179 in hydra mode

sebastienblor:
This PR ports the logic from #802 to the hydra delegate, and ensures that primvars (without the arnold namespace) corresponding to arnold builtin attributes, are skipped. In test_0179, the curves primivite has a width value, as well as primvars:radius. The later is now ignored in hydra too, which allows the test to pass. Note that a special case needed to be handled for uv attributes for which a special case was done.
This allows to add test_0179 and test_0209 to the hydra list

Fixes #1961 

**Additional context**


libs/render_delegate/utils.cpp
```diff
@@ -381,12 +381,18 @@ bool ConvertPrimvarToBuiltinParameter(
...
+    HdArnoldRayFlags* sidedness, HdArnoldRayFlags* autobumpVisibility, 
+    HdArnoldRenderDelegate *renderDelegate, bool skipExisting)
 {
     // Remap primvars:arnold:xyz parameters to xyz parameters on the node.
     if (ConvertPrimvarToBuiltinParameter(node, name, value, visibility, sidedness, autobumpVisibility, renderDelegate)) {
         return;
     }
+    // If this attribute already exists in the node entry parameters list, 
+    // we must skip it #1961
+    if (skipExisting && AiNodeEntryLookUpParameter(AiNodeGetNodeEntry(node), AtString(name.GetText())) != nullptr) {
```
cpichard, line 393: I am wondering if we could just test if arnoldAttributeName == str::t_uvs here ? and remove the extra argument skipExisting ?

sebastienblor: it's because this piece of code in utils is generic and doesn't know if the object is a curves (which expected "uvs"), or if it's a mesh (which would expect "uvlist"). So I expect that at some point we might need the meshes to also use this argument.

cpichard (approved):

PR #1963, opened 2024-07-10
Pr/1959

sebastienblor:
- 
- 

Fixes #
Fixes #

**Additional context**


libs/render_delegate/utils.cpp
```diff
@@ -681,30 +681,24 @@ void HdArnoldGetPrimvars(
             // The number of motion keys has to be matched between points and normals, so if there are multiple
             // position keys, so we are forcing the user to use the SamplePrimvars function.
             if (multiplePositionKeys && primvarDesc.name == HdTokens->normals) {
+                HdArnoldInsertPrimvar(primvars, primvarDesc.name, primvarDesc.role, primvarDesc.interpolation, 
 #ifdef USD_HAS_SAMPLE_INDEXED_PRIMVAR
```
cpichard, line 685: May be we could remove this ifdef as it was used for very old version ?

sebastienblor: it's needed for versions older than 21.05, so I think we still need it ?

cpichard (approved):

PR #1964, opened 2024-07-10
Tweak test_1476 to work in hydra

sebastienblor:
There is a small difference in the amount of key samples used in hydra as compared to what we do in USD : hydra sometimes uses an additional, useless sample.  This difference can only cause tiny differences, as it was happening in test_1476. In this PR I'm just tweaking the shutter start/end to ensure both code paths end up with the same amount of samples, which allows to add it to the hydra list.

I'm also adding a change in the testsuite for tests using the "resaved" argument. We want to ensure we add the params list when resaving, otherwise the resaved file doesn't take into account the frame to be translated, which makes debugging difficult


cpichard (dismissed):
lgtm, can't get a windows machine to run the tests on though

PR #1966, opened 2024-07-10
Write color manager attributes in render settings

### LINKED ISSUE #1965: Author color manager parameters in Render Settings
Currently, when we write a USD file, we get a RenderSettings primitive with an attribute `arnold:color_manager` pointing to a primitive "ArnoldColorManager". However this primitive is currently ignored by Hydra. In order to fix this, we can either an a usdImaging adapter for arnold color managers, or we can change our writer to author the color manager attributes directly in the render settings. This is similar to what we do for RenderVars and drivers.
Then, once we support this format in the procedural & delegate, we can switch our existing tests that are failing because of that (e.g. test_1459)

sebastienblor:
The USD writer now authors color manager attributes in the render settings primitive, with the namespace `arnold:color_manager` .
We no longer author the previous "color_manager" attribute that pointed to the arnold color manager node (this required to update test_0174 that counts the amount of authored attributes).

This format is now supported in the procedural as well as the render delegate. Note that we still need to duplicate the whole render settings translation in rendersetting_utils, which we should fix at some point.

This allows to add test_1459 to the hdra list

Fixes #1965


cpichard (approved):

cpichard:
It's passing the tests, but the compilation on 22.08 gives the following error: 
```
libs/translator/writer/write_options.cpp:187:44: error: no member named 'CreateRenderingColorSpaceAttr' in 'arnoldPxrInternal__pxrReserved__::UsdRenderSettings'
        writer.SetAttribute(renderSettings.CreateRenderingColorSpaceAttr(), renderingSpaceToken);
```

sebastienblor:
fixed

cpichard (approved):

PR #1969, opened 2024-07-11
Remove ifdefs for USD versions older than 21.05

sebastienblor:


cpichard (dismissed):
\o/ awesome.
I think this one needs a line in the changelog specifying the minimum version supported by arnold-usd now

PR #1975, opened 2024-07-12
Only create default hydra shaders when needed

### LINKED ISSUE #1974: Hydra: Default shaders should only be created when needed
Currently, the render delegate creates the default shaders (surface & volume) as soon as it starts. It should instead create them only when needed, as the procedural does

sebastienblor:
Instead of creating the default surface & volume shaders immediately when the render delegate is started, we now create them lazily, when actually needed. 
We're using a double check so that this can work in parallel

Fixes #1974


cpichard (approved):
lgtm

PR #1988, opened 2024-07-16
Author familyName & type when writing GeomSubset primitives

### LINKED ISSUE #1987: GeomSubsets not authored properly in USD
When an arnold polymesh has geometry subsets, a GeomSubset primitive is written to USD.
However, it's authored without familyName & familyType, which prevents it from working properly in Hydra.

sebastienblor:
When authoring GeomSubsets in the usd writer, we need to create them with a familyName & familyType, otherwise Hydra filters them out. Doing this change allows to run test_0070 and test_0102 in hydra mode.

Also, the usd files from tests 35, 36 and 224 were missing these fields. Updating them allows to add them to the hydra list too.

While doing the change in the writer, I saw we were creating hardcoded TfToken strings, so I replaced them with the usual private token list.

Fixes #1987


cpichard (commented):
It needs a line in the changelog, otherwise it's looking good and passing the tests

cpichard (approved):

PR #1993, opened 2024-07-17
Allow hydra skinned positions to have more than 3 keys

### LINKED ISSUE #1992: Hydra skinned positions cannot have more than 3 keys
When an object is skinned and we render it through Hydra, we call 
`HdExtComputationUtils::SampleComputedPrimvarValues`  with  `maxSamples = 3` . 
If there should be more keys, the returned times are clamped to 3 and we then get only the beginning of the shutter range.
This shows up in test_1443 where the motion length is very different from the expected one. It happens because we only get the first 3 keys out of 7.

sebastienblor:
I verified that this third argument in `HdExtComputationUtils::SampleComputedPrimvarValues` does nothing more than clamping the returned time samples. So I'm increasing this max sample count argument, which is required in the function. We could consider that we don't want any max value, however we can imagine that in some use case perhaps the amount of computed primvars keys can be overkill. I'm choosing an arbitrary number of 30 for the maximum keys here, but we could choose any other value.

Also, the value store was created with a template value called `HD_ARNOLD_MAX_PRIMVAR_SAMPLES` which is a confusing name since this doesn't change the max primvars samples at all, it's just the default samples & times count that will be allocated in the vectors. So I'm renaming this variable to avoid confusions.

These changes improve a lot the result of test_1443 in hydra mode and we now have a motion range that is consistent. However, the test is still not passing as there are other, unrelated, discrepancies. So I'm not adding it to the list yet

Fixes #1992


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 - [usd#1961](https://github.com/Autodesk/arnold-usd/issues/1961) - Random curves width in Hydra when radius primvars are authored
 - [usd#1977](https://github.com/Autodesk/arnold-usd/issues/1977) - Aov shaders not set properly in hydra mode of the scene format
 - [usd#1984](https://github.com/Autodesk/arnold-usd/issues/1984) - Cylinder lights not taking normalization into account through USD
+- [usd#1992](https://github.com/Autodesk/arnold-usd/issues/1992) - Hydra skinned positions cannot have more than 3 keys
```
cpichard, line 22: I suggest to rephrase this line, as when reading it I am not sure if it is a fix or if it is forcing the position to have 3 keys or less. May be something like "Allow hydra skinned positions to have more than 3 keys" ?

cpichard (approved):
Looks good ! just a minor change in the changelog would be great

PR #2016, opened 2024-07-20
Support points with empty widths

### LINKED ISSUE #2015: Support empty widths in hydra curves
USD accepts points with empty widths, and we support that in our usd translator. Since Arnold doesn't support that, we need to provide a widths array of the same size as the amount of points.

However, this logic in our usd procedural hasn't been ported to Hydra. This is needed to support test_0185 in hydra mode

sebastienblor:
We need to port the logic from the usd procedural to hydra. If we find widths as an empty array, we need to create in arnold an array of the same size as the amount of points, with a value of 0. This way, renders don't error out.
This solves test_0185 in hydra mode

Fixes #2015

### DIFF
--- a/libs/render_delegate/points.cpp
+++ b/libs/render_delegate/points.cpp
@@ -49,11 +49,6 @@ void HdArnoldPoints::Sync(
         transformDirtied = true;
     }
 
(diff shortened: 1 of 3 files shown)


cpichard (commented):
LGTM ! it's just missing the changelog update.

sebastienblor:
added

cpichard (approved):

PR #2022, opened 2024-07-22
Proper support of arnold:visibility primvar in hydra

### LINKED ISSUE #1610: usd from Maya does not render if shapes have arnold:visibility primvar
**Describe the bug**
USD geometry exported from Maya does not render in the viewport.

**To Reproduce**
1. In Maya, disable one of the visibility flags (for example, diffuse reflection)
2. Export to usd (using either Maya USD or Arnold USD)
3. In Houdini, bring in the usd file as a Sublayer (or Reference)
4. Change the renderer to Arnold. The geometry is not visible (but it is with Houdini GL, Karma, and Storm).

Another way to repro, with a usd file that does not have arnold:visibility in it, is to just Edit arnold:visibility
As soon as you do, the object disappears from the render.

If you export USD from Houdini and load it into usdview, same problem.

My repro usda files and a hip file:
[visibility.zip](https://github.com/Autodesk/arnold-usd/files/12252607/visibility.zip)

**Used Software Versions**
 - Arnold: Latest beta

sebastienblor:
In the render delegate, getting the arnold:visibility primvar was assuming that it was necessarily authored as an integer. 
I'm just using VtValueGetInt instead and this fixes the issue. Added test_1610 with such a scene

Fixes #1610


sebastienblor:
Note that I'm seeing quite a few other occurrences of VtValue::Get being called with an explicit type, which would also fail when the type is different. 
After this issue is merged to fix and back to master, I'll do another round of changes to use VtValueGet* whenever it can be useful

cpichard (approved):

PR #2029, opened 2024-07-24
Ensure drivers are authored through node graphs

### LINKED ISSUE #2025: Write imagers to USD through a node graph
When we write a USD scene based on an arnold scene  that has imagers, we need to write them through an ArnoldNodeGraph, as it can be exported from Solaris. This way the usd scenes can be rendered in both usd and hydra

sebastienblor:
When imagers are linked to a driver, we now author them through a ArnoldNodeGraph primitive.
Also, instead of authoring them as arnold-specific schemas, they're now written as UsdShade shaders (since we already declare them in the Sdr registry). 

I had to support the use case where AddConnection is called with a a "link"  type (which is what shaders do), but with an attribute of type "node". In this case we need to set the connection with AiNodeSetPtr. 

I'm adding test_2025 that exercises the whole workflow of writing such a scene to usd
This allows to enable tests 78, 85, 89 in hydra.

Note that we're no longer reading user data on imagers since they're now shaders, and hydra will skip all the attributes that are not "inputs". But this is not a use case that is being used at the moment. Therefore I need to update test_1315 that was creating a user data on an imager and instead write it on a light. This allows to also enable test_1315 in hydra mode.

Fixes #2025


cpichard (commented):
looks good ! it is missing changelog update though

sebastienblor:
added

PR #2034, opened 2024-07-25
Write background / atmosphere / aov_shaders under node graphs

### LINKED ISSUE #2030: Write background and atmosphere shaders to USD through a node graph
Similar to what is done in #2025 for imagers, we also need to write background and atmosphere shaders to usd through node graphs, so that they can work in hydra mode

### LINKED ISSUE #2031: Write AOV shaders to USD through a node graph
Similar to what is done in #2025 for imagers, we need to author aov shaders through node graphs so that they can be loaded in hydra mode

sebastienblor:
When we author the arnold options as USD, we handle background, atmosphere and aov_shaders in a special way now. 
When shaders are connected to these attributes, we author them under a ArnoldNodeGraph primitive, which allows to support them in Hydra. I'm using a single, factorized function for the 3 attributes, since they're doing roughly the same thing.
It's a bit more complicated for aov_shaders though, which is an array attribute, so we need to add the suffix :id1, :id2, etc... to the node graph terminal

This solves 7 more tests in hydra mode. We need to adjust test_0174 which counts the options attributes, since these 3 attributes are no longer being authored to USD by default.

Fixes #2030 
Fixes #2031


cpichard:
The build fails on older versions, for 7.2.5 we got the following error:

libs/translator/writer/write_shader.cpp:53:57: error: use of undeclared identifier 'AI_NODE_IMAGER'
   53 |     if (AiNodeEntryGetType(AiNodeGetNodeEntry(node)) == AI_NODE_IMAGER) {

cpichard:
It was introduced earlier, here is a fix #2035

PR #2054, opened 2024-08-06
Visibility and sidedness attributes not supported in Arnold native hydra prims

### LINKED ISSUE #2051: Motion blur difference in scene imported with an ArnoldUsd prim
**Describe the bug**
The test_1331 renders slightly differently in the hydra testsuite, it could be a motion blur issue. This test uses an ArnoldUsd procedural to import the skinned geometry and if we replace it by a reference the test passes the hydra testsuite.
So the issue seems related to the use of a procedural with usdskel.
We want to "fix" the test_1331 by using a reference instead of an ArnoldUsd since we want to test the motion blur of skinned geometry, but we also want to create a new test with a simplified scene, easier to debug, that exhibits this motion blur issue with a procedural. And fix this issue.

### LINKED ISSUE #2053: arnold:visibility attribute not supported in Hydra native rprims
In arnold nodes rendered in hydra through native_rprim (e.g. ArnoldProcedural, ArnoldUsd, etc...), the attribute arnold:visibility isn't taken into account. 
This causes test_0090 to fail in hydra mode

sebastienblor:
The delegate code for native_rprims (used for arnold shapes, like ArnoldUsd, ArnoldProcedural, etc...) wasn't supporting properly some attributes, like visibility and sidedness. 
The code was a bit different than other parts of the code, and hadn't been touched for some time. I tried to make it consistent with our other prim types (mesh, procedural_custom, etc...). This fixes test_1331, test_0090 and test_0100

Fixes #2051 
Fixes #2053

### DIFF
--- a/libs/render_delegate/native_rprim.cpp
+++ b/libs/render_delegate/native_rprim.cpp
@@ -39,6 +39,7 @@ void HdArnoldNativeRprim::Sync(
     HdArnoldRenderParamInterrupt param(renderParam);
     const auto& id = GetId();
 
+    int defaultVisibility = AI_RAY_ALL;
     // Sync any built-in parameters.
     if (*dirtyBits & ArnoldUsdRprimBitsParams && Ai_likely(_paramList != nullptr)) {
         param.Interrupt();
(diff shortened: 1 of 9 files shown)


libs/render_delegate/native_rprim.cpp
```diff
@@ -49,22 +50,15 @@ void HdArnoldNativeRprim::Sync(
...
                         param.second, GetRenderDelegate());
+                if (param.first == str::t_visibility)
+                    defaultVisibility = (int)AiNodeGetByte(GetArnoldNode(), str::visibility);
             }
         }
     }
 
     // The last argument means that we don't want to check the sidedness.
     // The doubleSided attribute (off by default)  should not 
     // affect arnold native prims. Visibility should be taken into account though
```
cpichard, line 61: should this comment be removed ?

sebastienblor: right, done

cpichard (approved):

PR #2056, opened 2024-08-06
Support animated curves orientations in hydra

### LINKED ISSUE #2055: Support animated curves orientations in hydra
test_0095 is failing in hydra because a single key is read for the orientations. Since the points have multiple keys, this causes an arnold error

sebastienblor:
We need to do a special treatment when we're reading arnold:orientations, so that they can have multiple keys. Otherwise arnold would error out because the points keys doesn't match the orientation's.

This fixes test_0095. This test had a big threshold due to different results on Mac. I tried bumping the sampling instead and I lowered the threshold, we need to check if this passes on all OSes

Fixes #2055


libs/render_delegate/utils.cpp
```diff
@@ -595,6 +595,33 @@ void HdArnoldSetRadiusFromPrimvar(AtNode* node, const SdfPath& id, HdSceneDelega
...
+    int timeIndex = 0;
+    for (size_t i = 0; i < xf.times.size(); ++i) {
+        if (xf.times[i] >= 0) {
+            timeIndex = i;
+            break;
+        }
+    }
+    const auto& v0 = xf.values[timeIndex];
+    AtArray* arr = AiArrayAllocate(v0.size(), xf.times.size(), AI_TYPE_VECTOR);
+    AtVector* out = static_cast<AtVector*>(AiArrayMap(arr));
```
cpichard, line 617: I think here we normally have to resample the values such that they are uniformly distributed in time before passing them to arnold. Unless it's done before which I didn't check.

cpichard: there is an example here https://github.com/Autodesk/arnold-usd/blob/fa3ca6194bfde47856ffe69a8b2af3decda052b8/libs/render_delegate/utils.cpp#L527

sebastienblor: Ported the code to the new function.

cpichard (approved):

PR #2059, opened 2024-08-07
Support Plane primitives in the usd procedural

sebastienblor:
**Changes proposed in this pull request**
Adding a translator reader for UsdPlane primitives, using as a reference the code from UsdImaging planes https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usdImaging/usdImaging/implicitSurfaceMeshUtils.cpp#L438

Adding test_2058 with such a prim, that was already passing in hydra, and that now passes in USD too

**Issues fixed in this pull request**
Fixes #2058

Issue #1972, opened 2024-07-12
Disable subdivisions in hydra meshes if subdiv_iterations is 0


sebastienblor:
In the procedural, when subdiv_iterations is equal to 0, we ensure that subdiv_type is set to "none" #679 . We're not doing the same thing in hydra, which is failing test_1366 in hydra mode

PR #2068, opened 2024-08-09
fix changelog entry

cpichard:
- issue 1974 was copied as issue 1972, this PR updates issue 1972 with its original title


sebastienblor:
#1972  is already noted a few lines below. So it was a duplicated line in the changelog. I removed it in #2059 that was just merged. If it's ok we can close this PR

PR #2074, opened 2024-08-21
Add stats file/stats mode and profile file to render settings

### LINKED ISSUE #1118: Add profile/stats settings from the Render Settings
In a similar way that we are setting the log file, we need to set the profile/stats files from the render settings LOP.

This is for the delegate and the reader for kicking usd files when there is no parent proc.

TomMinor:
- stats:file, stats:mode and profile:file are now used by the render delegate if set on render settings
image
- HDARNOLD_stats_file has been added to override the stats file
- HDARNOLD_profile_file and HDARNOLD_log_file now work as expected

- Fixes #1118 

**Additional context**


cpichard (commented):
The changelog needs to be updated as well

sebastienblor (approved):

PR #2097, opened 2024-09-12
Shaders exports should be bound to a material

### LINKED ISSUE #2047: Shaders exports should be placed under a Material prim
When shaders are exported (alone) from arnold plugins, the root shader has a user data "material_surface" giving the name of the Material in the DCC. Same for shaders used as displacement with a user data "material_displacement". This is what we have in the Arnold scene description.

We should make it so that our writer puts such shaders under a Material primitive. This way the shaders can be rendered through hydra

sebastienblor:
When arnold plugins export shading tree libraries, a user data is added to the "root" shaders assigned to materials as surface / displacement / volume shaders. This allows to restore the materials when importing back the scenes to the DCC.
When such scenes are saved as USD, we want to use this user data to create a material primitive. This way, the representation in USD is the same as in the DCC.

This allows for example to export shaders from Maya and then use them in Solaris and be able to assign materials (which is not possible when the materials are missing)

Adding test_2047

Fixes #2047


libs/translator/writer/write_shader.cpp
```diff
@@ -82,4 +82,40 @@ void UsdArnoldWriteShader::Write(const AtNode *node, UsdArnoldWriter &writer)
...
+        if (!materialName.empty()) {
+            std::string matName(materialName.c_str());
+            _SanitizePrimName(matName);
+            std::string mtlScope = writer.GetScope() + writer.GetMtlScope();
+            writer.CreateScopeHierarchy(SdfPath(mtlScope));
+            matName = mtlScope + matName;
+            UsdShadeMaterial mat = UsdShadeMaterial::Define(writer.GetUsdStage(), SdfPath(matName));
+            const TfToken arnoldContext("arnold");
+            std::string shaderOutput = prim.GetPath().GetString() + std::string(".outputs:out");
+            UsdShadeOutput matOutput = (isDisplacement) ?
```
cpichard, line 114: here we only use the displacement and surface as the volume are treated the same way as surface. In the api there is a [CreateVolumeOutput](https://openusd.org/release/api/class_usd_shade_material.html#a4abdf5ff051e043d5b4b8bc5dd0a70f1) function that we could also use, that would handled internal changes by the api. What do you think ?

cpichard (approved):

PR #2113, opened 2024-09-26
Add support for transform_keys in xform primitives

### LINKED ISSUE #2111: Discrepancy in transformation motion blur between usd and hydra
We have a user scene rendering differently in USD vs Hydra mode, with artefacts showing up in hydra.
This affects rendering in Solaris

sebastienblor:
This PR includes different changes :
1) In Hydra, when receiving transform matrices that were not regularly sampled, we were not remapping them, which caused "ghosting" artefacts.
2) In USD we were considering deform_keys to control the transform samples, but not in Hydra. Now both are taking it into account
3) We now take into account a new primvar `arnold:transform_keys` to control the amount of output keys for transforms, in order to distinguish this from deform_keys.

So at the moment, for transforms we're first looking for transform_keys, and then at deform_keys, if they're present. This way we don't break existing scenes that were relying on deform_keys. However in a future breaking release, we will no longer take into account deform_keys for transformation, and only rely on transform_keys, which is what makes more sense on the user side.

Added test_2111 that reproduced the ghosting artefacts in hydra

Fixes #2111


libs/render_delegate/utils.cpp
```diff
@@ -195,12 +195,51 @@ void HdArnoldSetTransform(AtNode* node, HdSceneDelegate* sceneDelegate, const Sd
...
-        ConvertValue(mtx, xf.values[i]);
+
+    for (size_t i = 0; i < xf.count; i++) {
+        if (i == 0 || i == xf.count - 1) {
+            ConvertValue(mtx, xf.values[i]);
+        } else {
+            float time = xf.times[0] + i * (xf.times[xf.count-1] - xf.times[0]) /
+                (static_cast<float>(xf.count)-1.f);
+            
+            if (std::abs(xf.times[i] - time) < AI_EPSILON) {
```
cpichard, line 234: not sure why it is necessary to resample the values, a comment would be helpful

libs/translator/reader/utils.cpp
```diff
@@ -697,21 +698,21 @@ void ReadCameraShaders(const UsdPrim& prim, AtNode *node, UsdArnoldReaderContext
...
-    // The deform keys attribute takes precedence on any prims
-    if (UsdAttribute deformKeysAttr = prim.GetAttribute(_tokens->PrimvarsArnoldDeformKeys)) {
-        UsdGeomPrimvar primvar(deformKeysAttr);
-        if (primvar) {
-            int deformKeys = 0;
-            if (deformKeysAttr.Get(&deformKeys, UsdTimeCode(time.frame))) {
-                numKeys = deformKeys > 0 ? deformKeys : 1;
-            }
-            return numKeys;
+    const auto getNumKeys = [&](const UsdPrim &prim, const TfToken& attr, float frame) -> int {
```
cpichard, line 701: prim argument is not needed it's already passed as reference, and time.frame as well.

cpichard (approved):

sebastienblor:
applied review comments

cpichard (approved):

PR #2120, opened 2024-10-01
Support the shader_override render setting

### LINKED ISSUE #2119: Support shader_override
The options node supports the shader_override parm, which will override every material in the scene (which is super handy for debugging/benchmarking)

**Additional context**

TomMinor:
- Sets shader_override on the options node if **arnold:global:shader_override** exists on the render settings
- _WIP - this doesn't work yet as the material doesn't exist to assign unless there's an existing material binding_

Fixes #2119

**Additional context**


sebastienblor:
I included the code from this PR in #2136 , so we can close this one

PR #2130, opened 2024-10-16
Fix crashes with instancers that have empty / invalid points #2129

### LINKED ISSUE #2129: Crashes with instancer empty values
We have a repro scene, where some instancer have empty / invalid values. It ends up crashing at render time, with the usd translator (hydra is fine)

sebastienblor:
The instancer translator needs to ensure it's not trying to access invalid pointers when the amount of positions is empty. We just need to do some out-of-bounds checks.
Adding test_2129 with the repro scene

Fixes #2129


MasterZap (approved):
Looks good to me

PR #2136, opened 2024-10-30
Support options shader_override

### LINKED ISSUE #2119: Support shader_override

sebastienblor:
Adding support for `options.shader_override` as it's already done for background, atmosphere, imagers, etc...
The usd writer needs to author an ArnoldNodeGraph pointing to the expected shader. Both the usd reader and the render delegate must look for such a nodegraph for this attribute.
Added test_2119 

Fixes #2119

**Additional context**


cpichard:
test_0174 is not passing on the CI, do you have the same issue ?

sebastienblor:
indeed, fixed

cpichard (approved):

PR #2175, opened 2025-01-09
Author unassigned shaders under the mtl scope

### LINKED ISSUE #2174: Unassigned arnold shaders written to USD should go under the mtl scope
Currently when shaders need to be authored to USD without a material binding, they are saved at the root of the stage. They should at least go under the "mtl" scope

sebastienblor:
When we author unassigned shaders to usd, we want them to go under the mtl scope. So we need to apply a similar code to what is already done when shaders are assigned under a material. At the moment we only want to do that for actual arnold shaders, and not the other node types which are expressed as UsdShade
Adding test_2174

Fixes #2174


cpichard:
Some tests are failing on my machine and the CI :
Screenshot 2025-01-09 at 09 46 17
Screenshot 2025-01-09 at 09 46 10
is it working on windows ?

PR #2182, opened 2025-01-10
Stop using deprecated functions in the testsuite

sebastienblor:
Several cpp tests are calling functions that are marked as deprecated. This PR fixes them


testsuite/test_0148/data/test.cpp
```diff
@@ -7,14 +7,14 @@ int main(int argc, char **argv)
 {
     AiMsgSetConsoleFlags(nullptr, AI_LOG_ALL);
     AiBegin();
-    AiSceneLoad(nullptr, "scene.ass", nullptr);
-    AiSceneWrite(nullptr, "scene_exported.usda", nullptr);
+    AiSceneLoad(nullptr, AtString("scene.ass"), nullptr);
```
ThiagoIze, line 10: AiSceneLoad/Write take a `const char* filename` so not sure there is any benefit to use an AtString like this

cpichard (approved):

PR #2192, opened 2025-01-15
Fix build issue with schemas in MaxUsd

sebastienblor:
In the MaxUSD sdk, the resource file pluginInfo.json is not under lib/usd/, but under plugin/usd. Because of that, enabling the build of the schemas is failing the compilation because we try to copy this file that doesn't exist


cpichard (approved):

sebastienblor:
closing as this doesn't seem to be needed at the moment

PR #2215, opened 2025-02-03
Fix incorrect reset of mesh parameters

### LINKED ISSUE #2214: Error AiArrayCopy: source array is null
**Describe the bug**
Using the latest commit of arnold-usd, 9b0ec0d5e7d41ce8a96863600465efcbefc29d8b, 
when deallocating the render delegate the console shows the following error:

ERROR   | AiArrayCopy: source array is null
ERROR   | AiArrayCopy: source array is null
ERROR   | AiArrayCopy: source array is null
ERROR   | AiArrayCopy: source array is null
ERROR   | AiArrayCopy: source array is null

cpichard:
- Since we share the mesh parameters we have to reset them when the HdMesh is destroyed. The previous way of doing it was to set the parameter array to nullptr which is incorrect. This PR uses the AiNodeResetParameter function to correctly reset the parameter.

Fixes #2214

### DIFF
--- a/libs/render_delegate/mesh.cpp
+++ b/libs/render_delegate/mesh.cpp
@@ -181,12 +181,12 @@ HdArnoldMesh::~HdArnoldMesh() {
     // be pointing to deallocated memory.
     AtNode *node = GetArnoldNode();
     if (node && !_arrayHandler.empty()) {
-        AiNodeSetArray(node, str::nsides, nullptr);
-        AiNodeSetArray(node, str::vidxs, nullptr);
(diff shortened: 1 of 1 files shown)


sebastienblor:
I first thought that we shouldn't have to reset the attribute since the node is being deleted immediately after (when HdArnoldRprim is deleted with its HdArnoldShape). But I realize the AtNode destruction triggering the AtArrays destruction would happen too late as the shared arrays would already be deleted.

sebastienblor (approved):

PR #2233, opened 2025-02-10
Fix left handed indexed meshes with normals 

### LINKED ISSUE #2232: Incorrect render of meshes with indexed normals
**Describe the bug**
Since the commit 60e66769d5898828ed74b612d25608e513eb4dc2 the meshed with indexed normals are not rendered correctly.
Here is a test scene that shows the issue :
[test.usd_rop1.usd.zip](https://github.com/user-attachments/files/18732923/test.usd_rop1.usd.zip)

cpichard:
When rendering left handed indexed meshes with normals with husk, the sync function is called twice and the second time the leftHanded variable is not correctly set which ends up having the normal indices incorrectly set. 

Fixes #2232

### DIFF
--- a/libs/render_delegate/mesh.cpp
+++ b/libs/render_delegate/mesh.cpp
@@ -204,7 +204,6 @@ void HdArnoldMesh::Sync(
     HdArnoldRenderParamInterrupt param(renderParam);
     const auto& id = GetId();
     HdArnoldSampledPrimvarType _pointsSample;
-    VtValue vertexCountsVtValue;     ///< Vertex nsides
     const auto dirtyPrimvars = HdArnoldGetComputedPrimvars(sceneDelegate, id, *dirtyBits, _primvars, nullptr, &_pointsSample) ||
                                (*dirtyBits & HdChangeTracker::DirtyPrimvar);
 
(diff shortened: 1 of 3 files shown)


libs/render_delegate/mesh.cpp
```diff
@@ -235,15 +234,13 @@ void HdArnoldMesh::Sync(
         }
     }
 
-    bool isLeftHanded = false;
+    // We have to flip the orientation if it's left handed.
+    const auto topology = GetMeshTopology(sceneDelegate);
+    bool isLeftHanded = topology.GetOrientation() == PxOsdOpenSubdivTokens->leftHanded;
```
sebastienblor, line 239: I'm a bit worried of moving the GetMeshTopology outside of the "if (dirtyTopology)" as this function seems to have a cost, and we can be called multiple times in Sync even when there's nothing to change.

I think we should either store a bool member _isLeftHanded in the class, of call GetMeshTopology again inside of "if (dirtyPrimvars)", of create an empty HdMeshTopology and set it only "if (dirtyPrimvars || dirtyTopology)" so that the topology can be used inside of both tests, but the more we can remove from outside of the dirtiness tests, the better

PR #2244, opened 2025-02-18
Add build variable for TESTSUITE_PASSES

sebastienblor:
As we have in the arnold testsuite, I'm adding the build variable TESTSUITE_PASSES that can be used to customize it and just render "usd" or "hydra" test. This can be useful for debugging a specific issue and only running one of those tests


cpichard:
Cool ! this will be really useful. Should we have a line in the changelog even though it's for the testsuite ?

sebastienblor:
I wasn't planning to add it since it's just the testsuite. Do you think we should put it ?

cpichard (approved):

PR #2250, opened 2025-02-21
Enable the hydra mode of the procedural by default

### LINKED ISSUE #2248: USD Procedural should now default to the hydra mode
We want to change the default behaviour of the usd procedural to be in hydra mode by default.
An options attribute will be used to roll back to the previous default

sebastienblor:
This PR enables the hydra mode of the usd procedural by default. It also looks for an options attribute "usd_legacy_translation" (to be added in the arnold side) that can force the hydra mode to be disabled. 

The environment variables always take precedence on all these settings

Fixes #2248

### DIFF
--- a/plugins/procedural/main.cpp
+++ b/plugins/procedural/main.cpp
@@ -53,19 +53,31 @@
 static std::unordered_map<AtNode*, ProceduralReader*> s_readers;
 static std::mutex s_readersMutex;
 
-inline ProceduralReader *CreateProceduralReader(AtUniverse *universe, bool hydra = false, AtNode* procParent = nullptr)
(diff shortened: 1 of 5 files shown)


plugins/procedural/main.cpp
```diff
@@ -89,7 +101,7 @@ node_parameters
     AiParameterArray("overrides", AiArray(0, 1, AI_TYPE_STRING));
     AiParameterInt("cache_id", 0);
     AiParameterBool("interactive", false);
-    AiParameterBool("hydra", false);
+    AiParameterBool("hydra", true);
```
cpichard, line 104: may be we could get rid of this parameter ?

cpichard (approved):

PR #2251, opened 2025-02-28
Incorrect transform when exporting parented objects to USD

### LINKED ISSUE #1402: Incorrect transform when exporting an object hierarchy to USD
When exporting objects from an Arnold scene via `AiSceneWrite()`, the transform is wrong, if the object hierarchy is coded into the node names.

Arnold does not support object hierarchy, in an Arnold scene all objects are on the same level. When we export objects from DCCs which support hierarchy, this information is encoded into the node name. For instance if a Pyramid is under a Cube object in Cinema 4D, the name of the Pyramid will be '/Cube/Pyramid'. When exporting to USD, it recognizes the hierarchy, and puts the Pyramid object under the Cube, but it does not change the transform to relative to the Cube and still uses the global transform. That's causing the problem.

These are ASS and USD exports of the same scene.
ASS file: [object_hierarchy.ass.zip](https://github.com/Autodesk/arnold-usd/files/10499356/object_hierarchy.ass.zip)
USD file: [object_hierarchy.usda.zip](https://github.com/Autodesk/arnold-usd/files/10499359/object_hierarchy.usda.zip)

To verify, load the ASS and USD files with a procedural.

| ASS | USD | 
| :---: | :---: |
| [scene_ASS.zip](https://github.com/Autodesk/arnold-usd/files/10499401/scene_ASS.zip) | [scene_USD.zip](https://github.com/Autodesk/arnold-usd/files/10499404/scene_USD.zip) |
| scene_ASS | scene_USD |
| correct | incorrect |

sebastienblor:
When we convert an arnold scene to usd, all the matrices are in world space since there is no transform hierarchy in arnold. 
Thus, when we see that some output primitives are parented under other primitives, we need to apply the inverse matrix of the parent, in order to get the expected result when the scene will be rendered

adding test_1402

Fixes #1402


libs/translator/writer/prim_writer.cpp
```diff
@@ -1118,6 +1118,27 @@ bool UsdArnoldPrimWriter::WriteAttribute(
 void UsdArnoldPrimWriter::_WriteMatrix(UsdGeomXformable& xformable, const AtNode* node, UsdArnoldWriter& writer)
 {
     _exportedAttrs.insert("matrix");
+
+    UsdPrim &prim = xformable.GetPrim();
```
cpichard, line 1122: the CI shows the following issue:
```
libs/translator/writer/prim_writer.cpp:1122:14: error: non-const lvalue reference to type 'arnoldPxrInternal__pxrReserved__::UsdPrim' cannot bind to a temporary of type 'arnoldPxrInternal__pxrReserved__::UsdPrim'
    UsdPrim &prim = xformable.GetPrim();
```

cpichard (approved):

PR #2267, opened 2025-04-11
Fix velocity motion blur with varying number of instances

### LINKED ISSUE #2231: Incorrect motion blur with varying amount of instances
**Describe the bug**
For each instance created by an Instancer LOP, the prototype also shows up in the render.
The attached scene has this setup:

- point instancer with changing point count per frame
- velocity blur (because of changing point count)
- pscale set on points with a marked scale difference from prototypes
- multiple time samples exists (whether with a cache node or when written to usd file)

If you write out a usd file and then load it as a sublayer, it has the same problem.
Note that Karma didn't render the sublayer properly either.

The viewport renders were a bit erratic. As I played around with the scene, the motion blur would disappear sometimes. Then come back. Then the problem would go away and I had to re-load the scene to repro again.

**Repro data set**

[support_case_00031665.zip](https://github.com/user-attachments/files/18711446/support_case_00031665.zip)

**Screenshots**
If you increase the shutter size, you can clearly see it's the prototype torus that renders:

Image

Sometimes, not always, if you add an Arnold Render Settings LOP and enable Ignore Motion Blur, you also see the prototype torii:

Image

In a new Houdini scene, with the usda loaded as a Sublayer, this is what Karma rendered (in the original scene, the Karma render was ok):

Image

**Used Software Versions**
 - Arnold: Latest
- Houdini

### LINKED ISSUE #2285: Point instancer angular velocity not taken into account in the render delegate
**Describe the bug**
The angular velocity which can be set on instances via the point instancer is not used in the render delegate. The procedural does use it. One of our example using angular velocity is test_0024 and extracting single instances from it shows motion blur differences between the render delegate and the procedural.

**Used Software Versions**
 - Arnold: 7.4.1.0
 - USD: 24.11

cpichard:
- In case of velocity motion blur we now sample all the primvars at frame time. Previously we were sampling only the translation at the frame, apply the velocity and sample the other transforms at the remaining sampling times. However the velocity blur is used a lot in particle systems when instances are created and killed, which can lead to problems when sampling different times, it is possible that the transforms don't have the same number of values (instances) and this generally brings an incoherent motion blur and/or potential memory errors.
- We also take into account angular velocity which was missing the render delegate.
- This PR does fix the problem with varying instances only if the user adds velocity. If there are varying instances with no velocity then the motion blur is undefined and will probably end up in rendering issues. 
- It is now recommended to use velocity and angular velocity for motion blur when using varying instances.

Fixes #2231
Fixes #2285


sebastienblor (approved):
a diff_hardfail higher than 0.1 is a bit too much, we could miss future issues, but still better to have this new test

PR #2297, opened 2025-04-30
Fix stats mode in the render delegate

### LINKED ISSUE #2296: Cannot append stats
**Describe the bug**
If you set the Stats Mode to Append, it doesn't work. 

**To Reproduce**
1. In the Arnold Render Settings LOP, on the Diagnostics tab, set Stats Mode to Append.
2. Render a bunch of frames, or do a bunch of IPR renders.
3. Stats are not appended.

husk and kick both have the problem.

**Used Software Versions**
 - Arnold: 7.4.1

**Additional context**

stats mode is set twice:

https://github.com/Autodesk/arnold-usd/blob/d9c32fcb220c015b42721f9a728bfd4c80929249/libs/render_delegate/render_delegate.cpp#L711C11-L712C44

sebastienblor:
The stats mode was not being set properly in the render delegate. It was always forced to be "overwrite".
This PR takes into account the provided value, and it also supports string values for this setting, which is what HtoA returns us.

Fixes #2296


cpichard (commented):
should we have a changelog entry for this ?

cpichard (approved):

PR #2367, opened 2025-07-15
Fix husk animated renders with arnold product type

### LINKED ISSUE #2349: "arnold" product type with husk overwrites the same output path when rendering mutliple frames in same process
### Bug Description
When rendering a usd file with husk (multiple frames in the same process) that contains a renderproduct with the type "arnold", only one output file is generated, that is getting overwritten after each frame is done rendering.

Expected output:
render.1001.exr
render.1002.exr
render.1003.exr
render.1004.exr
render.1005.exr

Actual output:
render.1001.exr (overwritten 5 times)

### To Reproduce
In Houdini:
1. Add render product with "arnold" type
2. Export usd file and render with husk cli command:
`"C:\Program Files\Side Effects Software\Houdini 20.5.550\bin\husk" --usd-input "usd.usd"  -f 1001 -n 5 -i 1 -R HdArnoldRendererPlugin`
  OR 
3. use usd_render rop with option: "Render all frames with a single process" turned on

### Screenshots 
Arnold product type:
Image

Confirming productName is time varying:
Image

### Used Software Versions
 - Arnold: 7.4.1.0
 - HtoA: 6.4.1.0
 - Houdini 20.5.550

**Additional context**
[example_hip_and_usd_file.zip](https://github.com/user-attachments/files/21103676/example_hip_and_usd_file.zip)

sebastienblor:
We need to ensure the render delegate products are updated at every render.
So instead of doing an early out at the beginning of `_ParseDelegateRenderProducts` (which is called for every new frame), I'm now clearing the previous render delegate products, and regenerating them.

However, since these functions handling the drivers can be called multiple times in a row, I also need to ensure we don't recreate the previous drivers and filters that were created for the previous frame. Otherwise there will be a naming conflict, and they will be given an empty name in the arnold universe, which will prevent from having proper outputs. So I'm adding an optional boolean to `CreateArnoldNode` that will first check if the requested node already exists, and if it does it just returns it.

Fixes #2349

### DIFF
--- a/libs/render_delegate/render_delegate.h
+++ b/libs/render_delegate/render_delegate.h
@@ -585,12 +585,8 @@ public:
     /// Method used to create any node in the context of the render delegate. 
     /// This method should always be called, instead of explicit AiNode() creations
     inline 
-    AtNode * CreateArnoldNode(const AtString &nodeType, const AtString &nodeName, bool checkExisting = false) {
-        AtNode *node = checkExisting ? LookupNode(nodeName.c_str()) : nullptr;
-        if (node != nullptr && AiNodeIs(node, nodeType))
-            return node;
-
-        node = AiNode(GetUniverse(), nodeType, nodeName, _procParent);
+    AtNode* CreateArnoldNode(const AtString &nodeType, const AtString &nodeName) {
+        AtNode* node = AiNode(GetUniverse(), nodeType, nodeName, _procParent);
         if (_procParent) {
 
             // All shape nodes should have an id parameter if we're coming from a parent procedural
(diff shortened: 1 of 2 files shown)


libs/render_delegate/render_delegate.h
```diff
@@ -573,8 +585,12 @@ class HdArnoldRenderDelegate final : public HdRenderDelegate {
     /// Method used to create any node in the context of the render delegate. 
     /// This method should always be called, instead of explicit AiNode() creations
     inline 
-    AtNode * CreateArnoldNode(const AtString &nodeType, const AtString &nodeName) {
-        AtNode *node = AiNode(GetUniverse(), nodeType, nodeName, _procParent);
+    AtNode * CreateArnoldNode(const AtString &nodeType, const AtString &nodeName, bool checkExisting = false) {
```
cpichard, line 588: Can we have a function FindOrCreateArnoldNode instead of adding a `checkExisting` argument ?

cpichard (approved):

PR #2394, opened 2025-08-01
Support fieldIndex in VDBs

### LINKED ISSUE #2315: VDB fieldIndex isn't used
**Describe the bug**
When Solaris imports multiple VDB grids with the same field name (i.e. 3 density grids), it will set the fieldIndex on the prim so it can correctly look up the specific grid index. This doesn't work in Arnold.

**To Reproduce**
1. Create multiple density VDB primitives in a sop create with unique transforms (to visualize them better)
2. Render in the arnold delegate

**Expected behavior**
In the viewport, I see
Image

Delegate results:
Image

kick results:
Image

**Used Software Versions**
 - Arnold: [e.g. 5.4.0.2]
 - USD: [e.g. 0.19.11]
 - Compiler: [e.g. GCC 4.8.5]
 - OS: [e.g. MacOS 10.15]
 - Any 3rd-party app: [e.g. Katana-3.2]

**Additional context**

sebastienblor:
This PR includes the changes previously pushed in #2316 that were reverted. It also includes a support of fieldIndex in regular vdb translation, both in usd and hydra. It also adds a test_2315 with that use case. 

It's marked as a draft as I still don't have something satisfying, at the moment I'm only supporting the field index if it is > 0 which is not really the intended behaviour. Problem is that the default value of fieldIndex is 0, so if we always take it into account it would end up changing the code path for all vdb renders, in a way that would break existing scenes.

Fixes #2315


testsuite/test_2239/data/test.usda
```diff
@@ -26,6 +26,7 @@ def Xform "adsktest" (
     )
     {
         custom rel field:density = </adsktest/testdata/density>
+        rel material:binding = </mtl/material>
```
cpichard: When I run the usd checker I got: 
`Found material bindings but no MaterialBindingAPI applied on the prim </adsktest/testdata>`

PR #2398, opened 2025-08-08
MeshLightAPI support

### LINKED ISSUE #2390: Support MeshLightAPI with hydra2
When the MeshLightAPI is applied on a mesh, the mesh should behave like a light.

cpichard:
- add meshlight scene index for hydra2  
- add meshlight support in the procedural

Fixes #2390


libs/translator/reader/read_light.cpp
```diff
@@ -296,6 +226,86 @@ AtNode *_ReadLightShaping(const UsdPrim &prim, UsdArnoldReaderContext &context)
...
+    This is preventing distant lights from working properly, so we should only
+    do it where it makes sense VtValue normalizeAttr.
+    if(light.GetNormalizeAttr().Get(&normalizeAttr, time.frame))
+       AiNodeSetBool(node, "normalize", normalizeAttr.Get<bool>());
+    */
+    // Normalize attribute only for meshlights
+    if (AiNodeEntryGetNameAtString(AiNodeGetNodeEntry(node)) == str::mesh_light) {
+        VtValue normalize;
+        if (GET_LIGHT_ATTR(light, Normalize).Get(&normalize, time.frame)) {
+            AiNodeSetBool(node, str::normalize, VtValueGetBool(normalize));
```
sebastienblor, line 290: to be consistent with the existing code, this normalize attribute should be set from the mesh translator just before or after it calls `ReadLightCommon`. This way we're sure it's only set in that case and not e.g. if the mesh_light is created from another translator, like GeometryLightTranslator

PR #2408, opened 2025-08-22
Fix render region calculation when based on windowNDC 

sebastienblor:
When the render pass computes the render region based on windowNDC, the calculation needs to be more robust, to ensure the region corresponds to the render resolution that is requested.

Fixes HTOA-2970

### DIFF
--- a/libs/render_delegate/render_pass.cpp
+++ b/libs/render_delegate/render_pass.cpp
@@ -496,7 +496,35 @@ void HdArnoldRenderPass::_Execute(const HdRenderPassStateSharedPtr& renderPassSt
             if (windowNDC[1] > windowNDC[3])
                 std::swap(windowNDC[1], windowNDC[3]);
 
-                // we want the output render buffer to have a resolution equal to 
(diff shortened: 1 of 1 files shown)


libs/render_delegate/render_pass.cpp
```diff
@@ -548,8 +569,24 @@ void HdArnoldRenderPass::_Execute(const HdRenderPassStateSharedPtr& renderPassSt
             }
 
             // we want region_max_y - region_min_y to be equal to _height - 1
-            AiNodeSetInt(options, str::region_min_y, int(windowNDC[1] * _height));
-            AiNodeSetInt(options, str::region_max_y, int(windowNDC[3] * _height) - 1);
+            float region_min_y_flt = windowNDC[1] * (renderSettingsRes[1] > 0 ? renderSettingsRes[1] : _height);
```
cpichard, line 572: since it's the same computation as above we could use a function

cpichard (approved):
tested with htoa

Issue #2385, opened 2025-07-25
Cannot hide geo from camera ray in Houdini if USD already has anrold:visibility primvar (from Maya)


hollov:
**To Reproduce**
1. Create a cube in Maya, and disable e.g. "Casts Shadows" under Arnold-Visibility options in attribute editor
2. Export the cube from to USD with Arnold job context
3. In Houdini try to hide it from camera rays using "Render Geometry Settings" node: uncheck "Camera Ray Visibility" under Arnold tab - Visibility
4. In Arnold render you'll see the cube remained visible in camera

**Expected behavior**
I expect the visibility override to work so the cube should become invisible for camera rays.

What happens is during USD export in Maya the Mesh prim gets an arnold:visibility = 253 prim var.
Houdini creates an arnold:visibility:camera = False prim var. 
And it seems these are not merged, but arnold:visibility takes precedence. I'd expect it would result an arnold:visibility = 252

**Used Software Versions**
 - Maya 2024.2, mayausd-0.32.0_202505060957-eeb3aaa, mtoa-5.4.8
 - Houdini 20.0.590, htoa-6.3.7.0_r71394d1_houdini-20.0.896.py39

PR #2424, opened 2025-09-16
Add test_2385

sebastienblor:
Adding a test that exercises the issue described in #2385, with the `visibility` attribute set along with `visibility:camera`


cpichard:
it's failing on mac unfortunately

sebastienblor:
Now the PR is no longer about just a test, but has changes to fix the issue

PR #2436, opened 2025-09-25
Support cryptomatte with hydra2 

sebastienblor:

When instancers are rendered with hydra 2, and cryptomatte AOVs are enabled, we need to ensure the cryptomatte hashes will be consistent from frame to frame. By default they will vary as the prototype names aren't consistent anymore, and cryptomatte is based on node names. For that we use the "crypto_object" and "crypto_object_offset" user data that will be used by cryptomatte instead of the node names, and we ensure they're given a constant value.

In a previous PR the "crypto_object" user data was set on the prototype shape. But now, in order to have different crypto hashes per instance, we must set the user data "crypto_object_offset" on each instance. to do that, we can set the array "instance_crypto_object_offset" on the instancer node, that will then set the corresponding value on each instance.

Fixes HTOA-2999

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - [usd#2408](https://github.com/Autodesk/arnold-usd/issues/2408) - Fix render region calculation when based on windowNDC
 - [usd#2404](https://github.com/Autodesk/arnold-usd/issues/2404) - Support asset_searchpath as a replacement for procedural & texture search paths.
 - [usd#2407](https://github.com/Autodesk/arnold-usd/issues/2407) - Add support for RenderPass prim when using hydra2.
+- [usd#2436](https://github.com/Autodesk/arnold-usd/issues/2436) - Support cryptomatte with instances in hydra2
 
 ### Bug Fixes
 


sebastienblor:
Note that we could have chosen to always set the user data "instance_crypto_object_offset" on the instancer. It would have been simpler, but would have added an additional cost even when cryptomatte isn't used. Here we want to only add this when cryptomatte aovs are enabled, which is a bit tricky because we can't be sure of the order in which translation will happen between the instancer and the aovs. 
So when a cryptomatte aov is detected in render_pass.cpp we raise a flag "hasCryptomatte" and loop through all the instancers that were previously translated to set their user data.
And if new instancers get translated later on, they will see that the render delegate has cryptomatte aovs, and will therefore add this data on themselves. This way the result is independent of the order of translation.

Also, we're not trying to detect if during IPR sessions, cryptomatte AOVs are removed from the options. In that case we could eventually remove this user data from all instancers in the scene. But this would complicate the code even more, and it doesn't seem necessary as it would just be a small optimization for this specific use case.

cpichard (approved):

PR #2502, opened 2025-12-17
Fix incorrect transforms and materials when authoring usd files with meshes under instances

### LINKED ISSUE #2501: Wrong matrices when authoring usd files with meshes under ArnoldGinstances hierarchy
When converting the attached .ass file to usd, we end up with incorrect transforms. The meshes under the ArnoldGinstance primitives don't take into account the matrix of their parent

sebastienblor:
In the case of a mesh under another shape, if the leaf mesh has an empty matrix (=identity), we were skipping the process of applying the parent inverse matrix, assuming it wasn't needed. Now we don't skip it if there is a parent matrix to apply.

Also, for material bindings, before we create a new material we first check if a shape already has a material binding (e.g. for mayaUSD exports it lets us reuse the materials authored by mayaUSD). But the function we were calling was taking into account the hierarchy, so in the case described in this ticket we would get the material from the parent mesh. So the shader from the leaf mesh was being authored in the same material as the parent, which was breaking the material assignments.

Adding test_2501 that exercises both issues

Fixes #2501

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@
 
 ### Bug Fixes
 
-- [usd#2501](https://github.com/Autodesk/arnold-usd/issues/2501) - Fix incorrect transforms when authoring usd files with meshes parented under instances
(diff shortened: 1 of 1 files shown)


ZapAndersson (approved):
LGTM

cpichard (approved):

PR #2508, opened 2026-01-06
Avoid using different proxy lists to fill the references

### LINKED ISSUE #2507: Crash on linux when collecting assets
**Describe the bug**
Calling AiSceneGetAssets on a USD scene crashes on linux.

cpichard:
- We were accessing begin and end of 2 different proxy lists which ended up crashing on linux. This PR fixes the problem.  
- We don't need to update the changelog since it's an ongoing development for a new arnold api

Fixes #2507


peterhorvath111 (approved):
Looks good.

PR #2539, opened 2026-02-04
Fix issues with render output nodes

sebastienblor:
The recent support for render output nodes was missing some changes to be properly working. 
- With render outputs it's easy to get nodes with `*` in their arnold node names, so our writer was crashing when trying to author these nodes to usd due to the invalid character
- the render output type wasn't properly ready (it's lower case "rgba" as opposed to upper case "RGBA" in the outputs string). Because of that, the test we added for render outputs wasn't rendering with an alpha channel as expected, so I had to update the ref
- some tests are saving out .ass files and checking specifically the outputs string. For this to work we need to set a param value so that arnold doesn't convert the outputs to render output nodes

**Additional context**


sebastienblor:
Since this is fixing an issue with an unreleased feature, I don't think it's needed to add it to the changelog

cpichard (approved):

PR #2551, opened 2026-02-10
Add COPs support

TomMinor:
-  In the material translation we need to detect `op:` paths and do something about it. If an `op:` path is detected, I create the original image node as normal but also create a HtoA `image_cop` node with the original image node as reference, so all the image attributes can be copied automatically. That way all of the Houdini library pain is avoided by arnold-usd and the Houdini business of dumping the pixel data is done for me. This should work for any Houdini version.
- As explained in the [HDK Hydra docs](https://www.sidefx.com/docs/hdk/_h_d_k__u_s_d_hydra.html#HDK_USDHydraCopTextures), we need to check for the `houdini:cop_texture_changed` render setting and update our texture cache. Flushing the universe texture cache does the trick for now but I would like to investigate the performance implications of this. This is only supported in Houdini 21+

**Additional context**
image

### DIFF
--- a/libs/common/materials_utils.cpp
+++ b/libs/common/materials_utils.cpp
@@ -420,17 +420,17 @@ AtNode* ReadArnoldShader(const std::string& nodeName, const TfToken& shaderId,
 #define HOUDINI_COPS_SUPPORT 1
 #ifdef HOUDINI_COPS_SUPPORT
     bool is_op_path = false;
-    std::string internalNodeName = nodeName;
     if (shaderId == str::t_image) {
         const auto filenameAttr = inputAttrs.find(str::t_filename);
         if (filenameAttr != inputAttrs.end() && !filenameAttr->second.value.IsEmpty()) {
             std::string filename = VtValueGetString(filenameAttr->second.value);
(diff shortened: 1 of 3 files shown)


libs/common/materials_utils.cpp
```diff
@@ -412,11 +412,31 @@ const ShaderReadFuncs& _ShaderReadFuncs()
...
+        const auto filenameAttr = inputAttrs.find(str::t_filename);
+        if (filenameAttr != inputAttrs.end() && !filenameAttr->second.value.IsEmpty()) {
+            std::string filename = VtValueGetString(filenameAttr->second.value);
+            if (filename.substr(0, 3) == "op:" ||
+                filename.find("opdef:") != std::string::npos) {
+                is_op_path = true;
+            }
+        }
+    }
+    const std::string& internalNodeName = nodeName;
```
cpichard, line 433: do we need internalNodeName, at first glance it seems that for both HOUDINI_COPS_SUPPORT path, the value is the same

cpichard:
We need a line in the changelog and a clang-format pass.

PR #2623, opened 2026-04-22
MTOA-2714: Keep primvars on drivers

compso:
**Check for and convert primvars to userdata for RenderProducts**
- 
-


sebastienblor:
as it's not actually needed, and since it's not a common use case, let's avoid adding dead code. We can close the PR now and eventually reopen if needed

PR #2493, opened 2025-12-03
Issue 2439 - UsdLux add cone and shaping attributes to non-sphere lights

libbyomeara:
**Changes proposed in this pull request**
- 
- 

**Issues fixed in this pull request**
Fixes #2439
Fixes #2447 

**Additional context**
Add any other context or screenshots about the pull request here.

PR #2624, opened 2026-04-23
Add support for shaping attributes with usdLux

sebastienblor:
This PR is identical to #2493 but updated with the latest changes. Also, for now a test is being temporarily disabled until it can be fixed in the arnold side.


cpichard (dismissed):

cpichard:
we need an entry in the changelog for this feature as well

PR #2630, opened 2026-05-06
Instance light linking scene index filter

cpichard:
- This adds a scene index filter splitting the instances by light groups. This allows light linking on instances.


sebastienblor:
let's first check if there is a solution in the render delegate side, so that the fix is not only for Houdini 21

PR #2651, opened 2026-05-20
Optionally use Hydra Render Settings primitive instead of settings map 

### LINKED ISSUE #2650: Optionally use Hydra Render Settings primitive instead of render settings map
We want to support hydra2 render settings when they're requested by the host. Currently, only MayaHydra can pass the render settings through a render settings primitive, so we want to trigger it depending on specific settings

sebastienblor:

This PR removes the ifdefs around ENABLE_HYDRA2_RENDERSETTINGS and instead makes it an optional code path triggered by a specific value `renderSettingsSrc` being set to `hydraSceneRenderSettingsSrc`. This setting is specific to MayaHydra which is currently the only tool expecting render delegates to use the render settings primitive, but it can be extended later on to other tools if needed 

Fixes #2650


sebastienblor:
Here I'm making it so that the hydra2 testsuite uses the new code path

sebastienblor:
passes the automation

cpichard (approved):

PR #2652, opened 2026-05-20
Use GPU textures for viewport rendering

sebastienblor:
This PR returns GPU textures to Hydra through `HdRenderBuffer::GetResource` . We get them through the new Arnold API `AiGetRenderOutput`. The render setting is called accelerated_viewport


cpichard:
:white_check_mark: CI passed (build 19).

cpichard:
I am testing on windows and having some issues at runtime, so getting to the bottom of them before approving the PR.
`render_delegate.lib(render_buffer.obj) : error LNK2019: unresolved external symbol "__declspec(dllimport) void (__cdecl* pxrInternal_v0_25_5__pxrReserved__::internal::GLApi::glGetIntegerv)(unsigned int,int *)" (__imp_?glGetIntegerv@GLApi@internal@pxrInternal_v0_25_5__pxrReserved__@@3P6AXIPEAH@ZEA) referenced in function "public: bool __cdecl pxrInternal_v0_25_5__pxrReserved__::`anonymous namespace'::_GpuFlipBlit::Flip(unsigned int,unsigned int,int,int)" (?Flip@_GpuFlipBlit@?A0x8179f035@pxrInternal_v0_25_5__pxrReserved__@@QEAA_NIIHH@Z)`

sebastienblor:
we were missing a dependency for Cmake, it should be better now

cpichard (commented):
Hit a link failure building this branch through HtoA on Windows (Houdini 21.0.631, shared USD). `usd_proc.dll` fails with 40 unresolved externals, all garch `GLApi` function pointers:

```
render_delegate.lib(render_buffer.obj) : error LNK2019: unresolved external symbol
  ... pxrInternal_v0_25_5__pxrReserved__::internal::GLApi::glGetUniformLocation ...
  referenced in function ...::_GpuFlipBlit::Flip(unsigned int,unsigned int,int,int)
```

(same for `glLinkProgram`, `glUseProgram`, `glViewport`, `glShaderSource`, `glUniform1i`, `glVertexAttribPointer`, `glIsEnabled`, …)

### Cause

The new accelerated-viewport path in `libs/render_delegate/render_buffer.cpp` (`_GpuFlipBlit`, `garch/glApi.h`, `HgiGLTexture`) makes `render_delegate` depend on `garch` / `hgiGL` / `hgi`. Those deps were added for two of the three consumers, but not the third:

| Consumer of `render_delegate` | Build file | Updated |
|---|---|---|
| `hdArnold` plugin (CMake) | `plugins/render_delegate/CMakeLists.txt` | yes |
| `hdArnold` plugin (SCons) | `tools/utils/dependencies.py:95-97` | yes |
| **`usd_proc` procedural (SCons)** | `plugins/procedural/SConscript` | **no** |

`plugins/procedural/SConscript` keeps its own hardcoded `usd_libs` list for the shared-USD branch and links `render_delegate`, so it pulls in `render_buffer.obj` without the new libraries. Only shared-USD builds are affected — under `static`/`monolithic` the symbols come from `libusd_m`.

### Suggested fix

In the shared-libs branch of `plugins/procedural/SConscript`:

```python
    # The accelerated viewport code path (SUPPORT_ACCELERATED_VIEWPORT) in
    # render_delegate references Hgi/HgiGL and garch's GLApi, so the procedural
    # needs those too when it embeds the render delegate.
    if local_env['ENABLE_HYDRA_IN_USD_PROCEDURAL'] and local_env['USD_HAS_HGI_GL']:
        usd_libs += ['hgi', 'hgiGL', 'garch']
```

With that, the full HtoA build completes.

### Minor, separate point

`USD_HAS_HGI_GL` is `PUBLIC` in CMake (`libs/render_delegate/CMakeLists.txt:74`) but only a private `CPPDEFINE` in SCons (`libs/render_delegate/SConscript:65-66`). Since `render_buffer.h:135,158` gates member variables on `SUPPORT_ACCELERATED_VIEWPORT`, any translation unit outside that library including the header would see a different `HdArnoldRenderBuffer` layout under SCons. Nothing outside `libs/render_delegate` includes it today, so this is latent rather than broken — but the two build systems disagreeing on the visibility of that define seems worth aligning.

sebastienblor:
applied the review changes

cpichard:
:x: CI failed (build 22).

cpichard:
CI build 22 failed (all platforms). Dug into the logs — three distinct causes, all traced back to the accelerated-viewport (GPU texture) changes in this PR:

**1. Linux — link error (all matrix builds)**
```
undefined reference to symbol 'pthread_join@@GLIBC_2.2.5'
/lib64/libpthread.so.0: error adding symbols: DSO missing from command line
```
`tools/utils/dependencies.py` now appends `hgi`, `hgiGL`, `garch` to `usd_libs` for non-static USD builds. That reorders the SCons link line so `libtbb.a` (which pulls in `pthread_join`) is no longer followed by an explicit `-lpthread`, and the linker drops it.

**2. Windows — link error (all matrix builds)**
```
LNK2019: unresolved external symbol __imp_SymInitialize / SymFromAddr / gethostname / OpenProcessToken / ...
```
Same root cause: linking `garch`/`hgiGL` pulls in USD's Arch stack-trace/file-access code, which needs `Dbghelp.lib` and `Advapi32.lib` — not currently added to the Windows link libs.

**3. macOS — real test regression, `test_2719`**
```
FAILED: HdArnoldRenderDelegate's paused/stopped state must survive/clear correctly around scene edits, Resume() and Stop()
```
This build succeeded, so this is an actual logic bug: `render_pass.cpp` now calls `renderParam->Interrupt(true, false)` whenever `acceleratedViewportChanged` is true (in addition to `framingChanged`), which appears to clobber the paused/stopped state instead of preserving it when accelerated viewport toggles.

`test_2719` looks new/tied to this feature, so the Linux link failure (can't even build that test) and the macOS logic failure both point at the same code path.

cpichard:
:x: CI failed (build 24).

cpichard:
:x: CI failed (build 24).

cpichard:
:x: CI failed (build 25).

cpichard (dismissed):

PR #2661, opened 2026-05-22
Fix #12: use Z AOV for depth and background detection

### LINKED ISSUE #12: [Render Delegate] Use Z depth to detect background pixels and handle pre-multiplication
At the moment the Render Delegate is using the P AOV to calculate the OpenGL depth and the ID AOV to differentiate between the background and foreground pixels.

The Z AOV should be better suited for both problems, as the P and ID AOV's behavior is inconsistent when using the GPU backend or mixing volumes and other objects.

cpichard:
Fixes #12.

## Summary

- Replace the **P AOV** with **Z AOV** for depth normalization — the P AOV is unreliable on the GPU backend and when mixing volumes with other geometry.
- Replace the **ID AOV** with a `z >= AI_BIG` threshold check for background pixel detection and alpha pre-multiplication — background pixels (ray misses) carry `Z == AI_BIG` consistently on both CPU and GPU backends and across all geometry types.
- Remove the `HYDRA_NORMALIZE_DEPTH` compile-time branch in `render_pass.cpp`; depth output is now always `Z FLOAT`.

## Changes

- `libs/render_delegate/render_pass.cpp` — always output `Z FLOAT` for depth (removed `#ifdef HYDRA_NORMALIZE_DEPTH` / `P VECTOR` path).
- `libs/render_delegate/nodes/driver_main.cpp` — remove `positionData` / P AOV branch; normalize Z to NDC `[0,1]` via `projMtx.Transform(GfVec3f(0,0,z))`; clamp background pixels to `depth = 1.0`; use `z >= AI_BIG` for alpha pre-multiplication instead of `ids[i] == -1`.

## Test plan

- [ ] Render a scene with CPU backend — verify depth and alpha are correct.
- [ ] Render the same scene with GPU backend — verify depth and alpha match CPU output.
- [ ] Render a scene mixing volumes and surfaces — verify background alpha is 0 and volume pixels are not incorrectly masked.

### DIFF
--- a/libs/render_delegate/nodes/driver_main.cpp
+++ b/libs/render_delegate/nodes/driver_main.cpp
@@ -156,11 +156,9 @@ driver_process_bucket
     int pixelType = AI_TYPE_RGBA;
     const void* bucketData = nullptr;
     const auto pixelCount = bucket_size_x * bucket_size_y;
-    // We should almost always have depth and id.
     auto& ids = driverData->ids[tid];
     ids.clear();
     const void* colorData = nullptr;
-    const void* positionData = nullptr;
     const void* depthData = nullptr;
     // Apply an offset to the pixel coordinates based on the region_min,
     // since we don't own the render buffer, which just knows the output resolution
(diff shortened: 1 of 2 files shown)


sebastienblor:
The work for this tickets has actually been mostly done a few months ago when we switched to Z AOV by default for Houdini. We don't want to remove the recent HYDRA_NORMALIZE_DEPTH build flag yet, we want to ensure that things are working properly in all DCCs, some of them expect a normalized depth.
The original ticket's description seems outdated : the ID AOV is used for selection, through each primitive's primId, and not just to distinguish the render from the background. So we can't really replace that with a Z AOV

cpichard:
:white_check_mark: CI passed (build 249).

cpichard:
Closing based on review feedback: the HYDRA_NORMALIZE_DEPTH flag should be kept until all DCCs are verified, and the ID AOV is needed for prim selection and cannot be replaced by a Z AOV check.

PR #2663, opened 2026-05-22
Various cleanup and fixes

sebastienblor:
List of simple fixes in the codebase : 

- DeregisterLightLinking(_shadowLink, this, false) should pass _lightLink — the surrounding if checks _lightLink and the false (non-shadow) flag confirms it. Result: the wrong link is deregistered on light-type change, leaking the original _lightLink registration
- HdArnoldSetRadiusFromPrimvar calls AiArrayMap(arr) but never AiArrayUnmap. Added the missing unmap before AiNodeSetArray.
- usdlux_version is read with AiNodeGetByte in a couple of places, it should be AiNodeGetInt
- TfStringStartsWith(name, str::arnold) matches a 6-char prefix ("arnold"), then substr(7) strips 7 chars — drops an extra character for any primvar like arnoldFoo that doesn't have the :. Switched to str::arnold_prefix ("arnold:") for both check and length.
- libs/translator/reader/read_skinning.cpp:440_ExtendWorldTransformTimeSamples walks the parent chain with p = p.GetParent(), but every iteration constructs UsdGeomXformable(prim) (original prim) instead of UsdGeomXformable(p). All ancestors get queried as the leaf prim — skinning time samples missed any animated parent xform. Changed prim → p.
- libs/translator/reader/utils.cpp:611ApplyParentMatrices branch for matrixNumKeys > parentMatrixNumKeys indexes AiArrayGetMtx(parentMatrices, i) with i up to matrixNumKeys-1, going out of bounds when the parent has fewer keys. Switched to AiArrayInterpolateMtx(parentMatrices, t, 0) with normalized t, matching the equal-key branch above it and the reference in common_utils.cpp.
- libs/render_delegate/render_settings.cpp:704Typo: this branch checks arnold:globals: (plural) and strips 15 chars. Every other token in the codebase uses arnold:global: (singular, 14 chars). Dead branch as written. Fixed prefix and length.

### DIFF
--- a/libs/render_delegate/light.cpp
+++ b/libs/render_delegate/light.cpp
@@ -179,7 +179,7 @@ void readUserData(
     HdArnoldGetPrimvars(delegate, id, dirtyBits, primvars, &interpolations);
     for (const auto &p : primvars) {
         // Get the parameter name, removing the arnold:prefix if any
-        std::string paramName(TfStringStartsWith(p.first.GetString(), str::arnold) ? p.first.GetString().substr(7) : p.first.GetString());
+        std::string paramName(TfStringStartsWith(p.first.GetString(), str::arnold_prefix.c_str()) ? p.first.GetString().substr(str::arnold_prefix.length()) : p.first.GetString());
         const auto* pentry = AiNodeEntryLookUpParameter(AiNodeGetNodeEntry(light), AtString(paramName.c_str()));
         if (pentry) {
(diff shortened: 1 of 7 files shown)


cpichard:
:x: CI failed (build #241).

cpichard:
:white_check_mark: CI passed (build 243).

cpichard (approved):

cpichard (approved):

PR #2664, opened 2026-05-22
More cleanup and fixes

sebastienblor:
- ApplyInstancerVisibilityToArnoldNode calls AiNodeSetInt(node, str::visibility, …) for the instancer's visibility attribute. Arnold's visibility is a BYTE parameter (8 ray-flag bits) — every other site in the file/folder uses AiNodeSetByte. The int setter writes to the wrong slot, so the instancer's visibility primvar is silently dropped on the Arnold node. Fixed to AiNodeSetByte with a uint8_t cast.
- When building the per-AOV layer_half_precision boolean array for driver_deepexr, the default fill value is read with AiNodeGetFlt(customProduct.driver, str::depth_half_precision). depth_half_precision is a BOOL parameter, so AiNodeGetFlt returns 0/garbage and the implicit float→bool conversion ends up false for every layer. Fixed to AiNodeGetBool.
- isOrtho only becomes true on the branch where the user supplied a render-pass camera (useOwnedCamera == false), and currentCamera = camera->GetCamera() is that user camera. But both if (currentCamera && isOrtho) blocks then set screen_window_min/max on _camera — the render-pass-owned perspective camera, which isn't even the active camera in this branch. The user's ortho camera never receives the computed screen window. Switched both blocks to currentCamera.
- HdArnoldSetTransform resamples to a user-provided transform_keys count by computing i * (timeEnd - timeStart) / (transformKeys - 1.f). For transformKeys == 1 the divisor is 0, so 0/0 produces NaN, xf.times[0] becomes NaN, xfOrig.Resample(NaN) is UB, and an Arnold matrix array of NaN entries is written. Added a transformKeys == 1 special case that places the single key at timeStart.
- When HasPendingChanges consumes _dependencyTrackQueue, it calls _dependencyToDirtyBitsMap.insert({{target, source}, bits}) for each new dependency. std::unordered_map::insert is a no-op on an existing key, so if the same source re-registers the same target with different dirty bits (e.g. a material adds another DirtyResource→DirtyMaterialId pairing), the new bits are silently dropped and the dependency dirties the source with stale bits. Switched to []= so the bits are overwritten
- HdArnoldGenericLight::Sync resolves a non-empty instancerId via renderIndex.GetInstancer(instancerId) and then unconditionally calls instancer->Sync(…) / instancer->CreateArnoldInstancer(…). Hydra can legitimately return nullptr (the instancer prim hasn't been registered yet, or it's a non-HdArnoldInstancer type), so the dereference is a null deref / wrong-type access. The sibling code in shape.cpp:171-173 already null-checks the same call. Wrapped the whole instancer setup in if (instancer != nullptr) { … }.
- HdArnoldSetNormalsFromPrimvar builds the array purely via AiArrayAllocate + AiArraySetKey (no AiArrayMap ever) but then calls AiArrayUnmap(arr). Unmapping an array that was never mapped is an API misuse — sister functions (HdArnoldSetRadiusFromPrimvar, ExtrapolatePositions, HdArnoldGetShidxs) all map first and only those call Unmap. Removed the stray AiArrayUnmap.
- GetRenderSetting has two else if (key == str::t_profile_file) branches in the same if/else if chain. The first (line 1071) returns the cached _profileFile; the second (line 1085) tries to return AiProfileGetFileName() and is unreachable. Removed the dead branch — clearly a stale duplicate (the same chain pattern exists for stats_file/report_file without an Arnold-side getter).
- HydraArnoldReader::SetCameraForSampling declares double shutterOpen, shutterClose; uninitialized, then calls Get<double>(&shutter…, timeCode). UsdAttribute::Get() returns false and leaves the destination untouched when the attribute has no authored or fallback value at that time. The return value is ignored, so _shutter = GfVec2f(shutterOpen, shutterClose) may pull from uninitialized stack — undefined behavior, and the bogus shutter then propagates into Arnold's options. Initialized both to 0.0.


cpichard:
:white_check_mark: CI passed (build 248).

cpichard (dismissed):

PR #2667, opened 2026-05-26
Cleanup & fixes in libs/common

sebastienblor:


libs/common/api_adapter.h
```diff
@@ -75,9 +75,11 @@ class ArnoldAPIAdapter {
             while (std::getline(ss, token, ' ')) {
                 AtNode *target = LookupTargetNode(token.c_str(), connection.sourceNode, connection.type);
                 if (target == nullptr)
-                    return false; // node is missing, we don't process the connection                
+                    return false; // node is missing, we don't process the connection
                 vecNodes.push_back(target);
             }
+            if (vecNodes.empty())
+                return false;
```
sebastienblor, line 82: An empty target string (or one that tokenises to nothing) leaves
vecNodes empty; &vecNodes[0] would then be UB. There is also no
meaningful array to assign, so bail out — the parameter keeps its
current value.

libs/common/api_adapter.h
```diff
@@ -91,6 +93,9 @@ class ArnoldAPIAdapter {
                     std::stringstream ss(connection.sourceAttr);
                     std::string arrayAttr, arrayIndexStr;
                     if (std::getline(ss, arrayAttr, '[') && std::getline(ss, arrayIndexStr, ']')) {
+                        if (arrayIndexStr.empty() ||
+                            arrayIndexStr.find_first_not_of("0123456789") != std::string::npos)
+                            return false;
```
sebastienblor, line 98: Guard the stoi: a malformed sourceAttr like "ramp_colors[abc]" would otherwise throw
std::invalid_argument. Skip processing this connection on a non-numeric index instead.

libs/common/api_adapter.h
```diff
@@ -99,7 +104,7 @@ class ArnoldAPIAdapter {
                             for (unsigned i=0; i<(unsigned) arrayIndex; i++)
                                 AiArraySetPtr(array, i, nullptr);
                             AiArraySetPtr(array, arrayIndex, (void *) target);
-                            AiNodeSetArray(connection.sourceNode, AtString(connection.sourceAttr.c_str()), array);
+                            AiNodeSetArray(connection.sourceNode, AtString(arrayAttr.c_str()), array);
                         }
```
sebastienblor, line 108: AiNodeSetArray takes the bare attribute name; passing sourceAttr (which still ends in "[N]") routes the call
 to a non-existent parameter, silently dropping the array and leaking it. 
Use arrayAttr, which is the substring before the bracket.

libs/common/common_bits.h
```diff
@@ -18,6 +18,8 @@
 /// @file common_bits.h
 ///
 /// Common bits for Hydra and Usd Imaging Adapters.
+#pragma once
+
```
sebastienblor, line 22: // Every other header in libs/common/ uses #pragma once. Without it, including this header from two paths in
// the same translation unit would redefine ArnoldUsdRprimBits.

libs/common/materials_utils.cpp
```diff
@@ -685,11 +697,12 @@ AtNode* ReadShader(const std::string& nodeName, const TfToken& shaderId,
         return nullptr;
 
     // First, we check if the shaderId starts with "arnold:", in which case
-    // we're expecting to read an arnold native shader with a 1:1 mapping 
+    // we're expecting to read an arnold native shader with a 1:1 mapping
     if (TfStringStartsWith(shaderId.GetString(), str::t_arnold_prefix.GetString())) {
-        TfToken arnoldShaderType(shaderId.GetString().substr(7).c_str());
+        TfToken arnoldShaderType(
+            shaderId.GetString().substr(str::t_arnold_prefix.GetString().size()).c_str());
```
sebastienblor, line 703: Derive the strip length from the prefix itself rather than hardcoding
7, which would leave a silent failure mode if str::t_arnold_prefix is ever changed.

libs/common/materials_utils.cpp
```diff
@@ -730,13 +743,16 @@ AtNode* ReadShader(const std::string& nodeName, const TfToken& shaderId,
 #endif
     if (shaderNodeEntry) {
         AtString shaderNodeEntryName = AiNodeEntryGetNameAtString(shaderNodeEntry);
+        AtNode* result = nullptr;
```
sebastienblor, line 746: `params` is allocated by AiParamValueMap() above and must be freed
before we return on every path — both successful and failing.
Previously, only the fall-through nullptr return at the end did so,
so every successful materialx shader translation leaked the map.

libs/common/parameters_utils.cpp
```diff
@@ -396,6 +396,9 @@ void ReadArrayLink(
 
     indexStr = indexStr.substr(1); // remove the first "i" character
 
+    if (indexStr.find_first_not_of("0123456789") != std::string::npos)
+        return;
+
```
sebastienblor, line 401: The caller only checks that the basename starts with 'i', so the
remainder can be any string (e.g. "nput" from "input", "nteger" from
"integer"). std::stoi would throw std::invalid_argument on those
perfectly legitimate Arnold attribute names. Require the remainder to
be purely digits before parsing.

libs/common/parameters_utils.cpp
```diff
@@ -730,7 +733,7 @@ AtMatrix VtValueGetMatrix(const VtValue& value)
 {
     if (value.IsEmpty())
         return AiM4Identity();
-    AtMatrix result;
+    AtMatrix result = AiM4Identity();
```
sebastienblor, line 736: Initialize to identity so that if the VtValue holds neither GfMatrix4f
nor GfMatrix4d, _VtValueGetRecursive leaves `result` alone and we return
a defined value instead of an uninitialized AtMatrix.

libs/common/parameters_utils.cpp
```diff
@@ -780,7 +783,8 @@ inline AtArray *_VtValueGetArray(const std::vector<VtValue>& values, uint8_t arn
                 }
             }
         }
-        AiArrayUnmap(array);
+        if (array)
+            AiArrayUnmap(array);
```
sebastienblor, line 787: Match the scalar branch above: AiArrayAllocate may return nullptr
(e.g. when every v.size() is zero on the first iteration and the
allocator declines), in which case calling AiArrayUnmap on a null
array would crash. The scalar branch already guards with `if (array)`.

libs/common/procedural_reader.cpp
```diff
@@ -37,6 +37,7 @@ void ProceduralReader::Read(const std::string &filename,
         // Only open the usd file as a root layer
         if (rootLayer == nullptr) {
             AiMsgError("[usd] Failed to open file (%s)", filename.c_str());
+            _filename = "";
```
sebastienblor, line 40: The trailing cleanup below clears _filename/_overrides on the
success path; the early return here used to skip it, leaving
 _filename pointing at the failed file for any later inspection.

libs/common/rendersettings_utils.cpp
```diff
@@ -89,7 +89,8 @@ ArnoldAOVTypes GetArnoldTypesFromFormatToken(const TfToken &type)
         return {"FLOAT", str::aov_write_float, str::user_data_float, true};
     } else if (type == _tokens->_float) {
         return {"FLOAT", str::aov_write_float, str::user_data_float, false};
-    } else if (type == _tokens->_int || type == _tokens->i8 || type == _tokens->uint8) {
+    } else if (type == _tokens->_int || type == _tokens->i8 || type == _tokens->int8 ||
+               type == _tokens->ui8 || type == _tokens->uint8) {
```
sebastienblor, line 93: Both spellings of each token are declared above ((i8) (int8) and
(ui8) (uint8)) for client compatibility. The previous version only
matched two of them, so AOVs declared with format "int8" or "ui8"
silently fell through to the default RGB branch.

libs/common/rendersettings_utils.cpp
```diff
@@ -309,7 +310,7 @@ AtNode * DeduceDriverFromFilename(const UsdRenderProduct &renderProduct, ArnoldA
     }
 
     // Get the proper driver type based on the file extension
-    if (extension == "tif")
+    if (extension == "tif" || extension == "tiff")
```
sebastienblor, line 313: ".tiff" was previously not recognized and silently fell through to
the default driver_exr, producing a file named "foo.tiff" with EXR
data inside. Treat the two TIFF extensions identically.

libs/common/shape_utils.cpp
```diff
@@ -91,6 +91,9 @@ void ArnoldUsdReadCreases(
         ll += 1;
     }
 
+    AiArrayUnmap(creaseIdxsArray);
+    AiArrayUnmap(creaseSharpnessArray);
+
```
sebastienblor, line 96: The two crease arrays were obtained via AiArrayMap above; every other
mapped array in this file is paired with AiArrayUnmap, but these two were
missing it. Without the unmap, the array's mapped state is leaked and
subsequent reads/writes from Arnold may observe stale data.

libs/common/shape_utils.cpp
```diff
@@ -301,6 +304,8 @@ inline bool _FlattenIndexedValue(const VtValue& in, const VtIntArray& idx, VtVal
         return false;
 
     const VtArray<T>& inArray = in.UncheckedGet<VtArray<T>>();
+    if (inArray.empty())
+        return false;
```
sebastienblor, line 308: If the source array is empty, the clamp range below collapses to
[0, -1] (max < min), AiClamp's behaviour becomes undefined, and any
subsequent inArray[...] indexes an empty container — UB. Bail out so the
caller can treat the primvar as missing instead.

cpichard:
:x: CI failed (build #240).

cpichard:
:x: CI failed (build 245).

cpichard:
Build 245 failed: `test_1906` (Light blocker) is failing on all platforms (macOS, Linux, Windows).

sebastienblor:
Should be fixed now

cpichard:
:white_check_mark: CI passed (build 253).

cpichard (approved):

PR #2668, opened 2026-05-26
Cleanup & fixes in the writer

sebastienblor:


libs/translator/writer/prim_writer.cpp
```diff
@@ -1203,8 +1212,11 @@ void UsdArnoldPrimWriter::_WriteMatrix(UsdGeomXformable& xformable, const AtNode
         }
     }
     // Identity matrix, nothing to write
-    if (!hasMatrix)
+    if (!hasMatrix) {
+        if (matrices)
+            AiArrayUnmapConst(array);
```
sebastienblor, line 1217: The matrix array was mapped above with AiArrayMapConst; the
matching AiArrayUnmapConst at the bottom of the function would
be skipped on this early return, leaking the map until the next
call (or forever). Unmap before returning.

libs/translator/writer/prim_writer.cpp
```diff
@@ -1389,11 +1401,13 @@ void UsdArnoldPrimWriter::_WriteMaterialBinding(
...
-                // If we are in append mode, we want to check if the subsets already exist, in which case 
+                // If we are in append mode, we want to check if the subsets already exist, in which case
                 // we don't want to create them again
                 std::vector<UsdGeomSubset> prevSubsets = UsdGeomSubset::GetGeomSubsets(geom, _tokens->face);
-                if (prevSubsets.size() >= numSubsets)
-                   return; 
+                if (prevSubsets.size() >= numSubsets) {
+                    AiArrayUnmap(shidxsArray);
+                    return;
+                }
```
sebastienblor, line 1410: shidxsArray was mapped a few lines above; the early
return must unmap it, otherwise the array stays locked
for the rest of the export and leaks. The matching
unmap below (after the subsets loop) would be skipped.

libs/translator/writer/write_arnold_type.cpp
```diff
@@ -172,7 +180,7 @@ void UsdArnoldWriteGinstance::Write(const AtNode *node, UsdArnoldWriter &writer)
         UsdAttribute extentsAttr = targetBoundable.GetExtentAttr();
         if (extentsAttr) {
             VtVec3fArray extents;
-            extentsAttr.Get(&extents, (float)writer.GetTime().GetValue());
+            extentsAttr.Get(&extents, writer.GetTime());
```
sebastienblor, line 183: Pass the UsdTimeCode through directly. Going via
(float)writer.GetTime().GetValue() reconstructs a non-default
TimeCode even when writer.GetTime() *is* default — and
UsdTimeCode::Default().GetValue() is NaN, so the resulting
sample is queried at NaN time.

libs/translator/writer/write_camera.cpp
```diff
@@ -106,7 +106,7 @@ void UsdArnoldWriteCamera::Write(const AtNode *node, UsdArnoldWriter &writer)
     _WriteMatrix(cam, node, writer);
 
     AtArray* matrices = AiNodeGetArray(node, AtString("matrix"));
-    unsigned int numKeys = matrices ? AiArrayGetNumKeys(matrices) : 1;
+    unsigned int numKeys = matrices ? AiArrayGetNumKeys(matrices) : 0;
```
sebastienblor, line 109: The ternary above guards AiArrayGetNumKeys against a null array but the
loop below would still dereference it via AiArrayGetMtx, which crashes.
If there is no matrix array there is nothing to inspect — treat that the
same as "no non-identity matrix".

libs/translator/writer/write_geometry.cpp
```diff
@@ -91,7 +91,7 @@ void UsdArnoldWriteMesh::Write(const AtNode *node, UsdArnoldWriter &writer)
         if (nelems > 0) {
             vtArrNsides.resize(nelems);
             for (unsigned int i = 0; i < nelems; ++i)
-                vtArrNsides[i] = (unsigned char)AiArrayGetUInt(nsides, i);
+                vtArrNsides[i] = (int)AiArrayGetUInt(nsides, i);
```
sebastienblor, line 94: Casting through (unsigned char) here truncates polygons with
more than 255 vertices (e.g. a 300-sided face would be wrapped
to 44), silently corrupting the topology. vtArrNsides is a
VtArray<int>, so we should convert straight to int.

libs/translator/writer/write_options.cpp
```diff
@@ -238,8 +238,8 @@ void UsdArnoldWriteOptions::Write(const AtNode *node, UsdArnoldWriter &writer)
                        AiNodeGetInt(node, str::region_max_x),
                        AiNodeGetInt(node, str::region_max_y));
 
-    if (cropRegion[0] > 0 && cropRegion[1] > 0 && 
-        cropRegion[2] > 0 && cropRegion[3] > 0) {
+    if (cropRegion[0] >= 0 && cropRegion[1] >= 0 &&
+        cropRegion[2] >= 0 && cropRegion[3] >= 0) {
```
sebastienblor, line 242: Arnold's region_min_* defaults to INT_MIN, so we want to detect that the
user explicitly set them by checking >= 0 (0 is a legitimate value: it
means crop starts at the top/left edge). Using > 0 silently drops crop
regions whose min lies on the image border, e.g. (0, 0, w/2-1, h/2-1).

cpichard:
:white_check_mark: CI passed (build 246).

cpichard (approved):
All fixes are correct and well-scoped. Good to merge.

PR #2670, opened 2026-05-26
Fix compilation with 25.05

cpichard:
- the holeIndices patch does not compile with USD 25.05 because of insert not being exposed. We use push_back instead.


cpichard:
:white_check_mark: CI passed (build 256).

PR #2674, opened 2026-05-29
Fix spurious warnings "Arnold attribute not recognized"

### LINKED ISSUE #2673: Remove spurious warnings "Arnold attribute not recognized"
When rendering usd files coming from houdini, we can have output attributes like "r", "g", b", "a", "rgba", etc...
This can cause warnings that should be removed

sebastienblor:
When we find an unrecognized attribute, we first check if it's in the allowed-list before dumping a warning

Fixes #2673


cpichard:
:white_check_mark: CI passed (build 263).

cpichard (dismissed):

PR #2675, opened 2026-05-30
Support uv / st privmar AOVs

### LINKED ISSUE #1286: UV RenderVar does not render with kick
**Describe the bug**
If you set up a render var that uses the st primvar, it renders in the render delegate, but not when you render with kick (with kick, you get all 0 0 0)

For example
```

            def RenderVar "rendervar1"
            {
                uniform token dataType = "color3f"
                custom string driver:parameters:aov:channel_prefix = "UV"
                custom int driver:parameters:aov:clearValue = 0
                custom token driver:parameters:aov:format = "color3f"
                custom bool driver:parameters:aov:multiSampled = 0
                custom string driver:parameters:aov:name = "UV"
                uniform string sourceName = "st"
                uniform token sourceType = "primvar"
            }
```

**To Reproduce**
- Extract the attached zip file
- Kick the usda file
- st AOV is all 0 0 0

If you convert the usda file to ass, there is a shader setup for the "st" AOV (it uses a user data shader to read the "st" attribute). But there is no "st" user data on the polymesh.

**Used Software Versions**
 - Arnold: [7.1.3]

**Additional context**
[testing.zip](https://github.com/Autodesk/arnold-usd/files/9527939/testing.zip)

sebastienblor:
Adding support for these UV primvars that need to be treated in a specific way (i.e. without a user_data shader). Added test_1286

Fixes #1286


cpichard:
:white_check_mark: CI passed (build 262).

cpichard (approved):

PR #2681, opened 2026-06-04
Don't reset lights at every IPR iteration

### LINKED ISSUE #255: Remove AiNodeReset from the light sync function
Remove AiNodeReset from the light sync function and only reset the parameters that need to be set to their default value.

This needs some investigation to figure out if we are receiving all parameters whenever light params are dirtied.

sebastienblor:
In this PR we no longer reset the whole light parameters at every IPR iteration.
We just translate the parameters that have been changed, as for other nodes. We no longer need a special treatment for the light's transforms

Fixes #255


cpichard:
:white_check_mark: CI passed (build 269).

cpichard (approved):

PR #2683, opened 2026-06-07
Fix tbb exceptions with nested native instancing

sebastienblor:
When the amount of nested instances is too big with shape instancing, since we flatten all the instances into a single list, we can end up having crashes. This PR catches this exception and rolls back to the previous behaviour

### DIFF
--- a/libs/render_delegate/shape.cpp
+++ b/libs/render_delegate/shape.cpp
@@ -18,6 +18,8 @@
 #include "shape.h"
 #include <pxr/base/trace/trace.h>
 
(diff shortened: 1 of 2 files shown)


cpichard:
can we add an entry in the changelog ?

sebastienblor:
> can we add an entry in the changelog ?

it's a fix for a non-released feature, so we don't need it

cpichard (approved):

cpichard:
:white_check_mark: CI passed (build 270).

PR #2684, opened 2026-06-09
Fix #2641: animate instanced lights driven by a Point Instancer in Hydra

### LINKED ISSUE #2641: Cached Point Instancer lights and global light sampling controls appear broken in Arnold USD/Solaris
**Describe the bug**

Arnold USD / Solaris appears to have issues with animated instanced lights when using cached Point Instancer workflows, and global light sampling does not appear to behave as expected with many instanced lights.

In my tests, lights instanced through a cached animated Point Instancer did not update/animate correctly after caching. The lights appeared stuck or failed to reflect the expected animated Point Instancer motion. To get the lights animated after caching, I had to switch to a Reference Instancer-style workflow instead of relying on the cached Point Instancer workflow.

There also appears to be an issue with global light sampling when using many instanced lights. Instanced lights behave differently from explicitly authored/non-instanced lights, and the result appears unreliable when global light sampling is enabled or expected to affect the render.

I also could not find an equivalent global light sampling dropdown/control in the HtoA/Solaris ROP similar to the Arnold for Maya workflow, or the available HtoA/Solaris controls do not appear to drive the same behavior correctly.

Setting the sampling mode to `auto` on individual lights did not appear to produce the expected behavior either. In a scene with roughly 200 lights, changing those light sampling settings produced no noticeable render-time change. Allowing/enabling light sampling on the Arnold ROP also did not appear to produce the expected render-time or sampling behavior change, even though a render-time difference would be expected according to Arnold’s documentation around light sampling / global light sampling.

This makes it difficult to use animated instanced lights in Solaris with Arnold, especially for workflows where many secondary/accent lights need to be cached, animated, and sampled efficiently.

**To Reproduce**

1. Open the attached Houdini scene file.
2. In Solaris, inspect the animated instanced light setup driven by cached point data.
3. Render or preview the cached Point Instancer light setup with Arnold USD / Arnold Hydra delegate.
4. Observe that the instanced lights do not update/animate correctly after caching, or appear stuck.
5. Compare this against the Reference Instancer-style setup in the file.
6. Observe that the lights can animate correctly with the Reference Instancer workflow, while the cached Point Instancer workflow remains problematic.
7. Use the many-light setup in the file, roughly 200 lights.
8. Try changing the sampling mode to `auto` on the individual lights.
9. Enable/allow the available light sampling control on the Arnold ROP.
10. Render or preview again with Arnold USD / Solaris.
11. Observe that changing the per-light sampling mode and enabling/allowing light sampling on the Arnold ROP does not appear to produce the expected render-time or sampling behavior changes.
12. Try to locate an equivalent global light sampling control in the HtoA/Solaris ROP.
13. Observe that the HtoA/Solaris ROP does not appear to expose the same kind of direct light sampling dropdown/control available in Arnold for Maya workflows, or the available controls do not appear to drive the expected Arnold light-sampling behavior.

**Expected behavior**

Animated lights instanced through a cached Point Instancer should update correctly over time in Arnold USD/Solaris.

Instanced lights should also work reliably with Arnold’s global light sampling, or the limitations should be clearly documented if this workflow is unsupported.

For scenes with many lights, changing light sampling / global light sampling settings should produce the expected sampling and render-time behavior. This includes both per-light sampling settings and any available light sampling controls on the Arnold ROP. If the HtoA/Solaris ROP uses a different control path than Arnold for Maya, it would be useful for that to be exposed clearly or documented.

If Point Instancer-driven lights are unsupported, partially supported, or require specific USD authoring patterns, it would be useful to know the supported workflow for animated/cached instanced lights in Solaris.

**Reproduction file**

I am attaching a Houdini scene file demonstrating the issue.

The file contains the relevant setup for:

- animated instanced lights
- cached Point Instancer workflow
- Reference Instancer workaround
- many-light setup used for light sampling tests
- per-light sampling settings
- Arnold ROP / HtoA Solaris ROP light sampling settings used during testing

The issue can be checked by comparing the cached Point Instancer light setup against the Reference Instancer setup, and by observing the render-time / sampling behavior when changing the available light sampling controls on individual lights and on the Arnold ROP.

**Used Software Versions**

- Arnold / HtoA: `6.5.1.1`
- Arnold USD / Hydra Delegate: `included with HtoA 6.5.1.1`
- Houdini: `21.0.671`
- USD: `Houdini-bundled USD version`
- Compiler: `Not applicable — using prebuilt Houdini/HtoA packages`
- OS: `macOS 26.4.1`
- Hardware: `Apple Silicon / M2`
- Any 3rd-party app: `SideFX Houdini Solaris`

**Additional context**

The goal was to use instanced real light prims in Solaris for many secondary/accent lights, with animation driven by cached point data.

The cached Point Instancer workflow did not behave correctly in Arnold USD/Solaris: after caching, the instanced lights appeared stuck or did not update as expected. A Reference Instancer-style workflow was needed to make the lights animate after caching.

The second issue is related to sampling many lights efficiently. In a scene with roughly 200 lights, setting individual lights to `auto` sampling did not appear to change render time or sampling behavior. Enabling/allowing the available light sampling control on the Arnold ROP also did not appear to produce the expected render-time or sampling behavior change.

I also could not find an equivalent global light sampling dropdown/control in the HtoA/Solaris ROP similar to the Arnold for Maya workflow, or the available HtoA/Solaris controls do not appear to drive the same behavior correctly.

This appears separate from material/color-management issues and separate from geometry normal/subdivision issues. This report is specifically about animated instanced lights, cached Point Instancer behavior, Reference Instancer behavior, global light sampling with instanced lights, and missing or incorrectly behaving HtoA/Solaris ROP controls for light sampling.

Possible areas to investigate:

- Arnold USD support for Point Instancer-driven light prims
- animated cached Point Instancer evaluation
- time-sampled Point Instancer attributes
- light prim instancing support in Arnold USD/Solaris
- Reference Instancer vs Point Instancer behavior
- global light sampling with instanced lights
- missing or incorrectly behaving HtoA/Solaris ROP controls for global light sampling
- Arnold ROP light sampling control not affecting render behavior as expected
- per-light sampling mode not affecting render behavior as expected
- difference between explicitly authored lights and instanced lights
- cached USD/USDC animated instancer evaluation

I am attaching a Houdini scene file with the relevant setup so the behavior can be reproduced directly.

File link: https://www.swisstransfer.com/d/13cfb84b-c282-4d86-99f9-dab3a6715e40

cpichard:
Fixes #2641.

### Problem
Lights instanced through an animated / cached **Point Instancer** stay frozen at their first-frame positions in the Hydra render delegate (Solaris), while a Reference Instancer workflow animates correctly.

### Root cause
An instanced light is a **Sprim**, and Hydra does not propagate a Point Instancer's dirtiness to it the way it does for instanced **Rprims**. When an instancer's prototypes are *only* lights, nothing carries the dirty notification to the light Sprim, so `HdArnoldGenericLight::Sync` is never called again after the first frame and the Arnold instancer is never rebuilt with the new transforms.

### Fix
- `HdArnoldGenericLight::Sync` registers a dependency on its instancer.
- `HdArnoldRenderDelegate::HasPendingChanges` detects instancers that prims depend on and that are still dirty in the change tracker (because no Rprim caused Hydra to sync and clean them) and propagates the dirtiness to the dependent prims — mirroring how material dependencies already work. The light's re-sync calls the instancer's `Sync`, which clears the instancer's dirty bits and prevents this from looping every frame.
- Make the dependency `markPrimDirty` path **scene-index-emulation (Hydra 2) aware**: route dirtying through the terminal scene index instead of calling `HdChangeTracker::Mark{R,S}primDirty` directly, which raises a `requires emulation` coding error and drops the dirtiness. This also fixes the same latent issue for material/camera dependency propagation.

### Test
Adds `testsuite/test_2641` — animated Point-Instancer-driven `SphereLight`s over a ground plane, rendered at frame 16 to guard the instanced-light code path.

> [!NOTE]
> This is an interactive / re-render bug: a single batch render rebuilds the instancer on the first Sync, so the batch test cannot reproduce the freeze itself. The animation was verified by scrubbing the timeline in Houdini/Solaris.

### Scope
This addresses item (1) of the issue — animated instanced lights not updating. The other two points raised in the issue are out of scope here:
- Global light-sampling efficiency with many instanced lights → Arnold-core sampling behavior.
- Missing equivalent global-light-sampling control in the HtoA/Solaris ROP → HtoA UI, a different repository.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cpichard:
:white_check_mark: CI passed (build 271).

sebastienblor (dismissed):

PR #2685, opened 2026-06-09
Fix name collision with arnold instancer nodes

sebastienblor:
When creating arnold instancers with nested instances, there can be name collisions as the instancer names do not include the instancer id.
This causes messages like 

` WARNING | A node named "/Assets/asset565/pi{instance=566}_instancer" already exists: this newly added instancer node will be renamed to an empty string`


cpichard:
:white_check_mark: CI passed (build 277).

libs/render_delegate/instancer.cpp
```diff
@@ -563,9 +563,23 @@ void HdArnoldInstancer::CreateArnoldInstancer(HdArnoldRenderDelegate* renderDele
...
+    // The arnold instancer node name must be globally unique. A separate chain of arnold
+    // instancer nodes is built per prototype shape (see HdArnoldShape), so when several shapes
+    // share the same (parentInstancer, childInstancer) pair, naming a node solely from those
+    // ids collides. We therefore derive the name from the unique child node we just created.
+    //
+    // Invariant: both callers (HdArnoldShape / HdArnoldLight) clear the instancers vector before
+    // the first call, and the parent recursion below only runs after a node has been pushed, so
+    // an empty vector here means this is the innermost (leaf) call. For the leaf we combine the
+    // instancer id with the prototype (shape) id; for parents we prefix the child node's name,
+    // which is already unique by induction. See NESTED_INSTANCER_CRASHES.md.
```
cpichard, line 576: The comment ends with "See NESTED_INSTANCER_CRASHES.md"  but the file doesn't exist in the repo

cpichard:
Name length grows ~quadratically with nesting depth (minor). Each parent embeds the entire child node name, so an N-deep chain produces names of O(N × path-length).

cpichard (dismissed):

PR #2686, opened 2026-06-10
MTOA-2673: prevent subdivisionScheme from being written as timeSamples in USD export

compso:
subdivisionScheme is a uniform attribute in USD (SdfVariabilityUniform) and must never have time samples. When MayaUSD exports with the Arnold job context over a frame range, the Arnold scene is rebuilt per frame. If the Arnold mesh subdiv_type evaluates differently between frames (e.g. "catclark" on the first frame due to displaySmoothMesh, "none" on subsequent frames), the previous SetAttribute logic would detect the value change and create timeSamples, producing invalid USD output such as:

  uniform token subdivisionScheme.timeSamples = {
      1058: "catmullClark",
      1059: "none",
      ...
  }

This caused render errors: "wrong number of motion keys in vertex/normal arrays".

Write subdivisionScheme directly at default time via attr.Set() (bypassing the time-varying writer logic). When subdiv_type is "none" (Arnold default, no explicit subdivision requested), skip writing if a value was already authored -- this preserves a "catmullClark" value that MayaUSD's native mesh exporter may have written.

Fixes MTOA-2673

**Additional context**


sebastienblor:
ci passed

sebastienblor (approved):

PR #2690, opened 2026-06-11
Various fixes in render_param

sebastienblor:
Various fixes in libs/render_delegate/render_param.cpp.
In particular it allows to have several render delegates in parallel

### DIFF
--- a/libs/render_delegate/render_param.cpp
+++ b/libs/render_delegate/render_param.cpp
@@ -43,25 +43,29 @@ TF_DEBUG_CODES(HDARNOLD_RENDER_SETTINGS);
 
 namespace {
 
-std::string cachedLogMsg = "";
-std::mutex cachedLogMutex;
-
 void _MsgStatusCallback(int logmask, int severity, const char* msgString, AtParamValueMap* metadata, void* userPtr)
 {
-    std::lock_guard<std::mutex> guard(cachedLogMutex);
-
-    cachedLogMsg = std::string(msgString);
+    if (msgString == nullptr || userPtr == nullptr) {
+        return;
+    }
+    static_cast<HdArnoldRenderParam*>(userPtr)->SetCachedLogMessage(msgString);
 }
 
 } // namespace
 
 TF_DEFINE_ENV_SETTING(HDARNOLD_DEBUG_SCENE, "", "Optionally save out the arnold scene before rendering.");
 
+HdArnoldRenderParam::~HdArnoldRenderParam()
+{
+    StopRenderMsgLog();
+}
+
 HdArnoldRenderParam::HdArnoldRenderParam(HdArnoldRenderDelegate* delegate) : _delegate(delegate)
 
 {
     _needsRestart.store(false, std::memory_order::memory_order_release);
     _aborted.store(false, std::memory_order::memory_order_release);
+    _paused.store(false, std::memory_order::memory_order_release);
 
     ResetStartTimer();
 
@@ -72,6 +76,13 @@ HdArnoldRenderParam::HdArnoldRenderParam(HdArnoldRenderDelegate* delegate) : _de
 
 HdArnoldRenderParam::Status HdArnoldRenderParam::UpdateRender()
 {
+    // Mirror the null check Interrupt() now performs. Without it any caller
+    // that ends up with a null _delegate crashes on the first GetRenderSession
+    // dereference below; Aborted is a safer, observable signal that something
+    // is wrong with the setup.
+    if (_delegate == nullptr) {
+        return Status::Aborted;
+    }
     const auto aborted = _aborted.load(std::memory_order_acquire);
     // Checking early if the render was aborted earlier.
     if (aborted) {
@@ -84,6 +95,12 @@ HdArnoldRenderParam::Status HdArnoldRenderParam::UpdateRender()
 
         case AI_RENDER_STATUS_RESTARTING:
         case AI_RENDER_STATUS_RENDERING:
+            if (needsRestart) {
+                _needsRestart.store(true, std::memory_order_release);
+            }
+            if (paused) {
+                _paused.store(true, std::memory_order_release);
+            }
             return Status::Converging;
         
         case AI_RENDER_STATUS_FINISHED:
@@ -110,6 +127,8 @@ HdArnoldRenderParam::Status HdArnoldRenderParam::UpdateRender()
                 if (!_debugScene.empty())
                     WriteDebugScene();
                 AiRenderRestart(_delegate->GetRenderSession());
+                RestartRenderMsgLog();
+                ResetStartTimer();
             } else if (!paused) {
                 AiRenderResume(_delegate->GetRenderSession());
                 ResetStartTimer();
@@ -117,8 +136,12 @@ HdArnoldRenderParam::Status HdArnoldRenderParam::UpdateRender()
             return Status::Converging;
 
         case AI_RENDER_STATUS_FAILED:
-            _aborted.store(true, std::memory_order_release);
+            // Write _errorCode BEFORE publishing _aborted with a release store.
+            // The release semantics on _aborted ensure that any thread which
+            // observes _aborted == true via acquire load also sees the writes
+            // performed before the release — including the new _errorCode.
             _errorCode = AiRenderEnd(_delegate->GetRenderSession());
+            _aborted.store(true, std::memory_order_release);
             if (_errorCode == AI_ABORT) {
                 TF_WARN("[arnold-usd] Render was aborted.");
             } else if (_errorCode == AI_ERROR_NO_CAMERA) {
@@ -140,9 +163,19 @@ HdArnoldRenderParam::Status HdArnoldRenderParam::UpdateRender()
             } else if (_errorCode == AI_ERROR) {
                 TF_WARN("[arnold-usd] Generic error.");
             }
+            StopRenderMsgLog();
             return Status::Aborted;
         
         case AI_RENDER_STATUS_NOT_STARTED:
+            // If the caller already requested a pause before the render even
+            // begins (e.g., a viewport that opens in a paused state), honour
+            // it: do not call AiRenderBegin, and put `_paused` back so the
+            // next UpdateRender iterations continue to see the pause request
+            // until Resume() / Restart() clears it.
+            if (paused) {
+                _paused.store(true, std::memory_order_release);
+                return Status::Converging;
+            }
             if (!_debugScene.empty())
                 WriteDebugScene();
             AiRenderBegin(_delegate->GetRenderSession());
@@ -158,9 +191,11 @@ HdArnoldRenderParam::Status HdArnoldRenderParam::UpdateRender()
 
 void HdArnoldRenderParam::Interrupt(bool needsRestart, bool clearStatus)
 {
-    if (_delegate && _delegate->IsBatchContext()) return;
+    if (_delegate == nullptr || _delegate->IsBatchContext()) return;
     const auto status = AiRenderGetStatus(_delegate->GetRenderSession());
-    if (status != AI_RENDER_STATUS_NOT_STARTED) {
+    if (status == AI_RENDER_STATUS_RENDERING ||
+        status == AI_RENDER_STATUS_RESTARTING ||
+        status == AI_RENDER_STATUS_PAUSED) {
         AiRenderInterrupt(_delegate->GetRenderSession(), AI_BLOCKING);
     }
     if (needsRestart) {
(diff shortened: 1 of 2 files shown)


libs/render_delegate/render_param.cpp
```diff
@@ -43,25 +43,29 @@ TF_DEBUG_CODES(HDARNOLD_RENDER_SETTINGS);
...
 
-std::string cachedLogMsg = "";
-std::mutex cachedLogMutex;
-
 void _MsgStatusCallback(int logmask, int severity, const char* msgString, AtParamValueMap* metadata, void* userPtr)
 {
-    std::lock_guard<std::mutex> guard(cachedLogMutex);
-
-    cachedLogMsg = std::string(msgString);
+    if (msgString == nullptr || userPtr == nullptr) {
```
sebastienblor, line 48: Guard against a null msgString. Constructing std::string(const char*) from
a null pointer is undefined behavior — this guard protects against any
Arnold release that might deliver a null status string for a transient
state such as initial registration.

libs/render_delegate/render_param.cpp
```diff
@@ -43,25 +43,29 @@ TF_DEBUG_CODES(HDARNOLD_RENDER_SETTINGS);
...
+        return;
+    }
+    static_cast<HdArnoldRenderParam*>(userPtr)->SetCachedLogMessage(msgString);
```
sebastienblor, line 51: Route the callback to the originating HdArnoldRenderParam via its userPtr.
Previously this wrote to a file-scope global, which meant multiple render
delegates shared a single cachedLogMsg and would clobber each other.

libs/render_delegate/render_param.cpp
```diff
@@ -43,25 +43,29 @@ TF_DEBUG_CODES(HDARNOLD_RENDER_SETTINGS);
...
 }
 
 } // namespace
 
 TF_DEFINE_ENV_SETTING(HDARNOLD_DEBUG_SCENE, "", "Optionally save out the arnold scene before rendering.");
 
+HdArnoldRenderParam::~HdArnoldRenderParam()
+{
+    StopRenderMsgLog();
```
sebastienblor, line 60: The defaulted destructor used to leave any registered message-log
callback live in Arnold's global table after the instance was destroyed.
Explicitly tear it down here.

libs/render_delegate/render_param.cpp
```diff
@@ -43,25 +43,29 @@ TF_DEBUG_CODES(HDARNOLD_RENDER_SETTINGS);
...
+}
+
 HdArnoldRenderParam::HdArnoldRenderParam(HdArnoldRenderDelegate* delegate) : _delegate(delegate)
 
 {
     _needsRestart.store(false, std::memory_order::memory_order_release);
     _aborted.store(false, std::memory_order::memory_order_release);
+    _paused.store(false, std::memory_order::memory_order_release);
```
sebastienblor, line 68: Previously _paused was never explicitly initialized here, so on
pre-C++20 toolchains its initial value was indeterminate.

libs/render_delegate/render_param.cpp
```diff
@@ -84,6 +95,12 @@ HdArnoldRenderParam::Status HdArnoldRenderParam::UpdateRender()
 
         case AI_RENDER_STATUS_RESTARTING:
         case AI_RENDER_STATUS_RENDERING:
+            if (needsRestart) {
```
sebastienblor, line 98: We consumed needsRestart / paused with exchange(false) above, but
this branch cannot act on either (Arnold is mid-render and will
not transition to FINISHED/PAUSED until later). Put them back so
the next iteration — which will see the post-render state — can
honour them. Without this, calling Restart() while a render was
in progress was a silent no-op: the flag was consumed and lost,
and the render eventually converged with stale scene data.

libs/render_delegate/render_param.cpp
```diff
@@ -110,15 +127,21 @@ HdArnoldRenderParam::Status HdArnoldRenderParam::UpdateRender()
                 if (!_debugScene.empty())
                     WriteDebugScene();
                 AiRenderRestart(_delegate->GetRenderSession());
+                RestartRenderMsgLog();
+                ResetStartTimer();
```
sebastienblor, line 131: Mirror the FINISHED branch: a needsRestart from PAUSED is a fresh
start of rendering, so the elapsed-time clock and the message-log
callback must be reset/restarted. Previously these calls were only
present in the FINISHED branch, so restarting a paused render
reported wrong elapsed time and surfaced stale status messages.

libs/render_delegate/render_param.cpp
```diff
@@ -110,15 +127,21 @@ HdArnoldRenderParam::Status HdArnoldRenderParam::UpdateRender()
...
                 ResetStartTimer();
             }
             return Status::Converging;
 
         case AI_RENDER_STATUS_FAILED:
-            _aborted.store(true, std::memory_order_release);
+            // Write _errorCode BEFORE publishing _aborted with a release store.
+            // The release semantics on _aborted ensure that any thread which
+            // observes _aborted == true via acquire load also sees the writes
+            // performed before the release — including the new _errorCode.
```
sebastienblor, line 142: The previous order published _aborted=true *before* assigning
_errorCode, so a concurrent GetErrorCode() reader could observe
_aborted=true paired with a stale (previous-render) _errorCode.

libs/render_delegate/render_param.cpp
```diff
@@ -140,9 +163,19 @@ HdArnoldRenderParam::Status HdArnoldRenderParam::UpdateRender()
             } else if (_errorCode == AI_ERROR) {
                 TF_WARN("[arnold-usd] Generic error.");
             }
+            StopRenderMsgLog();
```
sebastienblor, line 166: Mirror the FINISHED branch: once the render is ended we no longer
want the status-message callback to keep firing and overwriting
cachedLogMsg with unrelated log lines. The previous code only
stopped the callback on FINISHED, so a failed render left the
callback live until the next StartRenderMsgLog or destruction.

libs/render_delegate/render_param.cpp
```diff
@@ -158,9 +191,11 @@ HdArnoldRenderParam::Status HdArnoldRenderParam::UpdateRender()
 
 void HdArnoldRenderParam::Interrupt(bool needsRestart, bool clearStatus)
 {
-    if (_delegate && _delegate->IsBatchContext()) return;
+    if (_delegate == nullptr || _delegate->IsBatchContext()) return;
```
sebastienblor, line 194: The previous check `if (_delegate && _delegate->IsBatchContext()) return;`
only early-returned when _delegate was non-null AND in batch context. If
_delegate was null the short-circuit skipped the return entirely and the
next line dereferenced the null pointer through GetRenderSession().

libs/render_delegate/render_param.cpp
```diff
@@ -158,9 +191,11 @@ HdArnoldRenderParam::Status HdArnoldRenderParam::UpdateRender()
...
     const auto status = AiRenderGetStatus(_delegate->GetRenderSession());
-    if (status != AI_RENDER_STATUS_NOT_STARTED) {
+    if (status == AI_RENDER_STATUS_RENDERING ||
+        status == AI_RENDER_STATUS_RESTARTING ||
+        status == AI_RENDER_STATUS_PAUSED) {
```
sebastienblor, line 198: Only interrupt when there is actually something to interrupt. The
previous condition `status != NOT_STARTED` also invoked AiRenderInterrupt
when the render had already FINISHED or FAILED, which is at best
wasteful (no work to abort) and on FAILED is undefined: UpdateRender's
FAILED branch is the one that owns the AiRenderEnd transition, and
interrupting after that has already been queued can race with it.

libs/render_delegate/render_param.cpp
```diff
@@ -173,14 +208,21 @@ void HdArnoldRenderParam::Interrupt(bool needsRestart, bool clearStatus)
...
+    // pause that was just requested.
     _paused.store(true, std::memory_order_release);
+    Interrupt(false, false);
 }
 
 void HdArnoldRenderParam::Resume() { _paused.store(false, std::memory_order_release); }
 
 void HdArnoldRenderParam::Restart()
 {
+    _aborted.store(false, std::memory_order_release);
```
sebastienblor, line 225: Restart needs to also clear the aborted flag. UpdateRender's first action
is to read _aborted and short-circuit with Status::Aborted *before* it
consumes _needsRestart, so without clearing _aborted here a Restart()
call against a previously-aborted render was a silent no-op: the new
_needsRestart flag was set but UpdateRender bailed out before noticing it
and the render stayed aborted indefinitely.

libs/render_delegate/render_param.cpp
```diff
@@ -210,16 +252,27 @@ void HdArnoldRenderParam::WriteDebugScene() const
 
     AiMsgWarning("Saving debug arnold scene as \"%s\"", _debugScene.c_str());
     AtParamValueMap* params = AiParamValueMap();
-    AiParamValueMapSetBool(params, str::binary, false);
-    AiSceneWrite(_delegate->GetUniverse(), AtString(_debugScene.c_str()), params);
-    AiParamValueMapDestroy(params);
+    const AtString filename(_debugScene.c_str());
```
sebastienblor, line 255: Guard against AiParamValueMap returning null on allocation failure. The
previous code unconditionally fed `params` into AiParamValueMapSetBool,
AiSceneWrite, and AiParamValueMapDestroy, every one of which dereferences
its first argument.

libs/render_delegate/render_param.cpp
```diff
@@ -210,16 +252,27 @@ void HdArnoldRenderParam::WriteDebugScene() const
...
+    bool ok = false;
```
sebastienblor, line 256: Check AiSceneWrite's return value so users see a clear error if the
write fails (bad path, missing directory, permission denied, disk full).
The previous code discarded the return value, leaving the user with only
the leading "Saving debug arnold scene…" message implying success.

libs/render_delegate/render_param.cpp
```diff
@@ -210,16 +252,27 @@ void HdArnoldRenderParam::WriteDebugScene() const
...
 double HdArnoldRenderParam::GetElapsedRenderTime() const
 {
-    _renderTimeMutex.lock();
-    const auto t0 = _renderStartTime;
-    _renderTimeMutex.unlock();
+    std::chrono::time_point<std::chrono::system_clock> t0;
+    {
+        std::lock_guard<std::mutex> guard(_renderTimeMutex);
+        t0 = _renderStartTime;
+    }
```
sebastienblor, line 275: Use std::lock_guard for the same reason as ResetStartTimer above: the
earlier manual lock()/unlock() would leak the lock and deadlock every
future caller if the time_point copy ever threw. GetRenderStatusString
was already migrated to a scope-guarded lock per its own comment — this
path was the missed twin.

libs/render_delegate/render_param.cpp
```diff
@@ -231,7 +284,16 @@ void HdArnoldRenderParam::StartRenderMsgLog()
...
+    // Deregister any previously registered callback before installing a new
+    // one. Otherwise repeated Start/Start sequences (without an intervening
+    // Stop) would overwrite `_msgLogCallback`, leaking the previous handle:
+    // the callback function would remain registered with Arnold but we'd no
+    // longer hold a handle to deregister it.
+    if (_msgLogCallback != 0) {
+        AiMsgDeregisterCallback(_msgLogCallback);
+        _msgLogCallback = 0;
+    }
+    _msgLogCallback = AiMsgRegisterCallback(_MsgStatusCallback, AI_LOG_STATUS, this);
```
sebastienblor, line 296: Pass `this` so the callback can route log lines to per-instance storage
(see SetCachedLogMessage) instead of the previously-shared file-scope
global cachedLogMsg.

libs/render_delegate/render_param.cpp
```diff
@@ -254,13 +316,22 @@ std::string HdArnoldRenderParam::GetRenderStatusString() const
     // Use unique_lock + try_to_lock so the mutex is always released — including
     // when the std::string copy below throws (e.g., bad_alloc). The previous
     // manual lock/unlock would deadlock all future callers if the copy threw.
-    std::unique_lock<std::mutex> lock(cachedLogMutex, std::try_to_lock);
+    std::unique_lock<std::mutex> lock(_cachedLogMutex, std::try_to_lock);
```
sebastienblor, line 319: Both the mutex and the cached string are now per-instance members so
concurrent renders no longer overwrite each other's status strings.

cpichard:
:white_check_mark: CI passed (build 275).

cpichard (approved):

PR #2701, opened 2026-06-23
Fix crash when changing the render purpose

### LINKED ISSUE #2628: Houdini crashes with nested USD instancing
**Describe the bug**
Houdini crashes in Solaris in several reproducible workflows involving point-instanced scene assembly and material/stage operations. 
The listed reproduction paths are only quick examples found on simple primitive-based setups. In complex production scenes, crashes occur much more frequently and are often non-deterministic, with no single reliable action sequence, which makes day-to-day work unstable and constantly crash-prone.

Path 1 - Purpose switch (Arnold Hydra):
1. Open the provided HIP scene.
2. After first-level point instancing, set an explicit purpose on the resulting prim (e.g. render).
3. Set Hydra render delegate to Arnold.
4. Toggle purpose.
5. Houdini crashes.

Path 2 - Component Output disk workflow (Arnold active):
1. Open the same scene and keep Arnold active.
2. On componentoutput, enable Load from Disk.
3. Click Save to Disk.
4. Click Reload.
5. Houdini crashes.

Path 3 - Material node deletion (no Arnold render mode required):
1. Open the scene.
2. Delete either componentmaterial or materiallibrary.
3. Houdini crashes.

**Expected behavior**
Houdini should remain stable when working with nested point instancers in Solaris/USD scenes.

**Used Software Versions**
Arnold: HTOA 6.5.1.1
USD: 25.5
OS: Windows 11 Pro
Any 3rd-party app: Houdini 21.0.671

Image
Image
Image

[nested_point_instancers.zip](https://github.com/user-attachments/files/27241904/nested_point_instancers.zip)

cpichard:
- The message callback id returned by AiMsgRegisterCallback is an unsigned int which can be zero. However we used the zero value to indicate no callback is set and never destroy the callback with id 0. So when the render is stopped/restarted after a render purpose change for example, the render param is destroyed but the pointer is still kept in Arnold memory and arnold ends up calling the callback function with invalid memory, which crashes the application/arnold.
- We use -1 to indicate that no callback is set. We use integer assuming the number of callbacks will be lower that INT_MAX which is a reasonable assumption.

Fixes #2628

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - [usd#2677](https://github.com/Autodesk/arnold-usd/issues/2677) - Fix crash reading RenderSettings with empty arnold:filter on a RenderVar
 - [usd#2641](https://github.com/Autodesk/arnold-usd/issues/2641) - Animate instanced lights driven by an animated Point Instancer in Hydra
 - [usd#2695](https://github.com/Autodesk/arnold-usd/issues/2695) - Fix MaterialX boolean shaders through USD
+- [usd#2628](https://github.com/Autodesk/arnold-usd/issues/2628) - Fix crash happening after changing the purpose in hydra
 
 ## [7.5.2.0] (Unreleased)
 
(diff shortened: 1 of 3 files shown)


cpichard:
:white_check_mark: CI passed (build 282).

sebastienblor (approved):

PR #2706, opened 2026-07-04
Fix overscan 1-pixel offset with certain resolutions

sebastienblor:
In certain resolutions, an overscan can produce a 1-pixel offset as compared to e.g. what husk renders.
This PR fixes it, by using the same rationale to apply them

Fixes HTOA-3345

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 ### Features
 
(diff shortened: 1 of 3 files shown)


cpichard:
:white_check_mark: CI passed (build 5).

cpichard (approved):

PR #2708, opened 2026-07-06
Properly read MeshLightAPI parameters

### LINKED ISSUE #2707: MeshLight attributes are not always translated properly
When a mesh primitive has a `MeshLightAPI` schema applied, we are supposed to read the light attributes, but this doesn't always happen properly in the different code paths that we support

sebastienblor:
This PR ensures that meshes with the MeshLightAPI schema are properly read as mesh lights. For the attributes set with the arnold prefix, it checks it they exist in the mesh_light node entry, and if they do it sets the attributes properly.
test_2707 exercises it

Fixes #2707


cpichard:
:white_check_mark: CI passed (build 7).

cpichard (approved):

PR #2709, opened 2026-07-07
Fix invalid null shader being authored to USD from Maya

### LINKED ISSUE #2621: string[] primvars:arnold:shader = [""] results in magenta shader even if material:binding is defined
**To Reproduce**
Context: I export separate USD layers for shader (from LookDev) and for geo cache (from Anim). And this two are composited in lighting (pretty standard imho)

I've attached 3 usd files but here's how to reproduce the behavior:
LKDV
1. Create a cube in Maya and assign e.g. an `aiStandardSurface` material
2. Export it to USD with assigned materials. I used "USD Export" with `jobContext=[Arnold]`
3. Result: LKDV.usd

ANM
1. [Optional] animate the cube
2. Assign some exotic shader that'll result in primvars:arnold:shader = [""]. It was `dx11Shader` for me.
3. Export one or more frames to USD without material; just geo. (Again, I used "USD Export" with `jobContext=[Arnold]` but with `exportMaterials=0`)
4. Result: ANM.usd

LIGHT:
Composit the two layers (ANM over LKDV).

**Expected behavior**
I'd expect that the LKDV material is visible in the composited USD scene because of the material:binding prim var (despite the fact that dx11Shader resulted in primvars:arnold:shader = [""] in ANM.usd). But the cube has magenta color instead

**Used Software Versions**
- Maya 2024.2
- mtoa-5.5.4
- mayausd-0.32.0_202505060957-eeb3aaa
-  OS: Windows 11 Pro 25H2

[arnold_usd_magenta_repro.zip](https://github.com/user-attachments/files/26933216/arnold_usd_magenta_repro.zip)

sebastienblor:
When we author the arnold shader attribute, we skip authoring it if it's empty. But when the shader is an unknown one then the shaders array has 1 element that has a null value.. In that case we also want to skip authoring it, since it overrides an eventual material binding

Fixes #2621

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@
 - [usd#2641](https://github.com/Autodesk/arnold-usd/issues/2641) - Animate instanced lights driven by an animated Point Instancer in Hydra
 - [usd#2695](https://github.com/Autodesk/arnold-usd/issues/2695) - Fix MaterialX boolean shaders through USD
 - [usd#2628](https://github.com/Autodesk/arnold-usd/issues/2628) - Fix crash happening after changing the purpose in hydra
+- [usd#2621](https://github.com/Autodesk/arnold-usd/issues/2621) - Fix invalid null shader being authored to USD from Maya
+
 
 ## [7.5.2.0] (Unreleased)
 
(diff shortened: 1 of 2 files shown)


cpichard:
:white_check_mark: CI passed (build 9).

cpichard (approved):

PR #2711, opened 2026-07-08
Fix for ignored primvars:crypto_asset

### LINKED ISSUE #2699: primvars:crypto_asset is ignored on instanceable references
**Describe the bug**
    When using an instanceable reference in Houdini Solaris/USD, the primvars:crypto_asset value is not used for Cryptomatte asset AOVs during rendering.
    Primvars can be authored on a parent prim and inherited by descendant prims until they are resolved on the actual geometry prim. For Cryptomatte workflows, this allows assigning primvars:crypto_asset higher in the hierarchy, while still expecting the resolved value to be available on the renderable Mesh prims.
    However, when the referenced prim is marked as instanceable, Arnold does not seem to pick up the inherited primvars:crypto_asset value from these primitives. As a result, the Cryptomatte asset AOV does not contain the expected asset ID information.
 
**To Reproduce**
1. In Houdini Solaris, create or reference a USD asset containing Mesh geometry.
2. Assign a primvars:crypto_asset value on a parent prim above the Mesh, relying on standard USD primvar inheritance so that the value resolves on the renderable Mesh prims.
3.Add this asset to the stage using a reference.
4. Enable instanceable on the referenced prim.
5. Render the scene with Arnold and enable the Cryptomatte asset AOV.
6. Inspect the rendered Cryptomatte asset AOV.
7. Observe that the expected primvars:crypto_asset value is missing or ignored for geometry inside the instanceable reference.
8. Disable instanceable on the same reference and render again.
9. Observe that the Cryptomatte asset AOV now contains the expected asset information.

**Expected behavior**
    Although `instanceable` references have limitations regarding per-instance edits, it should still be possible to customize the Cryptomatte asset ID for instanceable assets.
   If this is not supported due to instanceability limitations, the custom `primvars:crypto_asset` assignment should at least be ignored safely. In that case, Arnold should fall back to the default Cryptomatte asset ID generation instead of producing `out-of-range` errors or missing Cryptomatte data.

**Used Software Versions**
 - Arnold: 6.5.2.0
 - USD: 0.25.5
 - OS: indows 11 Pro
 - Any 3rd-party app: Houdini 21.0.729

**Additional context**
When the asset is used as an `instanceable` reference, Arnold reports an `out-of-range error` in the render log. This may be related to the missing Cryptomatte asset information, since the error is not observed when `instanceable` is disabled.

[crypto_asset_out_of_range.zip](https://github.com/user-attachments/files/29241940/crypto_asset_out_of_range.zip)

cpichard:
Issue: primvars:crypto_asset (and similarly resolved inherited primvars) was being ignored when an asset was added through an instanceable reference, causing the Cryptomatte asset AOV to come out empty.

Root cause: The primvar is authored as a constant value on a parent Xform and inherited down to the geometry. When resolved through an instanceable reference, it arrived at the shape as a single-element uniform (per-element) array rather than a true constant value. Arnold declared this as a size-1 uniform array, but then tried to index it per-face, producing an out-of-range error (index N of 1) and silently failing to apply the value.

Fix (libs/common/parameters_utils.cpp): DeclareAndAssignParameter now treats any single-element array as constant user data, unless it's explicitly a "constant array" scope (scope == constantArray). Previously it only did this for scope == constant or the special-cased displayColor attribute — so uniform/varying/indexed single-element arrays fell through and got declared as arrays, which Arnold couldn't index correctly. Now a single value correctly applies to the whole shape regardless of interpolation scope.

Test: testsuite/test_2699 sets up an instanceable reference with an inherited primvars:crypto_asset, and asserts on the render log (no out-of-range/failed-lookup errors) rather than pixel values, since Cryptomatte hashes differ across reading paths.

I tested it with HTOA + nuke cryptomatte.

Fixes #2699


cpichard:
:white_check_mark: CI passed (build 11).

sebastienblor (approved):

sebastienblor (approved):

PR #2720, opened 2026-07-30
Add support for USD coordinate systems (coordSys) for camera projection

### LINKED ISSUE #32: [Render Delegate] Add support for Hydra "coordSys"
Add support for Hydra's `coordSys` feature that allows shaders to define coordinate systems for transformations.

[OpenUSD page](https://graphics.pixar.com/usd/docs/Coordinate-Systems-in-USD-Proposal.html)

cpichard:

Add support for USD coordinate systems (`UsdShadeCoordSysAPI` / Hydra `coordSys`) in the render delegate, so MaterialX/OSL shaders can transform into a named coordinate system — typically a camera projection (`map_proj`, `map_proj:ndc`, `map_proj:raster`, `map_proj:screen` as authored by Houdini/Solaris).

Arnold's OSL render services resolve a named space by looking up a **camera node** whose name matches the part before the suffix (`<name>.camera`, `<name>.NDC`, `<name>.screen`, `<name>.raster`), via `AiNodeLookUpByName`. The implementation follows from that:

- **`HdArnoldCoordSys` sprim** (`libs/render_delegate/coord_sys.{h,cpp}`) — new `HdCoordSys` implementation that creates an Arnold camera node per coordinate system, named uniquely from the sprim id so several coordinate systems sharing a name never collide. When the coordinate system is bound to a camera prim, the bound camera's transform *and* frustum are mirrored (`persp_camera` or `ortho_camera`, recreating the node in place if the projection type changes), so the projective `.NDC`/`.screen`/`.raster` spaces match the projector. The bound camera is resolved from the sprim id in the scene-index path, and by world-transform matching against the `camera` sprims in the legacy delegate path (where the id only encodes the binding relationship), so `hydra` and `hydra2` agree.
- **`coordSys` registered as a supported sprim type** and created in `HdArnoldRenderDelegate::CreateSprim` (`render_delegate.cpp`). Registering the token is also what makes `UsdImagingDelegate` answer coord-sys binding queries.
- **Scene-index plugin** (`plugins/scene_index/coordSysSIP.{h,cpp}`) — `UsdImagingCreateSceneIndices` does not turn coordSys bindings into coordSys prims, so a renderer scene-index plugin appends `HdsiCoordSysPrimSceneIndex` for renderer "Arnold". This makes it work in *any* Hydra host (usdview, husk, the procedural), not just one. Correspondingly, the procedural's manual `AppendSceneIndicesForRenderer` call in `reader.cpp` was removed: `HdRenderIndex` already applies renderer plugins at its terminal, so it ran every plugin twice — harmless for the pre-existing filtering plugins, but the prim-adding coordSys plugin produced a duplicate camera.
- **Space-name translation** (`libs/common/materials_utils.cpp`) — MaterialX geometric nodes (`ND_position_vector3`, ...) become Arnold OSL nodes whose `space` input is rewritten to Arnold's naming: `map_proj:ndc` → `map_proj.NDC`, `:raster`/`:screen` → dotted, plain names → `<name>.camera`. Standard MaterialX/OSL spaces (`model`, `object`, `world`, ...) and already-dotted values are left untouched. Shared by the procedural and render-delegate paths.
- **Per-rprim camera binding** (`node_graph.{h,cpp}`, and every rprim that assigns a material) — because Arnold resolves named spaces *globally by node name*, the material's `space` strings must point at the camera bound by the specific rprim. `HdArnoldNodeGraph::RemapCoordSysSpaces` rewrites them, and the remap-aware `GetCached*Shader(remap)` overloads let a material shared by rprims with *different* bindings work: the first binding claims the base network (no overhead in the common case), and each conflicting binding gets its own re-translated variant with stable node names across re-syncs. The rprim side is the shared `HdArnoldGetCoordSysRemap()` helper in `coord_sys.{h,cpp}`, used by `mesh.cpp`, `basis_curves.cpp`, `points.cpp`, `native_rprim.cpp`, `gaussian_splat.cpp` and `volume.cpp`; each of those re-assigns its material on `DirtyCategories` (which carries binding changes) and includes that bit in its initial dirty bits. Light-filter graphs stay on the plain accessor — lights have no coordSys bindings.
- **Resolution-independent projection** (`render_delegate.{h,cpp}`, `render_pass.cpp`) — Arnold derives every camera's vertical fov from the *render frame* aspect ratio (`AiWorldToScreenMatrix`), not from the camera's own aperture, so a projector's vertical extent was wrong whenever the render resolution's aspect differed from the projector's aperture. Coordinate-system cameras register with the delegate and `UpdateCoordSysCameraProjections()` cancels the frame aspect from their vertical screen window in `HdArnoldRenderPass::_Execute`, after `xres`/`yres` are final for the render (a sync-time read is stale). It only writes when the value changed, so it does not restart the render every frame. This matches Karma, which uses the projector's own aperture.
- **Vertical-flip configuration** (`config.{h,cpp}`) — Arnold's `NDC` convention is Y-opposite to its `screen`/`raster`, so no single camera flip makes all four spaces agree. `HDARNOLD_coordsys_flip_ndc_v` (default on) routes `.NDC` through a second, extra-flipped camera node while `.camera`/`.screen`/`.raster` keep the primary one; `HDARNOLD_coordsys_flip_v` (default off) flips every space together. The defaults were chosen to agree with Karma out of the box for all four spaces.
- **Testsuite** — `test_13221` (perspective projectors), `test_13221.1` (orthographic projectors), `test_13221.2` (one shared material bound by two rprims to different cameras) and `test_13221.3` (curves + points, i.e. non-mesh rprims), exercising `.camera`/`.NDC`/`.raster`/`.screen` on both the `hydra` and `hydra2` passes. CoordSys is implemented in the render delegate only, so these opt out of the procedural `usd` pass through a per-pass skip mechanism: `tools/test/testsuite.py` now populates the previously-unused `pass_<name>` skip lists from `testsuite/groups`.

Fixes #32 (ARNOLD-13221)

**Additional context**

Reference: [Coordinate Systems in USD proposal](https://graphics.pixar.com/usd/docs/Coordinate-Systems-in-USD-Proposal.html)

Known limitations / follow-ups, all deliberate:

1. **The procedural (`usd`) pass has no coordSys support** — implemented in the render delegate only; the reader would need the same camera-node creation, frustum mirroring, name remap and aspect cancellation. Either factor the shared logic out of `coord_sys.cpp`, or let the reader route through Hydra.
2. **Shared material with conflicting bindings costs a network copy** — the conflicting rprim gets a re-translated variant of the material. Correct, but not free in memory.
3. **The space-name rewrite is heuristic for plain names** — without the rprim's binding list at material-translation time, a plain non-standard space name cannot be distinguished from an arbitrary custom OSL space, so it gets a `.camera` suffix. The colon forms are unambiguous. Gating on `GetCoordSysBindings()` membership would change the shared `materials_utils.cpp` signatures.
4. **Legacy delegate camera recovery is by transform matching** — two projectors at the identical world transform with different frustums are indistinguishable; the code returns `nullptr` (transform-only fallback) rather than pick wrong. The legacy delegate exposes no way to map the binding-relationship sprim id back to its backing camera prim.
5. **The "correct" projection orientation is still an open convention question** — the flip defaults match Karma, but neither Karma nor Arnold strictly follows the USD convention, so they may be revisited.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sebastienblor:
**node_graph.cpp —** _ResolveCoordSysTerminal: unconditional mutex on the parallel per-rprim material hot path

`GetCachedSurfaceShader/VolumeShader/DisplacementShader(binding) `now route through `_ResolveCoordSysTerminal`, which takes `_coordSysMutex `on every call. Since these are called from each Rprim's material-(re)assignment in Sync(), and rprims sync in parallel (and commonly share one material), this adds lock contention to a path that was previously lock-free — even for scenes/materials that never use a coordinate system.

For a material with no coord-sys space inputs the entire critical section is a no-op: `_coordSysNamesInGraph `is empty ⇒ `_CoordSysSignature `returns {} on its first line ⇒ `_AcquireCoordSysHold(owner, "") `early-returns (the material never handed out a non-empty hold) ⇒ we just return the base terminal. So we're paying lock/unlock + potential contention purely to do nothing.

Could we add a lock-free guard before the lock?

```
AtNode* HdArnoldNodeGraph::_ResolveCoordSysTerminal(
    const CoordSysBinding& binding, const TfToken& terminalName)
{
    // A material with no coordinate-system "space" inputs can never produce a
    // non-empty signature and can never be claimed by any rprim, so there is
    // nothing to serialise. Skip the mutex on the parallel per-rprim hot path.
    if (_coordSysNamesInGraph.empty())
        return _nodeGraphCache.GetTerminal(terminalName);

    std::lock_guard<std::mutex> guard(_coordSysMutex);
    ... // unchanged
}
```

`_coordSysNamesInGraph` is written only in `_CollectCoordSysNames `(material sync phase) and read here (rprim sync phase), which don't overlap, so the read is race-free. If you'd prefer to remove any memory-visibility doubt at zero cost, back it with a `std::atomic<bool> _hasCoordSysNames` set at the end of `_CollectCoordSysNames `and test that instead.

This keeps the mutex where it's genuinely needed (parallel rprims mutating a coord-sys material's shared claim bookkeeping) while restoring the lock-free path for the common no-coord-sys case.

cpichard:
Thanks, good catch — the mutex is pure cost for materials with no coordinate systems, and I've added the guard.

I used the `std::atomic<bool>` form you suggested, but widened the predicate: `_coordSysNamesInGraph.empty()` alone has a hole. It's recomputed from scratch on each material Sync, so it can go non-empty → empty *while claims are still outstanding*:

1. material uses `map_proj`, rprims claim variants;
2. the `space` input is edited to `world`; next Sync leaves the name set empty but `_coordSysHolds` / `_coordSysVariants` / `_baseCoordSysSignature` populated;
3. `DirtyDependency` re-syncs the rprims, they take the early-out, and the release side of `_AcquireCoordSysHold` never runs.

The abandoned variant is then never retired and gets re-translated on every later Sync. So the flag now covers "any coordinate-system state at all" — names, base signature, holds, live variants, retired variants — stored under the lock at the end of `_RebuildCoordSysRemaps`. It can only flip false→true inside a material Sync (creating any state requires a non-empty signature, which requires a non-empty name set, which already implies true), which is what makes the unlocked read sound rather than merely probably-sound.

**I measured it**, because "adds contention" deserves a number. I instrumented `_ResolveCoordSysTerminal` with a switch that reproduces the pre-guard path exactly rather than approximating it: with no coordSys state the critical section computes an empty signature on its first line, early-returns from `_AcquireCoordSysHold`, and ends in the same `GetTerminal`, so the only difference between the two runs is the `lock`/`unlock`. 20k meshes sharing one material, 40 000 calls, 10 threads:

|                 | mean/call  | of which lock acquire | worst single acquire |
| --------------- | ---------- | --------------------- | -------------------- |
| **guarded**     | 27–33 ns   | —                     | —                    |
| **forced lock** | 152–178 ns | 87–113 ns             | 164 µs               |

That's ~5 ms of a ~4–5 s run, so **~0.1%**. The guard is worth having because it's free and restores a previously lock-free path, not because the lock was a measured bottleneck. The one genuinely unpleasant number is the tail: a thread waited up to 164 µs on a critical section that does nothing.

**The measurement turned up something bigger, though.** On a scene that *does* use coordinate systems (2 projectors, one shared material, 20k meshes) the same 40 000 calls cost 96–109 ms — roughly **20× more than what the guard removes**, ~1.8% of the run — because that critical section does real work. About half of it was `_CoordSysSignature` sorting and concatenating a string *while holding the lock*, and it only reads the caller's own remap plus the phase-stable `_coordSysNamesInGraph`, so it can be built before locking by exactly the same argument that justifies the guard. Hoisting it took the lock wait from 1379–1709 ns to 961–1120 ns per call and the total from 96–109 ms to 73–81 ms, a ~30% reduction in waiting.

Both changes are pushed, and the coordSys tests (`test_13221*`) still pass.

The harness itself isn't part of the PR — it needs instrumentation in the shipped source that we don't want to carry. I've packaged it separately (bench header, scene generator, runner, and a patch that wires it in) and can attach it here or on ARNOLD-13221 if you want to reproduce or extend it.

cpichard:
For the record, here is the benchmark code [coordsys_lock_bench.tar.gz](https://github.com/user-attachments/files/30895860/coordsys_lock_bench.tar.gz)

sebastienblor (approved):

PR #2721, opened 2026-07-30
Support pausing the render in the render delegate (#2719)

### LINKED ISSUE #2719: Support for pausing the render
At the moment the render is stopped when the render delegate pauses the render. We want to use the pause rendering feature instead of stopping

cpichard:
## Summary

Implements resumable pause/resume in the Hydra render delegate (#2719). Until now `HdArnoldRenderDelegate` advertised `IsPauseSupported() == false` and every "pause" path went through `AiRenderInterrupt()`, which tears down the in-flight render and loses all accumulated samples. With Arnold 7.5.4+ we can instead park the render threads at a gate with `AiRenderPause()` / `AiRenderResume()`, preserving render progress.

## Changes

**`libs/render_delegate/render_param.{h,cpp}`**
- `Pause()` now issues `AiRenderPause()` on Arnold >= 7.5.4, keeping the render alive; the previous interrupt-based behaviour is kept as the fallback for older Arnold.
- `Resume()` explicitly calls `AiRenderResume()` instead of waiting for the next `UpdateRender()` poll — `AiRenderGetStatus()` keeps reporting `RENDERING` while gated, so `UpdateRender()` has no other signal to act on.
- `Interrupt()` gains a `clearPaused` parameter (default `true`). Arnold's pause gate has no way to discard accumulated samples across a scene edit, so any real interrupt resumes/tears down a gated render; the `_paused` bookkeeping is now cleared to match. Without this the flag stayed stuck at `true` forever, since nothing else clears it on that path. `Pause()`'s pre-7.5.4 fallback passes `clearPaused=false`, since there the interrupt *is* the pause mechanism.
- New `IsPaused()` accessor.
- `UpdateRender()` re-issues `AiRenderPause()` when it observes status `RENDERING`, covering the case where `Pause()` raced a brief `RESTARTING` window (during which `AiRenderPause()` silently no-ops).

**`libs/render_delegate/render_delegate.{h,cpp}`**
- `IsPauseSupported()` returns `true` on Arnold >= 7.5.4.
- New `Pause()` and `IsPaused()` overrides (Houdini and other Hydra hosts query `IsPaused()`).
- `Stop()`, the Houdini COP-texture-changed path, and the `flush_texture` command now call `Interrupt(false, false)` directly rather than `Pause()`: they need a hard stop so the subsequent cache flush is safe, and `Pause()` no longer provides one.

**`testsuite/test_2719/`** — new C++ test that constructs `HdArnoldRenderDelegate` directly and drives `Pause()`/`Interrupt()`/`Resume()`/`Stop()`/`IsPaused()` against a live render, asserting that a scene edit and a `Stop()` both clear the paused state. `HdRenderDelegate::IsPaused()` isn't reachable from a normal `scene.usda`-driven test (not exposed through `UsdImagingGLEngine` or the Python bindings), hence the direct approach. The scene is built with plain `AiNode()` calls and `ARNOLD_PLUGIN_PATH` is cleared in `main()`: the test statically links USD via `render_delegate`, and letting `AiBegin()` auto-load `usd_proc` would put a second private copy of USD's global registries in the process and crash. See `testsuite/test_2719/README` for the full rationale.

**`testsuite/SConscript`** — link `render_delegate` + `common` and their USD dependencies into `test.cpp` builds when `BUILD_RENDER_DELEGATE=1`. There's no per-test env override hook in `Test.prepare()`, so this applies to every `test.cpp`; harmless for tests that reference none of those symbols, as unused static library members aren't pulled in at link time.

**`tools/utils/dependencies.py`** — `render_delegate()` had no `USD_BUILD_MODE == 'static'` case (unlike `translator()`), so it could never link against a statically-built USD. Likely a pre-existing, previously-unexercised gap rather than something specific to this test.

## Notes

- Behaviour on Arnold < 7.5.4 is unchanged: `IsPauseSupported()` still returns `false` and `Pause()` still interrupts.
- `CHANGELOG.md` entry added under a new `7.5.4.0 Tentative` section.

Fixes #2719

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 <!-- SPDX-License-Identifier: Apache-2.0 -->
 # Changelog
 
+## [7.5.4.0 Tentative] (Unreleased)
+
+- [usd#2719](https://github.com/Autodesk/arnold-usd/issues/2719) - Add support for resumable rendering in the render delegate.
 
 ## [7.5.3.1] (Unreleased)
 
--- a/libs/render_delegate/render_delegate.cpp
+++ b/libs/render_delegate/render_delegate.cpp
@@ -1953,7 +1953,9 @@ bool HdArnoldRenderDelegate::IsPauseSupported() const
 #endif
 }
 
+#if PXR_VERSION >= 2203
 bool HdArnoldRenderDelegate::IsPaused() const { return _renderParam->IsPaused(); }
+#endif
 
 bool HdArnoldRenderDelegate::IsStopSupported() const { return true; }
 
@@ -1963,9 +1965,9 @@ bool HdArnoldRenderDelegate::Stop(bool blocking)
 bool HdArnoldRenderDelegate::Stop()
 #endif
 {
-    // A hard stop: fully interrupt the render rather than the resumable pause
-    // gate used by Pause() below.
-    _renderParam->Interrupt(false, false);
+    // A hard stop: fully interrupt the render, and keep it stopped until Restart(),
+    // rather than parking it at the resumable pause gate used by Pause() below.
+    _renderParam->Stop();
     return true;
 }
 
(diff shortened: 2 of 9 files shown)


sebastienblor:
LGTM, but we first need to resolve conflicts

sebastienblor (approved):

PR #2723, opened 2026-07-31
Fix the Solaris viewport compositing

TomMinor:
This PR fixes inconsistencies in our Solaris viewport compositing when rendering volumes without a domelight (i.e. the background alpha isn't pure white). Let's start with some Karma examples for context.

## Karma Behaviour

| Simple Light _**(Works)**_ | Domelight + Visible Background _**(Works)**_ | Domelight + No Background _**(Works)**_ |
| ------------- | ------------- |  ------------- | 
| image | image | image | 

## Current (Broken) Arnold Behaviour

| Default light **(Broken)** | Domelight + Visible Background _**(Works)**_ | Domelight + No Background **(Broken)** |
| ------------- | ------------- |  ------------- | 
| image|  image| image | 

Besides the shader difference (Karma defaults to emission via density, we don't, so our result is a way darker volume), it's clear that in any setup without a visible environment background, Arnold's volume is transparent when it shouldn't be. This is more obvious with lighter backgrounds, which is problematic as the 'grey' viewport colour is the default in Houdini 22.

## Fixed Arnold Behaviour

| Simple Light _**(Works)**_ | Domelight + Visible Background _**(Works)**_ | Domelight + No Background _**(Works)**_ |
| ------------- | ------------- |  ------------- | 
| image | image | image |

- HTOA-3344

**Additional context**
This will also require some HtoA changes:
- `HYDRA_NORMALIZE_DEPTH=True` - Use P for depth directly (P is written for volume samples, Z isn't)
- `"depthstyle": "opengl"` - The P-path math already outputs in the range of [0, 1], which the opengl style gives us

### DIFF
--- a/libs/common/constant_strings.h
+++ b/libs/common/constant_strings.h
@@ -436,7 +436,6 @@ ASTR(param_colorspace);
 ASTR(param_filename);
 ASTR(param_shader_file);
 ASTR(parent_instance);
-ASTR(passthrough);
 ASTR(path);
 ASTR(penumbra_angle);
 ASTR(periodic);
--- a/libs/common/constant_strings.h
+++ b/libs/common/constant_strings.h
@@ -436,6 +436,7 @@ ASTR(param_colorspace);
 ASTR(param_filename);
 ASTR(param_shader_file);
 ASTR(parent_instance);
+ASTR(passthrough);
 ASTR(path);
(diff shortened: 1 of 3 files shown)


sebastienblor (approved):

TomMinor:
I ran the testsuite locally on this and it passed

PR #2748, opened 2026-08-27
MTOA-3029 Fix crashes when rendering in maya with the bundle

sebastienblor:
When we render in maya with the new bundle, and in hydra2, we have crashes due to our plugins node being loaded by USD. This is due to an ordering problem with the loading of the different plugins.
We need to detect that use case, and create our own render delegate in that case.


sebastienblor:
this is probably not the right fix, let's hold on this one until we know more. We might prefer just to avoid the crash and not render anything

PR #2749, opened 2026-08-31
Geometry deduplication in the Hydra render delegate

sebastienblor:
### Summary

Renders geometrically identical geometries as instances of a single canonical Arnold node instead of duplicating the geometry (and its BVH). The primary target is **point-instancer prototype flattening**: UsdImaging re-roots a copy of each instancer's prototype, so geometrically identical prototypes reach the render delegate as separate shapes. This change detects those duplicates from the USD data — *before* any geometry is translated to Arnold — and shares one canonical node, cutting memory and BVH-build cost on scenes with repeated assets. Both **meshes** and **basis curves** are supported.

### Modes

Controlled by the `HDARNOLD_GEOM_DEDUP` environment variable, read fresh on each render-delegate construction (so a DCC can toggle it between renders and tests can exercise all modes in one process):

| Value | Mode | Behavior |
|-------|------|----------|
| `0` | `None` | No deduplication — every shape keeps its own geometry. |
| `1` | `Instances` | **(default)** Only point-instancer prototypes are deduplicated. The common non-instanced shapes are left untouched and never pay the hashing cost. |
| `2` | `All` | Plain non-instanced duplicates are deduplicated too, rendered as `ginstance` nodes. |

### How it works

Two flavors of sharing, common to meshes and curves:

- **Instanced prototype** (default flattening case): the prototype's Arnold instancer is redirected to the shared canonical node via `HdArnoldShape::SetPrototypeOverride`. The prototype is forced onto the instancer-node path (`SetForceInstancerNode`) so shape-instancing never bakes `instance_matrix` onto a node that may be shared.
- **Non-instanced duplicate** (`All` mode): the shape node is converted into a `ginstance` of the canonical (`ConvertToInstanceOf`), keeping its own transform and surface shader while sharing the geometry.

A registry in `HdArnoldRenderDelegate` (`AcquireCanonicalGeometry` / `ReleaseCanonicalGeometry` / `OnGeometryDestroyed`) maps a geometry hash to the canonical node and refcounts the instances. When a canonical rprim is destroyed while instances still reference it, the delegate **adopts** the node and keeps it alive until the last instance releases it. The registry is type-agnostic; the shape type (`polymesh` / `curves`) is folded into the hash key so a mesh and a curve with a colliding hash can never be merged.

### Shared rprim infrastructure

The dedup orchestration (state, revert, acquire/redirect wiring, and destructor handoff) lives once in the `HdArnoldRprim` base template as `_ApplyGeometryDedup` / `_RevertGeometryDedup` / `_HandOffDedupOnDestroy`. Each geometry rprim only supplies the two type-specific pieces: an eligibility check and a `_ComputeGeometryHash`. This keeps the mesh and curve paths from drifting.

### Geometry hashing

`_ComputeGeometryHash` folds in everything that ends up on the shared node so only truly interchangeable geometries merge:

- **Mesh:** topology, points, all primvars (uvs, normals, custom, constant), display-style refinement, and the resolved **displacement** shader (a ginstance can't override displacement).
- **Curves:** curve topology (vertex counts, indices, type, basis, wrap), points, and all primvars (widths/radius, orientations, uvs, custom, constant). Curves have no displacement or subdivision.

For instanced prototypes, both additionally fold in the prototype's transform and resolved surface shader. Both also fold in render tag (purpose) and light-linking categories to stay conservative and race-free regardless of which duplicate becomes the canonical.

**Motion blur:** deformation-motion-blurred geometry participates too. All position samples (every motion key *and* its sample time) are folded into the hash, so two shapes merge only if their deformation is identical at every shutter sample. Since Arnold interpolates linearly between keys, matching keys guarantee matching motion — the merge is exact, not a current-frame approximation.

### Scope / limitations

Deduplicates **static** and **deformation-motion-blurred** geometry. Still excluded: computed/skinned points, velocity/acceleration-based motion blur, geom subsets (mesh), and mesh lights. **Transform** motion blur on instanced prototypes is out of scope — the prototype transform is hashed as a single sample (a separate axis from deformation).

### Files changed

- **`render_delegate.{h,cpp}`** — `GeometryDedupMode` enum, `HDARNOLD_GEOM_DEDUP` env-var parsing, and the type-agnostic canonical-geometry registry (acquire/release/adopt-on-destroy).
- **`rprim.h`** — shared dedup state and the `_ApplyGeometryDedup` / `_RevertGeometryDedup` / `_HandOffDedupOnDestroy` helpers used by both geometry rprims.
- **`mesh.{h,cpp}`** — mesh eligibility + `_ComputeGeometryHash` (incl. multi-key points); geometry/primvar translation skipped for instances (per-instance constant primvars, transform, visibility, shader still applied; canonical dependency registered so duplicates re-sync when the canonical changes). Now uses the shared base helpers.
- **`basis_curves.{h,cpp}`** — curves eligibility + `_ComputeGeometryHash`; the same instance guards, per-instance constant-primvar handling, canonical-dependency tracking, and destructor handoff.
- **`shape.{h,cpp}`** — `ConvertToInstanceOf`, `ReleaseShapeOwnership`, `SetPrototypeOverride`, `SetForceInstancerNode`, and instancer wiring to reference the shared canonical (type-neutral).
- **`testsuite/test_18101/`** — exercises all three modes through the usd procedural's Hydra path, asserting the resulting Arnold scene directly (counting geometry polymeshes, geometry curves, and ginstances) rather than by image comparison.

### Testing

`test_18101` builds parallel mesh and curve groups — two identical prototypes (instancers), one differing prototype, and two identical non-instanced shapes — and verifies per mode:

| mode | geometry polymeshes | geometry curves | ginstances |
|------|--------------------|-----------------|------------|
| 0 None | 5 | 5 | 0 |
| 1 Instances | 4 | 4 | 0 |
| 2 All | 3 | 3 | 2 (1 mesh + 1 curve) |


cpichard:
:x: CI failed (build 45).

sebastienblor:
> ❌ CI failed (build 45).

it failed because it was building against an older core, it's passing with the latest one

PR #2755, opened 2026-09-09
Update references images to match latest 7.5.3.0 build

cpichard:


jimenea (approved):
References images regenerated using the right Arnold core version.

